PR #622, opened 2019-08-01
Flag allowing to disable strict hostname check of kafka brokers

wikp:
Currently in brooklin Kafka broker hostnames should pass Apache Commons `DomainValidator` validation. There are companies on the market (including mine ;) ) which for historical reasons have their internal hostnames with, for example, invalid or non-existing TLD. 

I understand that this check may be required for security reasons for example but, on the other hand, it would be great to have an option to disable it. This PR is all about it ;) 

One thing I also changed is removing redundant hostname validation in `KafkaBrokerAddress` constructor. Objects of this class are constructed in `valueOf` method only (and in tests) which perform validation itself. If you want me to revert this piece of code I have to mention that it will be needed to add third parameter to the constructor which allows me to pass it to `validateHostname` method and so on :) 

I'm open to all suggestions :)


ahmedahamid:
Thank you, @wikp. Strict hostname validation check has been removed by PR #656. I'm closing this PR as a result but feel free to reopen as needed.

yakhyadabo:
Hi, 
I'm having hostname validation error. 
How can I disable the check ?

brooklin version : 1.0.0

PR #626, opened 2019-08-09
fix a race condition in handleChangeAssignment

Jyouping:
As handleChangeAssignment currently uses a thread pool, there is a race condition that
the previous assignment may actually process later than the new assignment. This will make
the follower instances don't contain the latest assignment from Zookeeper. The lack of synchronization in handleAssignmentChange may also end up creating unmonitored consumer, causing a rebalance issue. This fix set the thread pool size to 1 as default to avoid above issues.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
@@ -119,7 +119,7 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
   protected abstract AbstractKafkaBasedConnectorTask createKafkaBasedConnectorTask(DatastreamTask task);
 
   @Override
-  public void onAssignmentChange(List<DatastreamTask> tasks) {
+  public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
     _logger.info("onAssignmentChange called with tasks {}", tasks);
 
     HashSet<DatastreamTask> toCancel = new HashSet<>(_runningTasks.keySet());
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -19,11 +19,10 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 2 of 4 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorConfig.java
```diff
@@ -51,6 +52,7 @@ public CoordinatorConfig(Properties config) {
     _retryIntervalMs = _properties.getInt(CONFIG_RETRY_INTERVAL, 1000 /* 1 second */);
     _heartbeatPeriodMs = _properties.getLong(CONFIG_HEARTBEAT_PERIOD_MS, Duration.ofMinutes(1).toMillis());
     _defaultTransportProviderName = _properties.getString(CONFIG_DEFAULT_TRANSPORT_PROVIDER, "");
+    _assignmentChangeThreadPoolThreadCount = _properties.getInt(CONFIG_DEFAULT_ASSIGNMENT_THREAD_COUNT, 1);
```
gauravrkulkarni: If making the count more than 1 is going to cause an issue, do you think we should just skip reading the value from config? Might want to ignore the config value and deprecate the config?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -119,7 +119,7 @@ public AbstractKafkaConnector(String connectorName, Properties config, GroupIdCo
   protected abstract AbstractKafkaBasedConnectorTask createKafkaBasedConnectorTask(DatastreamTask task);
 
   @Override
-  public void onAssignmentChange(List<DatastreamTask> tasks) {
+  public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
```
gauravrkulkarni: do all connectors need this to be synchronized?

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorConfig.java
Jyouping: I'm fine with that. I think the original design is to have one thread to per connector, but it requires more change to the thread pool, might be better to fix it into 1.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
Jyouping: Technically, it depends on connector implement this, a connector can choose to implement a fine grained logic themselves. I also see other connectors, like a fileConnectopr implement synchronized on the entire method though.

gauravrkulkarni: Cool! Might want to consider making it a part of the documentation.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -271,6 +266,9 @@ public void start() {
       ConnectorInfo connectorInfo = _connectors.get(connectorType);
       ConnectorWrapper connector = connectorInfo.getConnector();
 
+      // Creating a separate thread pool for making the onAssignmentChange calls to the connector
+      _assignmentChangeThreadPool.put(connectorType, Executors.newSingleThreadExecutor());
```
gauravrkulkarni: nice!

PR #631, opened 2019-08-10
Kafka partition movement

Jyouping:
Implement the ability to move partitions to different machines

the contribution guidelines.

### DIFF
--- a/datastream-client/src/main/java/com/linkedin/datastream/DatastreamRestClient.java
+++ b/datastream-client/src/main/java/com/linkedin/datastream/DatastreamRestClient.java
@@ -533,9 +533,9 @@ public class DatastreamRestClient {
   /**
    * move partitions to the target host
    * @param datastreamName
-   *    Name of the datastream to resume.
+   *    Name of the datastream to move.
    * @param partitions
-   *    partitions that need to be moved
+   *    partitions that need to be moved, separate by comma, ex topicA-0,topicB-1
    * @param host
    *    the target host the partition assignment
    * @throws RemoteInvocationException
--- a/datastream-client/src/main/java/com/linkedin/datastream/DatastreamRestClient.java
+++ b/datastream-client/src/main/java/com/linkedin/datastream/DatastreamRestClient.java
@@ -536,16 +536,16 @@ public class DatastreamRestClient {
    *    Name of the datastream to move.
    * @param partitions
    *    partitions that need to be moved, separate by comma, ex topicA-0,topicB-1
-   * @param host
+   * @param targetHost
    *    the target host the partition assignment
    * @throws RemoteInvocationException
    */
-  public void movePartitions(String datastreamName, String partitions, String host) throws RemoteInvocationException {
+  public void movePartitions(String datastreamName, String partitions, String targetHost) throws RemoteInvocationException {
     Instant startTime = Instant.now();
     PollUtils.poll(() -> {
       try {
         ActionRequest<Void> request = _builders.actionMovePartitions().id(datastreamName).
-            partitionsParam(partitions).hostParam(host).build();
+            partitionsParam(partitions).targetHostParam(targetHost).build();
         ResponseFuture<Void> datastreamResponseFuture = _restClient.sendRequest(request);
         return datastreamResponseFuture.getResponse();
       } catch (RemoteInvocationException e) {
--- a/datastream-client/src/main/java/com/linkedin/datastream/DatastreamRestClient.java
+++ b/datastream-client/src/main/java/com/linkedin/datastream/DatastreamRestClient.java
@@ -531,6 +531,41 @@ public class DatastreamRestClient {
   }
 
   /**
+   * move partitions to the target host
+   * @param datastreamName
+   *    Name of the datastream to resume.
+   * @param partitions
+   *    partitions that need to be moved
+   * @param host
+   *    the target host the partition assignment
+   * @throws RemoteInvocationException
+   */
+  public void movePartitions(String datastreamName, String partitions, String host) throws RemoteInvocationException {
+    Instant startTime = Instant.now();
+    PollUtils.poll(() -> {
+      try {
+        ActionRequest<Void> request = _builders.actionMovePartitions().id(datastreamName).
+            partitionsParam(partitions).hostParam(host).build();
+        ResponseFuture<Void> datastreamResponseFuture = _restClient.sendRequest(request);
+        return datastreamResponseFuture.getResponse();
+      } catch (RemoteInvocationException e) {
+        if (ExceptionUtils.getRootCause(e) instanceof TimeoutException) {
+          LOG.warn("Timeout: movePartitions. May retry...", e);
+          return null;
+        }
+        if (isNotFoundHttpStatus(e)) {
+          LOG.warn(String.format("Datastream {%s} is not found", datastreamName), e);
+          throw new DatastreamNotFoundException(datastreamName, e);
+        } else {
+          String errorMessage = String.format("move partitions failed with error, Datastream {%s}.", datastreamName);
+          ErrorLogger.logAndThrowDatastreamRuntimeException(LOG, errorMessage, e);
+        }
+        return null; // not reachable
+      }
+    }, Objects::nonNull, getRetryPeriodMs(), getRetryTimeoutMs()).orElseThrow(RetriesExhaustedException::new);
+    LOG.info("move partitions for datastream {} took {} ms", datastreamName, Duration.between(startTime, Instant.now()).toMillis());
+  }
+  /**
    * Get all the datastream objects that are in the same group as the datastream whose name is "datastreamName".
    * This method makes a GET REST call to the Datastream management service which in turn fetches all the Datastream
    * objects from the store (ZooKeeper). Entries will be return in lexicographical based on their getName() property.
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
@@ -68,12 +68,11 @@ public interface AssignmentStrategy {
   }
 
   /**
-   * Move a partition for a datastream group according to the targetAssignment. As we are only allowed to mutate the
-   * task once. It follow the steps
-   * Step 1) get the partitions that to be moved, and get their source task
-   * Step 2) If the instance is the instance we want to move, we figure the task that we want to assign the task
-   * Step 3) We mutate and compute new task if they belongs to these source tasks or if they are the
-   * target task we want to move to
+   * Move a partition for a datastream group according to the targetAssignment.
+   * It returns a map from instance -> tasks map with partition info stored in the task.
+   *
+   * This interface needs to be implemented if the Brooklin Coordinator is going to perform the
+   * partition movement.
    *
    * @param currentAssignment the old assignment
    * @param targetAssignment the target assignment retrieved from Zookeeper
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
@@ -9,6 +9,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.commons.lang.NotImplementedException;
+
 import com.linkedin.datastream.server.DatastreamGroup;
 import com.linkedin.datastream.server.DatastreamGroupPartitionsMetadata;
 import com.linkedin.datastream.server.DatastreamTask;
@@ -68,19 +70,22 @@ public interface AssignmentStrategy {
   }
 
   /**
-   * Move a partition for a datastream group according to the targetAssignment.
+   * Move a partition for a datastream group according to the targetAssignment. While It moves a single datastream
+   * group, it can move multiple partitions to different hosts which are from multiple calls
+   *
    * It returns a map from instance -> tasks map with partition info stored in the task.
    *
    * This interface needs to be implemented if the Brooklin Coordinator is going to perform the
    * partition movement.
    *
-   * @param currentAssignment the old assignment
-   * @param targetAssignment the target assignment retrieved from Zookeeper
+   * @param currentAssignment the old assignment, it is a mapping from instance name -> datastream task
+   * @param targetAssignment the target assignment retrieved from Zookeeper, it is a mapping
+   *                         from instance name -> topicPartition
    * @param partitionsMetadata the subscribed partitions metadata received from connector
    * @return new assignment
    */
   default Map<String, Set<DatastreamTask>> movePartitions(Map<String, Set<DatastreamTask>> currentAssignment,
       Map<String, Set<String>> targetAssignment, DatastreamGroupPartitionsMetadata partitionsMetadata) {
-    return currentAssignment;
+    throw new NotImplementedException("movePartitions are not implemented");
   }
 }
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
@@ -78,7 +78,8 @@ public interface AssignmentStrategy {
    * This interface needs to be implemented if the Brooklin Coordinator is going to perform the
    * partition movement.
    *
-   * @param currentAssignment the old assignment, it is a mapping from instance name -> datastream task
+   * @param currentAssignment the old assignment, it is a mapping from instance name -> datastream task. The mapping
+   *                          should cover all datastream groups across all instances
    * @param targetAssignment the target assignment retrieved from Zookeeper, it is a mapping
    *                         from instance name -> topicPartition
    * @param partitionsMetadata the subscribed partitions metadata received from connector
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
@@ -82,6 +82,6 @@ public interface AssignmentStrategy {
    */
   default Map<String, Set<DatastreamTask>> movePartitions(Map<String, Set<DatastreamTask>> currentAssignment,
       Map<String, Set<String>> targetAssignment, DatastreamGroupPartitionsMetadata partitionsMetadata) {
-
+    return currentAssignment;
   }
 }
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
@@ -10,6 +10,7 @@ import java.util.Map;
 import java.util.Set;
 
 import com.linkedin.datastream.server.DatastreamGroup;
+import com.linkedin.datastream.server.DatastreamGroupPartitionsMetadata;
 import com.linkedin.datastream.server.DatastreamTask;
 
 
@@ -48,4 +49,39 @@ public interface AssignmentStrategy {
    */
   Map<String, Set<DatastreamTask>> assign(List<DatastreamGroup> datastreams, List<String> instances,
       Map<String, Set<DatastreamTask>> currentAssignment);
+
+
+  /**
+   * Assign partition for a particular datastream group to all the tasks in current assignment
+   * It return a map from instance -> tasks map with partition info stored in the task
+   * This interface needs to implemented if the Brooklin Coordinator is going to perform the
+   * partition assignment.
+   *
+   *
+   * @param currentAssignment the old assignment for all the datastream groups across all instances
+   * @param datastreamPartitions the subscribed partitions for the particular datastream group
+   * @return new assignment mapping
+   */
+  default Map<String, Set<DatastreamTask>> assignPartitions(Map<String,
+      Set<DatastreamTask>> currentAssignment, DatastreamGroupPartitionsMetadata datastreamPartitions) {
+    return currentAssignment;
+  }
+
+  /**
+   * Move a partition for a datastream group according to the targetAssignment. As we are only allowed to mutate the
+   * task once. It follow the steps
+   * Step 1) get the partitions that to be moved, and get their source task
+   * Step 2) If the instance is the instance we want to move, we figure the task that we want to assign the task
+   * Step 3) We mutate and compute new task if they belongs to these source tasks or if they are the
+   * target task we want to move to
+   *
+   * @param currentAssignment the old assignment
+   * @param targetAssignment the target assignment retrieved from Zookeeper
+   * @param partitionsMetadata the subscribed partitions metadata received from connector
+   * @return new assignment
+   */
+  default Map<String, Set<DatastreamTask>> movePartitions(Map<String, Set<DatastreamTask>> currentAssignment,
+      Map<String, Set<String>> targetAssignment, DatastreamGroupPartitionsMetadata partitionsMetadata) {
+
+  }
 }
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
@@ -53,8 +53,8 @@ public interface AssignmentStrategy {
 
   /**
    * Assign partition for a particular datastream group to all the tasks in current assignment
-   * It return a map from instance -> tasks map with partition info stored in the task
-   * This interface needs to implemented if the Brooklin Coordinator is going to perform the
+   * It returns a map from instance -> tasks map with partition info stored in the task
+   * This interface needs to be implemented if the Brooklin Coordinator is going to perform the
    * partition assignment.
    *
    *
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -327,11 +327,15 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
    * @param pathKeys resource key containing the datastream name
    * @param partitions partitions that need to move to
    * @param host host to accommodate the partitions
+   * @param notify specify if we should notify the leader to start process the assignment, we can stage the update
+   *               and batch it later by setting it into false
    * @return result HTTP status
    */
   @Action(name = "movePartitions", resourceLevel = ResourceLevel.ENTITY)
   public ActionResult<Void> movePartitions(@PathKeysParam PathKeys pathKeys,
-      @ActionParam("partitions") String partitions, @ActionParam("host") String host) {
+      @ActionParam("partitions") String partitions,
+      @ActionParam("host") String host,
+      @ActionParam("notify") @Optional("true") boolean notify) {
     String datastreamName = pathKeys.getAsString(KEY_NAME);
     Datastream datastream = _store.getDatastream(datastreamName);
 
@@ -350,7 +354,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
     List<String> targetPartitions = Arrays.asList(partitions.split(","));
     TargetAssignment targetAssignment = new TargetAssignment(targetPartitions, host);
     try {
-      _store.updatePartitionAssignments(datastream.getName(), datastream, targetAssignment);
+      _store.updatePartitionAssignments(datastream.getName(), datastream, targetAssignment, notify);
     } catch (Exception ex) {
       LOG.error("Error to move partitions", ex);
       _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_500_INTERNAL_SERVER_ERROR,
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -8,6 +8,7 @@ package com.linkedin.datastream.server.dms;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -45,6 +46,7 @@ import com.linkedin.datastream.server.Coordinator;
 import com.linkedin.datastream.server.DatastreamGroup;
 import com.linkedin.datastream.server.DatastreamServer;
 import com.linkedin.datastream.server.ErrorLogger;
+import com.linkedin.datastream.server.TargetAssignment;
 import com.linkedin.datastream.server.api.connector.DatastreamValidationException;
 import com.linkedin.datastream.server.api.security.AuthorizationException;
 import com.linkedin.restli.common.HttpStatus;
@@ -316,6 +318,45 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
     }
 
     LOG.info("Completed request for pausing datastream {}", datastream);
+
+    return new ActionResult<>(HttpStatus.S_200_OK);
+  }
+
+  /**
+   * Move partitions to a particular host
+   * @param pathKeys resource key containing the datastream name
+   * @param partitions partitions that need to move to
+   * @param host host to accommodate the partitions
+   * @return result HTTP status
+   */
+  @Action(name = "movePartitions", resourceLevel = ResourceLevel.ENTITY)
+  public ActionResult<Void> movePartitions(@PathKeysParam PathKeys pathKeys,
+      @ActionParam("partitions") String partitions, @ActionParam("host") String host) {
+    String datastreamName = pathKeys.getAsString(KEY_NAME);
+    Datastream datastream = _store.getDatastream(datastreamName);
+
+    LOG.info("Received request to move datastream {}", datastream);
+
+    if (datastream == null) {
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_404_NOT_FOUND,
+          "Datastream does not exist: " + datastreamName);
+    }
+
+    if (!DatastreamStatus.READY.equals(datastream.getStatus())) {
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_405_METHOD_NOT_ALLOWED,
+          "Can only move partitions in a READY datastream state: " + datastreamName);
+    }
+
+    List<String> targetPartitions = Arrays.asList(partitions.split(","));
+    TargetAssignment targetAssignment = new TargetAssignment(targetPartitions, host);
+    try {
+      _store.updatePartitionAssignments(datastream.getName(), datastream, targetAssignment);
+    } catch (Exception ex) {
+      LOG.error("Error to move partitions", ex);
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_500_INTERNAL_SERVER_ERROR,
+          "ERROR");
+    }
+    LOG.info("Completed request for moving datastream {}", datastream);
 
     return new ActionResult<>(HttpStatus.S_200_OK);
   }
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -351,6 +351,15 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
           "Can only move partitions in a READY datastream state: " + datastreamName);
     }
 
+    // validate that partition assignment is supported for the datastream
+    try {
+      _coordinator.validatePartitionAssignmentSupported(datastream);
+    } catch (DatastreamValidationException e) {
+      _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, CALL_ERROR, 1);
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_400_BAD_REQUEST,
+          "Partition assignment is not supported for this datastream ", e);
+    }
+
     List<String> targetPartitions = Arrays.asList(partitions.split(","));
     TargetAssignment targetAssignment = new TargetAssignment(targetPartitions, host);
     try {
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -134,7 +134,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
    * </ol>
    */
   private void doUpdateDatastreams(Map<String, Datastream> datastreamMap) {
-    LOG.info("Update datastream call with request: ", datastreamMap);
+    LOG.info("Update datastream call with request: {}", datastreamMap);
     _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, UPDATE_CALL, 1);
     if (datastreamMap.isEmpty()) {
       LOG.warn("Update datastream call with empty input.");
@@ -206,7 +206,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
     } catch (Exception e) {
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, CALL_ERROR, 1);
       _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_400_BAD_REQUEST,
-          String.format("Datastreams being updated do not form the same datastream group", datastreamsToUpdate));
+          String.format("Datastreams being updated do not form the same datastream group %s", datastreamsToUpdate));
     }
 
     // 4. check that the list of updated datastreams share the same destination and source
@@ -214,13 +214,13 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
     if (!datastreamsToUpdate.stream().allMatch(ds -> source.equals(ds.getSource()))) {
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, CALL_ERROR, 1);
       _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_400_BAD_REQUEST,
-          String.format("Sources must be consistent ", datastreamsToUpdate));
+          String.format("Sources must be consistent %s", datastreamsToUpdate));
     }
     DatastreamDestination destination = datastreamsToUpdate.get(0).getDestination();
     if (!datastreamsToUpdate.stream().allMatch(ds -> destination.equals(ds.getDestination()))) {
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, CALL_ERROR, 1);
       _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_400_BAD_REQUEST,
-          String.format("Destinations must be consistent ", datastreamsToUpdate));
+          String.format("Destinations must be consistent %s", datastreamsToUpdate));
     }
 
     // 5. check that the updated datastreams form the same datastream group as existing datastreams. Can't update
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -46,7 +46,7 @@ import com.linkedin.datastream.server.Coordinator;
 import com.linkedin.datastream.server.DatastreamGroup;
 import com.linkedin.datastream.server.DatastreamServer;
 import com.linkedin.datastream.server.ErrorLogger;
-import com.linkedin.datastream.server.TargetAssignment;
+import com.linkedin.datastream.server.HostTargetAssignment;
 import com.linkedin.datastream.server.api.connector.DatastreamValidationException;
 import com.linkedin.datastream.server.api.security.AuthorizationException;
 import com.linkedin.restli.common.HttpStatus;
@@ -326,7 +326,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
    * Move partitions to a particular host
    * @param pathKeys resource key containing the datastream name
    * @param partitions partitions that need to move to
-   * @param host host to accommodate the partitions
+   * @param targetHost target host to accommodate the partitions
    * @param notify specify if we should notify the leader to start process the assignment, we can stage the update
    *               and batch it later by setting it into false
    * @return result HTTP status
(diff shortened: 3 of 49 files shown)


datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
```diff
@@ -66,4 +66,21 @@
...
+
+  /**
+   * Move a partition for a datastream group according to the targetAssignment.
+   * It returns a map from instance -> tasks map with partition info stored in the task.
+   *
+   * This interface needs to be implemented if the Brooklin Coordinator is going to perform the
+   * partition movement.
+   *
+   * @param currentAssignment the old assignment
+   * @param targetAssignment the target assignment retrieved from Zookeeper
```
srinipunuru: Since currentAssignment and targetAssignment has same naming convention it is slightly confusing that both of them are going to be a Map of host -> datastreamTasks. but targetAssignment is a map of host -> topic Partitions  right? Can you please make that clear through the documentation. 
Can you also make it clear that movePartitions can perform movePartitions a single datastream Group and but multiple move calls across different hosts?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -219,6 +225,7 @@ private void onBecomeLeader() {
 
     _datastreamList = new ZkBackedDMSDatastreamList();
     _liveInstancesProvider = new ZkBackedLiveInstanceListProvider();
+    _targetAssignmentProvider = new ZkTargetAssignmentProvider(_connectorTypes);
```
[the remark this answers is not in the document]
Jyouping: Sure, I'm adding a check before creating

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -113,6 +116,50 @@ public void createDatastream(String key, Datastream datastream) {
...
+      _zkClient.ensurePath(path + '/' + current);
+      _zkClient.writeData(path + '/' + current, json);
+    }
+
+    if (notifyLeader) {
+      try {
+        _zkClient.writeData(KeyBuilder.getTargetAssignmentBase(_cluster, datastream.getConnectorName()),
+            String.valueOf(System.currentTimeMillis()));
+      } catch (Exception e) {
+        // we don't need to do an atomic update; if the node gets update by others somehow or get deleted by
```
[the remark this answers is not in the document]
Jyouping: updated

datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
Jyouping: updated, thanks for the suggestion.

datastream-client/src/main/java/com/linkedin/datastream/DatastreamRestClient.java
```diff
@@ -530,6 +530,41 @@ public void resume(String datastreamName, boolean force) throws RemoteInvocation
...
+   * move partitions to the target host
+   * @param datastreamName
+   *    Name of the datastream to resume.
+   * @param partitions
+   *    partitions that need to be moved
+   * @param host
+   *    the target host the partition assignment
+   * @throws RemoteInvocationException
+   */
+  public void movePartitions(String datastreamName, String partitions, String host) throws RemoteInvocationException {
```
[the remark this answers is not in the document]
Jyouping: Sure, thanks for suggestion.

datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
```diff
@@ -66,4 +66,21 @@
...
+   * @param partitionsMetadata the subscribed partitions metadata received from connector
+   * @return new assignment
+   */
+  default Map<String, Set<DatastreamTask>> movePartitions(Map<String, Set<DatastreamTask>> currentAssignment,
+      Map<String, Set<String>> targetAssignment, DatastreamGroupPartitionsMetadata partitionsMetadata) {
+    return currentAssignment;
```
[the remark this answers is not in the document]
Jyouping: throw an exception instead.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -320,6 +322,49 @@ public UpdateResponse update(String key, Datastream datastream) {
...
+
+    if (!DatastreamStatus.READY.equals(datastream.getStatus())) {
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_405_METHOD_NOT_ALLOWED,
+          "Can only move partitions in a READY datastream state: " + datastreamName);
+    }
+
+    List<String> targetPartitions = Arrays.asList(partitions.split(","));
+    TargetAssignment targetAssignment = new TargetAssignment(targetPartitions, host);
+    try {
+      _store.updatePartitionAssignments(datastream.getName(), datastream, targetAssignment, notify);
```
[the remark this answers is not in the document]
Jyouping: add a check validatePartitionAssignmentSupported before proceeding the targetAssignment

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -921,6 +938,70 @@ public void waitForDependencies(DatastreamTaskImpl task, Duration timeout) {
...
+      for (String node : nodes) {
+        String content = _zkclient.readData(path + '/' + node);
+        TargetAssignment assignment = TargetAssignment.fromJson(content);
+
+        //Compute the correct instance
+        Map<String, String> hostInstanceMap = new HashMap<>();
+        if (_liveInstancesProvider.getLiveInstances() != null) {
+          _liveInstancesProvider.getLiveInstances().forEach(instance -> {
+            try {
+              hostInstanceMap.put(instance.substring(0, instance.lastIndexOf('-')), instance);
```
srinipunuru: Looks like this code assumes the format of instance. Can we just create colocate the logic on where this format is created and how this parsing is done?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -921,6 +938,70 @@ public void waitForDependencies(DatastreamTaskImpl task, Duration timeout) {
...
+            } catch (Exception ex) {
+              LOG.info("Fail to parse instance name {}", instance, ex);
+            }
+          });
+        }
+        if (hostInstanceMap.containsKey(assignment.getTargetHost()) && assignment.getPartitionNames() != null) {
+          LOG.info("Added assignment {} {}", assignment.getTargetHost(), assignment.getPartitionNames());
+          result.put(hostInstanceMap.get(assignment.getTargetHost()), new HashSet<>(assignment.getPartitionNames()));
+        } else {
+          LOG.warn("instance {} not found in the map {}", assignment.getTargetHost(), hostInstanceMap.keySet());
```
srinipunuru: The log statement should be more clear on where the instance is not found. Map is a datastructure not the location that people reading the log can understand.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -921,6 +938,70 @@ public void waitForDependencies(DatastreamTaskImpl task, Duration timeout) {
...
+            try {
+              hostInstanceMap.put(instance.substring(0, instance.lastIndexOf('-')), instance);
+            } catch (Exception ex) {
+              LOG.info("Fail to parse instance name {}", instance, ex);
+            }
+          });
+        }
+        if (hostInstanceMap.containsKey(assignment.getTargetHost()) && assignment.getPartitionNames() != null) {
+          LOG.info("Added assignment {} {}", assignment.getTargetHost(), assignment.getPartitionNames());
+          result.put(hostInstanceMap.get(assignment.getTargetHost()), new HashSet<>(assignment.getPartitionNames()));
```
srinipunuru: What happens when there are two requests to move two different partitions sets to the same host?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -921,6 +938,70 @@ public void waitForDependencies(DatastreamTaskImpl task, Duration timeout) {
...
+
+    String path = KeyBuilder.getTargetAssignment(_cluster, connectorType, datastreamGroupName);
+    if (_zkclient.exists(path)) {
+      List<String> nodes = _zkclient.getChildren(path);
+      Collections.sort(nodes);
+      for (String node : nodes) {
+        String content = _zkclient.readData(path + '/' + node);
+        TargetAssignment assignment = TargetAssignment.fromJson(content);
+
+        //Compute the correct instance
```
srinipunuru: Looks like you are actually validating whether the assignment.getTargetHost() is present right? This comment seems incorrect.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -921,6 +938,70 @@ public void waitForDependencies(DatastreamTaskImpl task, Duration timeout) {
...
+        Map<String, String> hostInstanceMap = new HashMap<>();
```
srinipunuru: May be its a good idea to factor out the validation code to a separate function?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -113,6 +116,54 @@ public void createDatastream(String key, Datastream datastream) {
     _zkClient.writeData(path, json);
   }
 
+  @Override
+  public void updatePartitionAssignments(String key, Datastream datastream, TargetAssignment targetAssignment,
```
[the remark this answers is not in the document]
Jyouping: added, the document is also available in the interface though.

datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
```diff
@@ -66,4 +68,24 @@
...
+   * group, it can move multiple partitions to different hosts which are from multiple calls
+   *
+   * It returns a map from instance -> tasks map with partition info stored in the task.
+   *
+   * This interface needs to be implemented if the Brooklin Coordinator is going to perform the
+   * partition movement.
+   *
+   * @param currentAssignment the old assignment, it is a mapping from instance name -> datastream task
+   * @param targetAssignment the target assignment retrieved from Zookeeper, it is a mapping
+   *                         from instance name -> topicPartition
```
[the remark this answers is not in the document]
Jyouping: Its a mapping between instanceName -> Set<TopicPartitions>.
The set is generated from a comma separated string in the restful api.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -113,6 +116,54 @@ public void createDatastream(String key, Datastream datastream) {
...
+      _zkClient.ensurePath(path + '/' + current);
+      _zkClient.writeData(path + '/' + current, json);
+    }
+
+    if (notifyLeader) {
+      try {
+        _zkClient.writeData(KeyBuilder.getTargetAssignmentBase(_cluster, datastream.getConnectorName()),
+            String.valueOf(System.currentTimeMillis()));
+      } catch (Exception e) {
+        // The failure can be ignored here as the assignment is stored and it could still be processed
```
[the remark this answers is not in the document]
Jyouping: Nvm, I think it's better to re-throw the exception here.. I've updated the code

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -410,6 +413,12 @@ public void onDatastreamUpdate() {
     _log.info("Coordinator::onDatastreamUpdate completed successfully");
   }
 
+  @Override
+  public void onPartitionMovement() {
```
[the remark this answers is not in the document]
Jyouping: added, the document is also available in the interface.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
Jyouping: I sorted them based on the timestamps, so the last write win. I've added a doc on that behaviorl

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
Jyouping: It's aimed to map the hostname to correct Zookeeper instance name. I have fixed the comment

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
Jyouping: Fixed, thanks for suggestion.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
Jyouping: The format is used in current ZkAdapter, have colocated these into a helper functions.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
     return newAssignment;
   }
 
+  /**
+   * Move a partition for a datastream group according to the targetAssignment. As we are only allowed to mutate the
+   * task once. It follow the steps
+   * Step 1) get the partitions that to be moved, and get their source task
+   * Step 2) If the instance is the instance we want to move, we find out the task that we want to assign the task
```
srinipunuru: typo: we find out the task that we want to assign the topic partition?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+   *
+   * @param currentAssignment the old assignment
+   * @param targetAssignment the target assignment retrieved from Zookeeper
+   * @param partitionsMetadata the subscribed partitions metadata received from connector
+   * @return new assignment
+   */
+  public Map<String, Set<DatastreamTask>> movePartitions(Map<String, Set<DatastreamTask>> currentAssignment,
+      Map<String, Set<String>> targetAssignment, DatastreamGroupPartitionsMetadata partitionsMetadata) {
+
+    LOG.info("Move partition, task: {}, target assignment: {}, all partitions: {}", currentAssignment,
```
srinipunuru: Looks like typo in the log line. first argument passed is current Assignment not the task

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+        targetAssignment, partitionsMetadata.getPartitions());
+
+    String dgName = partitionsMetadata.getDatastreamGroup().getName();
+
+    Set<String> allToReassignPartitions = new HashSet<>();
+    targetAssignment.values().stream().forEach(allToReassignPartitions::addAll);
+    allToReassignPartitions.retainAll(partitionsMetadata.getPartitions());
+
+    //construct a map to store the tasks and if it contain the partitions that need to be released
+    //map: <taskName, partitions that need to be released>
```
srinipunuru: Can you add similar documentation for the second map (partitionToSourceTaskMap) as well

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+    Map<String, Set<String>> toReleasePartitionsTaskMap = new HashMap<>();
+    Map<String, String> partitionToSourceTaskMap = new HashMap<>();
+
+    //We first confirmed that the partitions in the target assignment can be removed, and we found its source task
```
srinipunuru: typo: We first confirm 

Can you also add details on the scenarios where we cannot remove the partitions from the task?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+    currentAssignment.keySet().stream().forEach(instance -> {
+      Set<DatastreamTask> tasks = currentAssignment.get(instance);
+      tasks.forEach(task -> {
+        if (dgName.equals(task.getTaskPrefix())) {
```
srinipunuru: I think we added a new method to getDatastreamGroupName from task no? can we use that?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+          Set<String> toMovePartitions = new HashSet<>(task.getPartitionsV2());
+          toMovePartitions.retainAll(allToReassignPartitions);
+          toReleasePartitionsTaskMap.put(task.getDatastreamTaskName(), toMovePartitions);
```
srinipunuru: What if the toMovePartitions is empty? we still add the datastreamTaskName to the list of all the tasks that has partitions to be released?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+          toMovePartitions.forEach(p -> partitionToSourceTaskMap.put(p, task.getDatastreamTaskName()));
+        }
+      });
+    });
+
+    Set<String> tasksToMutate = toReleasePartitionsTaskMap.keySet();
```
srinipunuru: toReleasePartitionsTaskMap that contains empty values doesn't need mutation right?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+    //Compute new assignment from the current assignment
+    Map<String, Set<DatastreamTask>> newAssignment = new HashMap<>();
+
+    currentAssignment.keySet().stream().forEach(instance -> {
+      Set<DatastreamTask> tasks = currentAssignment.get(instance);
+
+      // check if this instance has any partition to be moved
+      final Set<String> toMovedPartitions = new HashSet<>();
+      if (targetAssignment.containsKey(instance)) {
+        toMovedPartitions.addAll(targetAssignment.get(instance).stream().filter(toReleasePartitions::contains).collect(Collectors.toSet()));
```
srinipunuru: Can you add some comments on what you are doing in this lambdas?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+    Set<String> toReleasePartitions = new HashSet<>();
+    toReleasePartitionsTaskMap.values().forEach(v -> toReleasePartitions.addAll(v));
+
+    //Compute new assignment from the current assignment
+    Map<String, Set<DatastreamTask>> newAssignment = new HashMap<>();
+
+    currentAssignment.keySet().stream().forEach(instance -> {
+      Set<DatastreamTask> tasks = currentAssignment.get(instance);
+
+      // check if this instance has any partition to be moved
```
srinipunuru: // Check if the instance has any partitions that needs to be added to right?

Move can either mean remove or add. It will be good if you can be as precise here as possible

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+      final Set<String> toMovedPartitions = new HashSet<>();
+      if (targetAssignment.containsKey(instance)) {
+        toMovedPartitions.addAll(targetAssignment.get(instance).stream().filter(toReleasePartitions::contains).collect(Collectors.toSet()));
+      }
+
+      Set<DatastreamTask> dgTasks = tasks.stream().filter(t -> dgName.equals(t.getTaskPrefix()))
+          .collect(Collectors.toSet());
+      if (toMovedPartitions.size() > 0 && dgTasks.isEmpty()) {
+        throw new DatastreamRuntimeException("No task is available in the target instance " + instance);
```
srinipunuru: How do these errors get communicated to the operator that partition movement failed because of this issue.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+
+      // check if this instance has any partition to be moved
+      final Set<String> toMovedPartitions = new HashSet<>();
+      if (targetAssignment.containsKey(instance)) {
+        toMovedPartitions.addAll(targetAssignment.get(instance).stream().filter(toReleasePartitions::contains).collect(Collectors.toSet()));
+      }
+
+      Set<DatastreamTask> dgTasks = tasks.stream().filter(t -> dgName.equals(t.getTaskPrefix()))
+          .collect(Collectors.toSet());
+      if (toMovedPartitions.size() > 0 && dgTasks.isEmpty()) {
```
srinipunuru: Please log the error before throwing.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +141,116 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+        throw new DatastreamRuntimeException("No task is available in the target instance " + instance);
+      }
+
+      //find the target task to store the moved partition to store the target partitions
```
srinipunuru: Can you add comments on the logic that you are using to find the task for which the partitions need to be added to?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: Fixed, thanks.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: Fixed into currentAssignment

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: We should ignore the partitions if we cannot find out the partitions. Added this in the doc

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: Actually I figured out I can use datastreamGroup.belongsTo(task) to replace this. Thanks for suggestion.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: added, it's to filter the target assignment by the partitions which can be released from the task.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: use add instead

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: added the comment on the logic, which is to find the task with minimum number of partitions within that instance

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: Added a metric for num_errors in Coordinator where caught this exception.

datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
```diff
@@ -66,4 +68,24 @@
...
+  /**
+   * Move a partition for a datastream group according to the targetAssignment. While It moves a single datastream
+   * group, it can move multiple partitions to different hosts which are from multiple calls
+   *
+   * It returns a map from instance -> tasks map with partition info stored in the task.
+   *
+   * This interface needs to be implemented if the Brooklin Coordinator is going to perform the
+   * partition movement.
+   *
+   * @param currentAssignment the old assignment, it is a mapping from instance name -> datastream task
```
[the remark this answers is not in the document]
Jyouping: The mapping needs to cover all datastream groups across all instances. I have added in the doc

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1069,81 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+              subscribedPartitions);
+          toCleanup.add(dg);
+        }
+      }
+
+      for (String key : assignmentByInstance.keySet()) {
+        newAssignmentsByInstance.put(key, new ArrayList<>(assignmentByInstance.get(key)));
+      }
+
+      _adapter.updateAllAssignments(newAssignmentsByInstance);
```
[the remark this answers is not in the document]
Jyouping: Its logged inside strategy.movePartitions.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1069,81 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+      _log.info("Partition movement failed, Exception: ", ex);
+    }
+    if (!shouldRetry) {
+      _adapter.cleanupOldUnusedTasks(previousAssignmentByInstance, newAssignmentsByInstance);
+      updateCounterForMaxPartitionInTask(newAssignmentsByInstance);
+      _dynamicMetricsManager.createOrUpdateMeter(MODULE, NUM_PARTITION_MOVEMENTS, 1);
+    }  else if (!leaderPartitionMovementScheduled.get()) {
+      _log.info("Schedule retry for leader movement tasks");
+      _dynamicMetricsManager.createOrUpdateMeter(MODULE, "handleLeaderPartitionMovement", NUM_RETRIES, 1);
+      leaderPartitionMovementScheduled.set(true);
```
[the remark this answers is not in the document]
Jyouping: The flag is to avoid to schedule multiple retries. This is atomicInteger so I don't think we need to make it volatile? Also, this variable should be only access in the coordinator (single threaded manner)

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1069,81 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+
+      _log.info("Partition movement completed: datastreamGroup, assignment {} ", assignmentByInstance);
+
+      shouldRetry = false;
+    } catch (DatastreamTransientException ex) {
+      _log.info("Partition movement failed, retry again after a configurable period", ex);
+      shouldRetry = true;
+    } catch (Exception ex) {
+      //We do not retry if it is not transient exception
+      _log.info("Partition movement failed, Exception: ", ex);
```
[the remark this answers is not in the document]
Jyouping: Yea, this should be an error. Thanks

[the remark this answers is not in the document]
Jyouping: Since it's async process, we should monitor ingraph (handleLeaderPartitionMovement.error)

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -735,6 +748,10 @@ private String createLiveInstanceNode() {
   public void ensureConnectorZNode(String connectorType) {
     String path = KeyBuilder.connector(_cluster, connectorType);
     _zkclient.ensurePath(path);
```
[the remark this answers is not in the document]
Jyouping: I will rename into addConnectorType

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -735,6 +748,10 @@ private String createLiveInstanceNode() {
...
+    if (_targetAssignmentProvider != null) {
+      _targetAssignmentProvider.addListener(connectorType);
+    }
+    _connectorTypes.add(connectorType);
```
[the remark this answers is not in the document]
Jyouping: The ZKTargetAssignmentProvider is created to subscribe only in the leader, but the connector type should be kept in ZkAdapter all the time. Also, as we do call ZKAdapter to create connector node in the beginning, it makes sense to store in ZkAdapter even for caching purpose.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1207,6 +1295,28 @@ public void validateDatastreamsUpdate(List<Datastream> datastreams) throws Datas
...
+   * @throws DatastreamValidationException if partition assignment is not supported
+   */
+  public void validatePartitionAssignmentSupported(Datastream datastream) throws DatastreamValidationException {
+    try {
+      String connectorName = datastream.getConnectorName();
+      ConnectorInfo connectorInfo = _connectors.get(connectorName);
+      if (connectorInfo == null) {
+        throw new DatastreamValidationException("Invalid connector: " + connectorName);
+      }
+      if (!(connectorInfo.getAssignmentStrategy() instanceof StickyPartitionAssignmentStrategy)) {
```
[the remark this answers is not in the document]
Jyouping: I probably need some suggestions on this. One alternative is to add an interface in coordinator to indicate if it is supported, but I'm wondering if there is a better way.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -921,6 +938,70 @@ public void waitForDependencies(DatastreamTaskImpl task, Duration timeout) {
...
+      _zkclient.deleteRecursive(path);
+    }
+    LOG.info("clean up Target assignment info for {}", datastreamGroupName);
+  }
+
+  /**
+   * Get the datastream group which contains the partition movement info
+   * @param connectorType Connector name
+   */
+  public List<String> getDatastreamsNeedPartitionMovement(String connectorType) {
```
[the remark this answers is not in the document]
Jyouping: added an extra logic in Coordinator to clean up this.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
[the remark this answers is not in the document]
Jyouping: Have fixed to use the coordinator.getDatastreamPartitions to verify instead. This can also allows us to verify not only the connector but also individual datastream. Thanks for your suggestion.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1071,94 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+      _adapter.cleanupOldUnusedTasks(previousAssignmentByInstance, newAssignmentsByInstance);
+      updateCounterForMaxPartitionInTask(newAssignmentsByInstance);
+      _dynamicMetricsManager.createOrUpdateMeter(MODULE, NUM_PARTITION_MOVEMENTS, 1);
+    }  else if (!leaderPartitionMovementScheduled.get()) {
+      _log.info("Schedule retry for leader movement tasks");
+      _dynamicMetricsManager.createOrUpdateMeter(MODULE, "handleLeaderPartitionMovement", NUM_RETRIES, 1);
+      leaderPartitionMovementScheduled.set(true);
+      _executor.schedule(() -> {
+        _eventQueue.put(CoordinatorEvent.createPartitionMovementEvent());
+        leaderPartitionMovementScheduled.set(false);
```
gauravrkulkarni: Didn't quite understand the purpose of this flag - shouldn't leaderPartitionMovementScheduled be set to true since partition movement event is added to the queue (i.e. movement is scheduled) just before this statement?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1207,6 +1310,32 @@ public void validateDatastreamsUpdate(List<Datastream> datastreams) throws Datas
     }
   }
 
+  /**
+   * Validate if the partition assignment is supported for the datastream
+   * @param datastream datastream which needs the verification
+   * @throws DatastreamValidationException if partition assignment is not supported
+   */
+  public void validatePartitionAssignmentSupported(Datastream datastream) throws DatastreamValidationException {
```
gauravrkulkarni: Isn't this also a function of assignment strategy?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +142,126 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+
+      Set<DatastreamTask> dgTasks = tasks.stream().filter(dg::belongsTo)
+          .collect(Collectors.toSet());
+      if (toAddedPartitions.size() > 0 && dgTasks.isEmpty()) {
+        String errorMsg = String.format("No task is available in the target instance %s", instance);
+        LOG.error(errorMsg);
+        throw new DatastreamRuntimeException(errorMsg);
+      }
+
+      // find the task with minimum number of partitions on that instance to store the moved partitions
```
gauravrkulkarni: Instead of assigning all the partitions to just one task, is it possible to distribute it amongst all the tasks on that host, with tasks with less# partitions getting more of new partitions? I am thinking of a corner case where a lot of partitions are moved to one host, and they all get assigned to just one task.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -141,6 +142,126 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+   *         from that instance
+   * Step 3) Scan the current assignment, compute new task if the old task belongs to these source tasks or if it
+   *         is the target task we want to move to
+   *
+   * @param currentAssignment the old assignment
+   * @param targetAssignment the target assignment retrieved from Zookeeper
+   * @param partitionsMetadata the subscribed partitions metadata received from connector
+   * @return new assignment
+   */
+  public Map<String, Set<DatastreamTask>> movePartitions(Map<String, Set<DatastreamTask>> currentAssignment,
```
gauravrkulkarni: instead of adding a movePartitions implementation in strategy, I was thinking if we can create an interface findNewTask/similar, which returns new task(s) to assign the new partitions to.  All the tasks modifications are taken care of outside of strategy (if the modifications are generic enough for all connector types/strategies). That way new strategies just need to implement this new findNewTask method. Default implementation of findNewTask will be just returning task with least #partitions.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -921,6 +939,64 @@ public void waitForDependencies(DatastreamTaskImpl task, Duration timeout) {
...
+  /**
+   * Get a partition movement info for a particular datastream
+   * @param connectorType Connector name
+   * @param datastreamGroupName the name of datastream group
+   * @return The target assignment mapping information with the instance name -> topicPartition
+   */
+  public Map<String, Set<String>> getPartitionMovement(String connectorType, String datastreamGroupName) {
+    Map<String, Set<String>> result = new HashMap<>();
+
+    String path = KeyBuilder.getTargetAssignment(_cluster, connectorType, datastreamGroupName);
```
gauravrkulkarni: nit: might want to rename it to getTargetAssignmentPath

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
gauravrkulkarni: Should we process all of them instead?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -320,6 +322,58 @@ public UpdateResponse update(String key, Datastream datastream) {
...
+    Datastream datastream = _store.getDatastream(datastreamName);
+
+    LOG.info("Received request to move datastream {}", datastream);
+
+    if (datastream == null) {
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_404_NOT_FOUND,
+          "Datastream does not exist: " + datastreamName);
+    }
+
+    if (!DatastreamStatus.READY.equals(datastream.getStatus())) {
```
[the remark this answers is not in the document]
Jyouping: No, move partition should happen only when the datastream is running. We don't need to pause it. This is especially important as we don't want to have other assignment accidentally changed due to pause/restart.

I will add an comment on this

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -113,6 +116,60 @@ public void createDatastream(String key, Datastream datastream) {
...
 
+  /**
+   * update the target assignment info for a particular datastream
+   * @param key datastream name of the original datastream to be updated
+   * @param datastream content of the updated datastream
+   * @param targetAssignment the target partition assignment
+   * @param notifyLeader whether to notify leader about the update
+   */
+  @Override
+  public void updatePartitionAssignments(String key, Datastream datastream, TargetAssignment targetAssignment,
```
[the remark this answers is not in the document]
Jyouping: As all other methods are using key to represent datatreamName, will probably keep this naming convention.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -113,6 +116,60 @@ public void createDatastream(String key, Datastream datastream) {
...
+      boolean notifyLeader)
+      throws DatastreamException {
+    Validate.notNull(datastream, "null datastream");
+    Validate.notNull(key, "null key for datastream" + datastream);
+    verifyHostname(targetAssignment.getTargetHost());
```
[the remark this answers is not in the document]
Jyouping: The verification will need to involve ZkClient. The targetAssignment, however, is a serializable object which can be put into Zk, I prefer not to put the ZkCilent info in this object.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
Jyouping: This verify if the partition is being managed by this connector, I will adjust the comment to avoid some confusions.

datastream-server/src/main/java/com/linkedin/datastream/server/TargetAssignment.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.List;
+
+import com.linkedin.datastream.common.JsonUtils;
+
+
+/**
+ * Data structure to store the target assignment for partitions
+ */
+public class TargetAssignment {
```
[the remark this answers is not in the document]
Jyouping: WIll rename into HostTargetAssignment

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
Jyouping: Yes, all of them will be processed, but the last write wins.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1071,94 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+      }
+
+      _log.info("Partition movement completed: datastreamGroup, assignment {} ", assignmentByInstance);
+
+      shouldRetry = false;
+    } catch (DatastreamTransientException ex) {
+      _log.warn("Partition movement failed, retry again after a configurable period", ex);
+      shouldRetry = true;
+    } catch (Exception ex) {
+      //We do not retry if it is not transient exception
```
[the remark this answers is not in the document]
Jyouping: No, it is hard failure.  We don't retry but log error.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
Jyouping: I actually follow the convention that we currently did in LEADER_DO_ASSIGNMENTS. I think it is to avoid multiple one be put in the executor which schedules the event. But it is not used by queue as the queue itself will handle dedup.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1071,94 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+      _log.error("Partition movement failed, Exception: ", ex);
+      _dynamicMetricsManager.createOrUpdateMeter(MODULE, "handleLeaderPartitionMovement", NUM_ERRORS, 1);
+
+    }
+    if (!shouldRetry) {
+      _adapter.cleanupOldUnusedTasks(previousAssignmentByInstance, newAssignmentsByInstance);
+      updateCounterForMaxPartitionInTask(newAssignmentsByInstance);
+      _dynamicMetricsManager.createOrUpdateMeter(MODULE, NUM_PARTITION_MOVEMENTS, 1);
+    }  else if (!leaderPartitionMovementScheduled.get()) {
```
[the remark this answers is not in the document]
Jyouping: I'm not sure what do you mean here? We may need to retry even toMoveDatastream is not empty as the assignment encounter transient failure.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: I think you mean split step 1) into findNewTask, I don't think we can benefit too much as I don't think there are other usecases. The task modifications are should also be considered as the part of assignment strategy.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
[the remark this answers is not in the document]
gauravrkulkarni: Does this mean that partition movement will go away after pause/restart?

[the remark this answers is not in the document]
Jyouping: The host - task mapping will get changed after restart so the partition movement may no longer valid after that.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1071,94 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+      }
+
+      for (String key : assignmentByInstance.keySet()) {
+        newAssignmentsByInstance.put(key, new ArrayList<>(assignmentByInstance.get(key)));
+      }
+
+      _adapter.updateAllAssignments(newAssignmentsByInstance);
+
+      //clean up stored target assignment after the assignment is updated
+      for (DatastreamGroup dg : toCleanup) {
```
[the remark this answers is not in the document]
Jyouping: I added a timestamp to cut off assignments from different batch.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
Jyouping: Yes, this is possible. However, I'm inclined to keep the current algorithm primary due to 1) less mutation tasks which may also affect performance 2) simplicity in the algorithm 3) easy to control the assignment, for example we can decide to put into one task by a single call, or put into possible two tasks by splitting into more calls.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1056,6 +1071,94 @@ private void onDatastreamChange(List<DatastreamGroup> datastreamGroups) {
...
+
+        // Filtered all live datastreamGrou as we process only datastream which have
+        // both partition assignment info and the target assignment
+        List<DatastreamGroup> toProcessedDatastreamGroups =
+            liveDatastreamGroups.stream().filter(group2 -> toMoveDatastream.contains(group2.getName()))
+                .filter(group3 -> datastreamPartitions.keySet().contains(group3.getName()))
+                .collect(Collectors.toList());
+
+        for (DatastreamGroup dg : toProcessedDatastreamGroups) {
+          DatastreamGroupPartitionsMetadata subscribedPartitions = connectorInstance.getDatastreamPartitions().get(dg.getName())
```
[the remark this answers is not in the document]
Jyouping: I'm inclined to keep the current behavior. It might be easier to debug to see all or none behavior especially the currentAssignment store things from every datastream group.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
Jyouping: I changed the logic so that we will process all of them. ex. host1 -> p1, p2 and then host -> p2, p3 it will become host 1 -> p1, p2, p3. Have also added a test.

PR #632, opened 2019-08-10
Fix a bug when putting an event with metadata in event queue

Jyouping:
CoordinateEvents are getting deduped in the CoordinatorEventBlockingQueue based on the type. However, events with metadata shouldn't been deduped just based on the type. This change fixes the issue by only deduping the events with metadata. 

Going forward, we can consider dedup for eventMetadata as well. But this will requires eventMetadata object to implemented a proper hashcode logic.

the contribution guidelines.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEvent.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEvent.java
@@ -9,8 +9,8 @@ package com.linkedin.datastream.server;
 /**
  * Represents different event types inside {@link Coordinator}.
  *
- * CoordinatorEvent will be deduped in the event queue based on the event type
- * However, any event with eventMetadata will not get deduped
+ * CoordinatorEvent will be deduped in the event queue {@link CoordinatorEventBlockingQueue}
+ * based on the event type. However, any event with eventMetadata will not get deduped
  */
 public class CoordinatorEvent {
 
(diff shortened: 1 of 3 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
```diff
@@ -46,8 +46,11 @@ public synchronized void put(CoordinatorEvent event) {
       }
     }
 
-    // always overwrite the event in the map
-    _eventMap.put(event.getType(), event);
+    // we put into the eventMap for dedup only if there is no metadata
+    if (event.getEventMetadata() == null) {
```
gauravrkulkarni: We can go ahead with this for time being. In the long term we probably want the dedup logic to take metadata into consideration. With this implementation we have inconsistency that only events without metadata get deduped. Also there is no way for someone to know about this behavior unless checking out the code.

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEvent.java
```diff
@@ -8,6 +8,9 @@
 
 /**
  * Represents different event types inside {@link Coordinator}.
+ *
+ * CoordinatorEvent will be deduped in the event queue based on the event type
```
gauravrkulkarni: Could you please add description for what the bug was?

Jyouping: added a link

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
Jyouping: Deduping eventMetadata will requires assumptions to have the metadata properly implement the hashcode. This should be covered in the future implementation.

PR #634, opened 2019-08-16
Refactor KMMConnector interface

Jyouping:
There is a need for user to extend KMMConnector to put extra logic in the send callback. 
As a result, I refactor the interface and put the send callback in the method parameter.

The getter for metryRegistry is also added as there is also a need to get metryRegistry class when hooking up with other clients.


datastream-common/src/main/java/com/linkedin/datastream/metrics/DynamicMetricsManager.java
```diff
@@ -402,4 +402,12 @@ private void validateArguments(String classSimpleName, String metricName) {
     Validate.notNull(classSimpleName, "classSimpleName argument is null.");
     Validate.notNull(metricName, "metricName argument is null.");
   }
+
+  /**
+   * Get metricRegistry object, it allows other module like Kafka client to wire in the same metricRegistry
+   * @return
```
ahmedahamid: nit: remove empty return doc

PR #635, opened 2019-08-16
Bug fix with custom SerDe being overriden by kafka factory implementations

Rakesh627:
We are running `KafkaMirrorMakerConnector` and noticed that the values we provide as part of  `value.serializer` in transporter config and `value.deserializer` in connector config is not getting reflected in the Kafka consumer/producer client instantiations. 

This is because user provided custom Kafka deserializer or serializer in connector/transporter configs is overwritten in the the Kafka consumer & producer factory implementations.

This PR is a simple fix: 
- Made changes to add `value.serializer` only if absent in `KafkaProducerFactory` implementation
- Made changes to add `value.deserializer` only if absent in `KafkaProducerFactory` implementation
- Added some unit tests around these factory implementation for future proofing.

Above fix will enable users to provide Kafka custom SerDe's as part of their connector/transporter configurations.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConsumerFactoryImpl.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConsumerFactoryImpl.java
@@ -17,15 +17,15 @@ import org.apache.kafka.common.serialization.ByteArrayDeserializer;
  * A factory for creating Kafka {@link Consumer} instances
  */
 public class KafkaConsumerFactoryImpl implements KafkaConsumerFactory<byte[], byte[]> {
-  private static final String KEY_DESERIALIZER = ByteArrayDeserializer.class.getCanonicalName();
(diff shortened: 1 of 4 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConsumerFactoryImpl.java
```diff
@@ -17,10 +17,21 @@
  * A factory for creating Kafka {@link Consumer} instances
  */
 public class KafkaConsumerFactoryImpl implements KafkaConsumerFactory<byte[], byte[]> {
+  private static final String KEY_DESERIALIZER = ByteArrayDeserializer.class.getCanonicalName();
```
celiakung: I'd recommend prepending 'DEFAULT_' to these variable names

Rakesh627: @celiakung Just did 👍

Rakesh627:
@Jyouping @celiakung Made the recommended change!

PR #637, opened 2019-08-22
Default partition strategy always sends data to single partition

Rakesh627:
- if `system.destination.identityPartitioningEnabled` is disabled, the default partitioning applied in
`KafkaTransportProvider` is always sending data to a single partition.

- Based on this piece of code `return new ProducerRecord<>(topicName, partition.get(), keyValue, payloadValue);` the keyValue here is `new byte[0]` when no partition key is provided and it always translates to `partition 9`. 

- I believe if user hasn't provided a partition key, we should make the partition key `null` to make Kafka do default partitioning.

Above fix will make partition key as null when user hasn't provided a partition key to make Kafka go with default partitioning.

### DIFF
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
@@ -118,8 +118,14 @@ public class KafkaTransportProvider implements TransportProvider {
     } else {
       // If the partition is not specified. We use the partitionKey as the key. Kafka will use the hash of that
       // to determine the partition. If partitionKey does not exist, use the key value.
-      keyValue = record.getPartitionKey().isPresent() ? record.getPartitionKey().get().getBytes() : keyValue;
-      return new ProducerRecord<>(topicName, keyValue, payloadValue);
+      ProducerRecord<byte[], byte[]> producerRecord = new ProducerRecord<>(topicName, payloadValue);
+
+      if (record.getPartitionKey().isPresent()) {
+        keyValue = record.getPartitionKey().get().getBytes();
+        producerRecord = new ProducerRecord<>(topicName, keyValue, payloadValue);
+      }
+
+      return producerRecord;
     }
   }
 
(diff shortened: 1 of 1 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -118,8 +118,14 @@ public KafkaTransportProvider(DatastreamTask datastreamTask, List<KafkaProducerW
     } else {
       // If the partition is not specified. We use the partitionKey as the key. Kafka will use the hash of that
       // to determine the partition. If partitionKey does not exist, use the key value.
-      keyValue = record.getPartitionKey().isPresent() ? record.getPartitionKey().get().getBytes() : keyValue;
-      return new ProducerRecord<>(topicName, keyValue, payloadValue);
+      ProducerRecord<byte[], byte[]> producerRecord = new ProducerRecord<>(topicName, payloadValue);
```
celiakung: To simplify the change, you could just modify the original code.

Old: 
keyValue = record.getPartitionKey().isPresent() ? record.getPartitionKey().get().getBytes() : keyValue;
New: 
keyValue = record.getPartitionKey().isPresent() ? record.getPartitionKey().get().getBytes() : null;

Or you can set byte[] keyValue = null on line 100.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -142,8 +148,11 @@ public void send(String destinationUri, DatastreamProducerRecord record, SendCal
           LOG.error(errorMessage, e);
           throw new DatastreamRuntimeException(errorMessage, e);
         }
+        // Update topic-specific metrics and aggregate metrics
+        int numBytes = outgoing.key() != null ? outgoing.key().length : 0 + outgoing.value().length;
```
celiakung: I think you need parentheses to get the right order of operations, otherwise numBytes will just be the key length if non-null and will not include the value length. 

int numBytes = (outgoing.key() != null ? outgoing.key().length : 0) + outgoing.value().length;

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -119,7 +119,7 @@ public KafkaTransportProvider(DatastreamTask datastreamTask, List<KafkaProducerW
       // If the partition is not specified. We use the partitionKey as the key. Kafka will use the hash of that
       // to determine the partition. If partitionKey does not exist, use the key value.
       keyValue = record.getPartitionKey().isPresent()
```
gauravrkulkarni: consider initializing keyValue to null on line 100.

PR #638, opened 2019-08-23
Remove topic list from KMM event metadata

celiakung:
In KafkaMirrorMakerConnectorTask, some metadata is added for every event, including the "kafka-source-cluster". Currently we are using the entire KafkaConnectionString as the value of this metadata, which can be very huge for large topic whitelists. This can cause OOMs because we include this for every single event processed.

This change only includes the Kafka broker address list for this metadata, rather than the full connection string with topic regex/list.

the contribution guidelines.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectionString.java
```diff
@@ -55,11 +55,19 @@ public boolean isSecure() {
     return _isSecure;
   }
 
-  @Override
-  public String toString() {
+  /**
+   * Returns a comma-separated list of the Kafka broker addresses.
+   * @return
```
ahmedahamid: nit: remove empty @return tag

PR #646, opened 2019-09-13
Fix the datastream state for multiple tasks, and the connector validation

Jyouping:
Fix the following issues

1) Correctly return the datastream state for all tasks under a single connector rather just return first of them. This will allow us to catch all the assignments correctly.

2) Fix an issue when validating if the connector should support the partition assignment. The original validation can only work on the leader but we should be able to validate in any instance which receives this request.

### DIFF
--- a/datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
+++ b/datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
@@ -50,6 +50,7 @@ import com.linkedin.datastream.server.Coordinator;
 import com.linkedin.datastream.server.DatastreamGroup;
 import com.linkedin.datastream.server.DatastreamGroupPartitionsMetadata;
 import com.linkedin.datastream.server.DatastreamProducerRecord;
+import com.linkedin.datastream.server.DatastreamTask;
 import com.linkedin.datastream.server.DatastreamTaskImpl;
 import com.linkedin.datastream.server.FlushlessEventProducerHandler;
 import com.linkedin.datastream.server.SourceBasedDeduper;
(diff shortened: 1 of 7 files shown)


datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
```diff
@@ -697,6 +698,44 @@ public void testChangeDatatstreamAssignment() throws Exception {
 
   }
 
+  @Test
+  public void testGetStateForMultipleTasks() throws Exception {
+    String yummyTopic = "YummyPizza";
```
ahmedahamid: nit: you can just store these strings in an array to reduce repetition:
```java
    String[] topics = {"YummyPizza", "SaltyPizza", "HealthySalad"};
    Arrays.stream(topics).forEach(t -> createTopic(_zkUtils, t));
```
or in a list
```java
    List<String> topics = Arrays.asList("YummyPizza", "SaltyPizza", "HealthySalad");
    topics.forEach(t -> createTopic(_zkUtils, t));
```

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
```diff
@@ -697,6 +698,44 @@ public void testChangeDatatstreamAssignment() throws Exception {
...
+
+    createTopic(_zkUtils, saladTopic);
+    createTopic(_zkUtils, yummyTopic);
+    createTopic(_zkUtils, saltyTopic);
+
+    // create a datastream to consume from topics ending in "Pizza"
+    Datastream datastream = KafkaMirrorMakerConnectorTestUtils.createDatastream("pizzaStream", _broker, "\\w+Pizza");
+
+    // create three tasks for this datastream
+    List<DatastreamTask> tasks = ImmutableList.of(new DatastreamTaskImpl(Collections.singletonList(datastream)),
```
ahmedahamid: This can be written more succinctly as:
```java
    DatastreamTaskImpl dst = new DatastreamTaskImpl(Collections.singletonList(datastream));
    List<DatastreamTask> tasks = Collections.nCopies(3, dst);
```
However, I think you intended to create three distinct objects instead. Even with multiple instantiations, `DatastreamTaskImpl::hashCode()` will return the same value for all three objects. This means the connector won't be really maintaining three independent DSTs/CTs but only a single one. You might need to alter the `id` on each DSTImpl for instance.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
```diff
@@ -697,6 +698,44 @@ public void testChangeDatatstreamAssignment() throws Exception {
...
+    tasks.forEach(t -> ((DatastreamTaskImpl) t).setEventProducer(datastreamProducer));
+
+    KafkaMirrorMakerConnector connector =
+        new KafkaMirrorMakerConnector("MirrorMakerConnector", getDefaultConfig(Optional.empty()), "testCluster");
+    connector.start(null);
+
+    // notify connector of new task
+    connector.onAssignmentChange(tasks);
+
+    boolean pollResult = PollUtils.poll(() -> {
```
ahmedahamid: nit: rename `pollResult` (e.g. `hasExpectedTopicPartitions`)

datastream-server-api/src/main/java/com/linkedin/datastream/server/api/connector/DatastreamChangeListener.java
```diff
@@ -48,4 +48,12 @@ default void handleDatastream(List<DatastreamGroup> datastreamGroups) {
   default Map<String, Optional<DatastreamGroupPartitionsMetadata>> getDatastreamPartitions() {
     return new HashMap<>();
   }
+
+  /**
+   * Get if the partition management is supported. Return true if the connector can rely on Brooklin to management
```
ahmedahamid: I think we need to improve and expand this javadoc a bit. This is a fairly critical feature and it deserves more explanation.

somandal: +1

Jyouping: Thanks for the suggestion. Will add some clarification.

PR #653, opened 2019-09-20
Add custom checkpoint datastream metadata field to override connector level custom checkpoint

somandal:
Add a custom checkpoint datastream metadata field which can be used to override
the connector level custom checkpoint flag.


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -621,11 +621,26 @@ private void uninitializeTask(DatastreamTask t) {
     _cpProvider.unassignDatastreamTask(t);
   }
 
+  private boolean getCustomCheckpointing(DatastreamTask task) {
```
ahmedahamid: nit: might be better to keep `uninitializeTask` and `initializeTask` adjacent.

somandal: Fixed this.

PR #654, opened 2019-09-20
Fix a race condition in retrying the partition assignment

Jyouping:
Fix the followings
1. a race condition in retrying partition assignment when multiple datastreams were scheduled to retry at the same time
2. remove a quick path to skip lock checking for dummy tasks, which may potentially introduce race conditions

### DIFF
--- a/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
+++ b/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
@@ -665,7 +665,7 @@ public class TestCoordinator {
 
     waitTillAssignmentIsComplete(8, WAIT_TIMEOUT_MS, connectors.toArray(new TestHookConnector[connectors.size()]));
 
-    final long interval = Math.min(WAIT_TIMEOUT_MS, 100);
+    final long interval = Math.min(WAIT_TIMEOUT_MS, 500);
     Map<String, List<String>> assignment = collectDatastreamPartitions(connectors);
 
     Assert.assertTrue(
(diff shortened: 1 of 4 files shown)


datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -712,6 +712,12 @@ public void handleDatastream(List<DatastreamGroup> datastreamGroup) {
         }
       }
 
+      @Override
+      public void onAssignmentChange(List<DatastreamTask> tasks) {
+        _tasks.forEach(t -> t.release());
+        _tasks = tasks;
+        _tasks.forEach(t -> t.acquire(Duration.ofSeconds(10)));
+      }
```
somandal: nit: add a space after this function

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -848,8 +848,7 @@ public void cleanupOldUnusedTasks(Map<String, Set<DatastreamTask>> previousAssig
         previousAssignmentByInstance.values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
     List<DatastreamTask> unusedTasks =
         oldTasks.stream().filter(x -> !newTasks.contains(x)).collect(Collectors.toList());
-    LOG.warn("Deleting the unused tasks {} found between previous {} and new assignment {}. ", unusedTasks,
-        previousAssignmentByInstance, newAssignmentsByInstance);
+    LOG.info("Deleting the unused tasks {} ");
```
somandal: nit: it looks like you want to print a value, but you aren't passing anything to print in the log. Also why change it to warn?

Jyouping: I will add unusedTask here. This is regular cleaning up process in every assignment, the log level should be info instead of warn

PR #656, opened 2019-09-21
Relax Kafka broker hostname validation checks

ahmedahamid:
Relax Kafka broker validation checks that required hostnames to
be either IPv4, IPv6, or RFC 1123-compliant. Before it got removed,
this check was performed only for brokers Brooklin consumed from
but not those it produced to. This check is not necessary since
it causes inconvenience to customers using nonconforming broker
addresses, let alone Kafka clients verify broker addresses through 
DNS resolution anyways.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBrokerAddress.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBrokerAddress.java
@@ -8,21 +8,13 @@ package com.linkedin.datastream.connectors.kafka;
 import java.util.Comparator;
 import java.util.Objects;
 
-import org.apache.commons.validator.routines.DomainValidator;
-import org.apache.commons.validator.routines.InetAddressValidator;
-
 
 /**
  * The object representation of a Kafka Broker
  */
 public class KafkaBrokerAddress {
-  public static final Comparator<KafkaBrokerAddress> BY_URL = (o1, o2) -> {
-    int nameComparison = o1._hostName.compareTo(o2._hostName);
-    if (nameComparison != 0) {
-      return nameComparison;
-    }
-    return Integer.compare(o1._portNumber, o2._portNumber);
-  };
+  public static final Comparator<KafkaBrokerAddress> BY_URL =
+      Comparator.comparing((KafkaBrokerAddress o) -> o._hostName).thenComparingInt(o -> o._portNumber);
 
   private final String _hostName;
   private final int _portNumber;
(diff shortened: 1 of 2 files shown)


radai-rosenblatt:
what about other sources/sinks of data that brooklin supports, that might not validate?

ahmedahamid:
> what about other sources/sinks of data that brooklin supports, that might not validate?

This should have no effect on other sources/sinks since each of them has its own independent sets of validations.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBrokerAddress.java
```diff
@@ -112,16 +104,6 @@ private static String validateHostname(String hostName) throws IllegalArgumentEx
     if (trimmed.isEmpty()) {
       throw new IllegalArgumentException("empty host name");
     }
-    // we allow ipv4/6 addresses and RFC 1123 host names
```
gauravrkulkarni: If we remove this, what would be the final error we will see if the address format isn't correct by accident for a typical "confirming" address? Would it be just generic send error failures? If that's the case, instead of removing the check altogether, if possible, might want to add a setting to relax the check . Esp. if majority of the broker addresses are going to be confirming types, as this gives us a clear message about what's wrong.

ahmedahamid: I agree. 

However, the problem is that he call chain is as follows:
`AbstractKafkaConnector`
  → `KafkaConnectionString::valueOf()`
  → `KafkaConnectionString::parseBrokers()`
  → `KafkaBrokerAddress::valueOf()`
  → `KafkaBrokerAddress::validateHostname()`

If we introduce a flag to toggle hostname validation, we'd have to wire it through this whole chain.

ahmedahamid: @gauravrkulkarni 
Another key thing I forgot to mention: removing this validation has not impact on reliability because even if a hostname is IPv4/6 or RFC 1123 compliant, there's no guarantee that it exists or is resolvable through DNS. Even with this validation, we can still create datastreams against perfectly valid hostnames that simply do not exist. However, we cannot create datastreams against hostnames that do exist just because their names aren't IPv4/6 or RFC 1123 compliant.

gauravrkulkarni: If its too much of a hassle to add a config and if the merits aren't that many, its ok to not have config.

PR #657, opened 2019-09-24
Invoke topic manger when the partitions are assigned

Jyouping:
As there will be no partition assignment callback triggered with low level, we will have to call topicManager explicitly when we created the task assignment. As we don't revoke the partitions but terminate the entire tasks, we dont need to call topicManger on revoke.

the contribution guidelines.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -200,6 +200,10 @@ protected void consumerSubscribe() {
       // Consumer can be assigned with an empty set, but it cannot run poll against it
       // While it's allowed to assign an empty set here, the check on poll need to be performed
       _consumer.assign(_consumerAssignment);
+
+      // Invoke topic manager
```
gauravrkulkarni: Might want to create a separate method for this and call that method? same code is being called from "onPartitionsAssigned" also

Jyouping: consolidated them into handleTopicMangerPartitionAssignment

PR #661, opened 2019-10-02
Add warning log for events having an event latency larger than a configurable threshold

somandal:


datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -249,6 +256,11 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
       reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdAlternateSlaMs,
           EVENTS_PRODUCED_WITHIN_ALTERNATE_SLA, EVENTS_PRODUCED_OUTSIDE_ALTERNATE_SLA);
 
+      if (sourceToDestinationLatencyMs > _warnLogLatencyThresholdMs) {
+        _logger.warn("Source to destination latency {} ms is higher than {} ms, Source Timestamp: {}, Metadata: {}",
```
gauravrkulkarni: We will end up logging this warning for every message. Might want to guard it with _logger.isDebugEnabled() if this is for debugging purpose only?  Also, looks like we are emitting a metric for this already.

somandal: @gauravrkulkarni I've kept the _warnLogLatencyThresholdMs at 100000 minutes, which is more than a day. The metrics are emitting 2 GB for this value, so we do want to log this data to find out what is causing such high latency.

PR #662, opened 2019-10-03
Initiate the Kafka position tracker when the task is created

Jyouping:
Fix an issue where Kafka position tracker is not triggered in a Brooklin-managed partition assignment


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -682,13 +682,15 @@ public void onPartitionsRevoked(Collection<TopicPartition> topicPartitions) {
 
   @Override
   public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
+    this.onPartitionsAssignedInternal(partitions);
```
ahmedahamid: nit: remove `this`?

PR #668, opened 2019-11-07
Add exception to warning log message on send failures for Kafka connectors

somandal:
On send failures, the exception is not logged. Adding the exception to the log.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -233,7 +233,7 @@ protected void translateAndSendBatch(ConsumerRecords<?, ?> records, Instant read
             sendDatastreamProducerRecord(datastreamProducerRecord, topicPartition, numBytes, null);
           }
         } catch (Exception e) {
-          _logger.warn("Got exception while sending record {}", record);
+          _logger.warn("Got exception while sending record {}, exception: ", record, e);
```
ahmedahamid: Agreed with @somandal to format record and use `_logger.warn(String, Throwable)` throughout this file.

PR #671, opened 2019-11-20
Ensure existing ZK session closed before opening a new one

akaritakai:
It looks like `ZkAdapter.joinLeaderElection()` calls `connect()` without an accompanying call to `disconnect()` when it can't find its node name in the ZK liveinstances path. The assumption is that the existing ZK session is dead and unrecoverable.

In practice, this does not seem to always be the case, and the consequence of calling `connect()` before `disconnect()` means that we leak a ZK session which hold onto ephemeral nodes and poison the ZK state.

A permanent fix to this issue is to guard `connect()` with `disconnect()`, which is the change this PR makes.

the contribution guidelines.


thomaslaw:
_lgtm_

PR #678, opened 2020-01-28
fix zk session leak

sdomalap:
This PR has a bug fix to a specific problem where the cluster could eventually go leaderless as leadership transitions occur. It handles the session re-establishment properly and avoid session leaks.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
@@ -20,14 +20,17 @@ import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
+import org.I0Itec.zkclient.IZkStateListener;
 import org.I0Itec.zkclient.exception.ZkException;
 import org.I0Itec.zkclient.exception.ZkNoNodeException;
 import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.Watcher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -201,9 +207,15 @@ public void disconnect() {
    * the actions that need to be taken with them, which are implemented in the Coordinator class
    */
   public void connect() {
-    disconnect(); // Guard against leaking an existing zookeeper session
```
vmaheshw: Removing this will not be sufficient. On Zookeeper session expiry, the live instance ephemeral node will be deleted. So, the leader will become follower and both needs to creates new live instance nodes. Once the role will change, all the listener cleanup will be required which is happening in disconnect.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1252,6 +1273,25 @@ public void handleDataDeleted(String dataPath) throws Exception {
     }
   }
 
+  /**
+   * Listener for ZooKeeper state changes.
+   */
+  public class ZkStateChangeListener implements IZkStateListener {
+    @Override
+    public void handleStateChanged(Watcher.Event.KeeperState state) {
+    }
```
vmaheshw: In the situation, when the session never gets reestablished, handleNewSession will not get invoked. So, handleStateChanged with state EXPIRY needs to be handled.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
sdomalap: Thanks for the review @vmaheshw.

It seems like you are correct. I might be leaking subscriptions. I may not have experienced any issues as the subscriptions are to the nodes that would never get updated. 

I will add the clean up logic as necessary.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
sdomalap: Noted. Will post a new commit. Thanks

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
akaritakai: Why remove this line anyway? If we are connecting or reconnecting, then we should not have an existing session. The guard code defensively ensures that any existing connection is definitely closed.

PR #680, opened 2020-02-06
Add support destination topic prefix, to allow mirroring source topics to destination topics prepended with a prefix

somandal:
Currently BMM expects the source and destination topics to have the same name for mirroring. This feature allows mirroring source topics to a destination topic prepended with a cluster level prefix.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnector.java
@@ -19,6 +19,7 @@ import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections.ListUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.PartitionInfo;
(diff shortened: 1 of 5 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnector.java
```diff
@@ -61,15 +62,20 @@
   // This config controls how frequent the connector fetches the partition information from Kafka in order to perform
   // partition assignment
   protected static final String PARTITION_FETCH_INTERVAL = "PartitionFetchIntervalMs";
+  // This config, if set to a non-blank value, prepends a prefix to the destination topic name for all topics
+  // being mirrored by this cluster. E.g. source topic: foo, destination prefix: bar, destination topic: barfoo
+  protected static final String DESTINATION_TOPIC_PREFIX = "destinationTopicPrefix";
```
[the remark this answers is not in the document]
somandal: done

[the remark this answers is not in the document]
gauravrkulkarni: I guess people can forget about cluster level config and create a stream without realizing about the config? I will let you decide about cluster level config though, I am OK either ways.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -139,12 +142,19 @@ public KafkaMirrorMakerConnectorTask(KafkaBasedConnectorConfig config, Datastrea
...
     _enablePartitionAssignment = config.getEnablePartitionAssignment();
+    _destinationTopicPrefix = task.getDatastreams().get(0).getMetadata()
+        .getOrDefault(DatastreamMetadataConstants.DESTINATION_TOPIC_PREFIX, destinationTopicPrefix);
     _dynamicMetricsManager = DynamicMetricsManager.getInstance();
 
     if (_enablePartitionAssignment) {
       LOG.info("Enable Brooklin partition assignment");
     }
 
+    if (!StringUtils.isBlank(_destinationTopicPrefix)) {
```
gauravrkulkarni: nit: might want to log the prefix anyway in case we need to debug this?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnector.java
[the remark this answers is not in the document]
somandal: I see how a config can be forgotten and then bad things happen. I've removed the cluster level config.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
somandal: done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -112,6 +116,7 @@
   private final boolean _isFlushlessModeEnabled;
   private final boolean _isIdentityMirroringEnabled;
   private final boolean _enablePartitionAssignment;
+  private final String _destinationTopicPrefix;
```
celiakung: Suggest to wrap in Optional so that you also don't need to add empty string "" to topicManager properties if doesn't exist.

somandal: Intellij seems to warn about using optionals as class members:

Reports any uses of java.util.Optional<T>, java.util.OptionalDouble, java.util.OptionalInt, java.util.OptionalLong or com.google.common.base.Optional as the type for a field or a parameter. Optional was designed to provide a limited mechanism for library method return types where there needed to be a clear way to represent "no result". Using a field with type java.util.Optional is also problematic if the class needs to be Serializable, which java.util.Optional is not.

So I'll leave this part as is for now.

PR #683, opened 2020-02-10
Remove KafkaPositonTracker and its usages

vishwajith-s:
KafkaPositionTracker has caused performance issues with Brooklin.
Hence, we have decided to get rid of KafkaPositionTracker. Instead
of reporting broker and consumer position offset, the Brooklin endpoint
will now report consumer groups, topics and partitions. These can be
used to query broker and consumer offsets from Kafka. This diff gets
rid of KafkaPositionTracker and its usages. There will be a separate
diff for adding the above support.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -96,7 +94,6 @@ public Thread newThread(@NotNull Runnable r) {
 
   enum DiagnosticsRequestType {
     DATASTREAM_STATE,
```
gauravrkulkarni: nit: remove the comma

PR #684, opened 2020-02-11
Spawn threads to stop tasks during the onAssignmentChange() code path  in AbstractKafkaConnector

somandal:
Currently in onAssignmentChange(), we mark removed tasks for shutdown and then let them stop naturally. This can lead to situations where the task is taking too long or stuck indefinitely (e.g. producer.flush() can get stuck). Such tasks are problematic. Instead, we should try to force stop these tasks. onAssignmentChange() is not expected to block for long by the Coordinator, due to which a separate thread should be spawned to deal with stopping tasks.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
@@ -46,6 +47,7 @@ import com.linkedin.datastream.common.DatastreamSource;
 import com.linkedin.datastream.common.DatastreamUtils;
 import com.linkedin.datastream.common.DiagnosticsAware;
 import com.linkedin.datastream.common.JsonUtils;
+import com.linkedin.datastream.common.ThreadUtils;
 import com.linkedin.datastream.server.DatastreamTask;
 import com.linkedin.datastream.server.api.connector.Connector;
 import com.linkedin.datastream.server.api.connector.DatastreamValidationException;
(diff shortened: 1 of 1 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -142,6 +151,7 @@ public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
           // Make sure to replace the DatastreamTask with most up to date info
           // This is necessary because DatastreamTaskImpl.hashCode() does not take into account all the
           // fields/properties of the DatastreamTask (e.g. dependencies).
+          _runningTasks.remove(task);
```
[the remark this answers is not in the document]
somandal: This is not a problem. We are not iterating through the map that we are removing things from. We are iterating over the tasks list, which we don't modify at all.

      **for (DatastreamTask task : tasks) {**
        ConnectorTaskEntry connectorTaskEntry = _runningTasks.get(task);
        if (connectorTaskEntry != null) {
          AbstractKafkaBasedConnectorTask kafkaBasedConnectorTask = connectorTaskEntry.getConnectorTask();
          kafkaBasedConnectorTask.checkForUpdateTask(task);
          // Make sure to replace the DatastreamTask with most up to date info
          // This is necessary because DatastreamTaskImpl.hashCode() does not take into account all the
          // fields/properties of the DatastreamTask (e.g. dependencies).
          **_runningTasks.remove(task);**
          _runningTasks.put(task, connectorTaskEntry);
          continue; // already running
        }

        _runningTasks.put(task, createKafkaConnectorTask(task));
      }

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -128,10 +133,14 @@ public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
 
       for (DatastreamTask task : toCancel) {
         ConnectorTaskEntry connectorTaskEntry = _runningTasks.remove(task);
-        // Stopping the connectorTask. This only marks the connector task as shutdown and does not actually wait for
-        // the connector task to stop. onAssignmentChange() must be completed quickly, otherwise the Coordinator
-        // kills the assignment threads.
-        connectorTaskEntry.getConnectorTask().stop();
+        // Spawn a separate thread to attempt stopping the connectorTask. The connectorTask will be canceled if it
+        // does not stop within a certain amount of time. This is force cleanup of connectorTasks which take too long
```
DEEPTHIKORAT: typo: This will force cleanup

somandal: done

somandal:
Currently in onAssignmentChange(), we mark removed tasks for shutdown and then let them stop naturally. This can lead to situations where the task is taking too long or stuck indefinitely (e.g. producer.flush() can get stuck). Such tasks are problematic. Instead, we should try to force stop these tasks. onAssignmentChange() is not expected to block for long by the Coordinator, due to which a separate thread should be spawned to deal with stopping tasks.

PR #685, opened 2020-02-14
 Verify/Clean orphan tasks under connector in zookeeper by leader coordinator

vmaheshw:
the contribution guidelines.

This PR addresses the verification/cleanup of orphan tasks under connector in ZooKeeper which is not assigned to any instance. Any exception in Coordinator during CleanupOfOldInstances and unused tasks can result in orphan nodes under connector. In this PR, the leader coordinator after becoming leader and finishing the assignment process for the first time, will verify if there is any orphan tasks and if there is a config to clean them, it will delete it from the zookeeper.


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -875,6 +890,44 @@ public void cleanupOldUnusedTasks(Map<String, Set<DatastreamTask>> previousAssig
     unusedTasks.forEach(t -> deleteConnectorTask(t.getConnectorType(), t.getDatastreamTaskName()));
   }
 
+  /**
+   * Remove orphan connector task nodes which are not assigned to any instance (live or pause).
+   *
+   * NOTE: this should only be called after the valid tasks have been
+   * reassigned or become safe to discard per strategy requirement.
+   * This can be a costly operation. So, it should be call once the leader get elected and has finished
+   * the assignment and cleaned up dead Tasks. Ideally, we should not find anything in this check to clean up.
```
somandal: Just a clarifying question: Any reason this shouldn't be done as part of cleaning up dead tasks in the original LEADER_DO_ASSIGNMENT event? Might be a good idea to add clear documentation around why this should only be performed afterwards.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1222,9 +1275,10 @@ public ZkBackedLiveInstanceListProvider() {
       List<String> liveInstances = new ArrayList<>();
       for (String n : nodes) {
         String hostname = _zkclient.ensureReadData(KeyBuilder.liveInstance(_cluster, n));
-        if (hostname != null) {
+        if (hostname == null) {
```
DEEPTHIKORAT: I remember reading this at some point and thinking I am going to see this in this the logs somewhere and assume something bad happened, and that exactly happened yesterday. Awesome that you fixed it.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -875,6 +890,44 @@ public void cleanupOldUnusedTasks(Map<String, Set<DatastreamTask>> previousAssig
     unusedTasks.forEach(t -> deleteConnectorTask(t.getConnectorType(), t.getDatastreamTaskName()));
   }
 
+  /**
+   * Remove orphan connector task nodes which are not assigned to any instance (live or pause).
```
DEEPTHIKORAT: typo: live or *paused*

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -875,6 +890,44 @@ public void cleanupOldUnusedTasks(Map<String, Set<DatastreamTask>> previousAssig
...
+   *
+   * NOTE: this should only be called after the valid tasks have been
```
DEEPTHIKORAT: nit: Can you format the comment lines to be of approximately the same lenght?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -875,6 +890,44 @@ public void cleanupOldUnusedTasks(Map<String, Set<DatastreamTask>> previousAssig
...
+   * reassigned or become safe to discard per strategy requirement.
+   * This can be a costly operation. So, it should be call once the leader get elected and has finished
```
DEEPTHIKORAT: nit: typos:
1. should be called; 2. leader gets elected

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -875,6 +890,44 @@ public void cleanupOldUnusedTasks(Map<String, Set<DatastreamTask>> previousAssig
...
+    for (String connector : allConnectors) {
+      Set<String> connectorTaskList = getConnectorTasks(connector);
+
+      connectorTaskList.removeAll(validTasks.stream()
+          .filter(x -> x.getConnectorType().equals(connector))
+          .map(DatastreamTask::getDatastreamTaskName)
+          .collect(Collectors.toSet()));
+
+      if (connectorTaskList.size() > 0) {
+        LOG.warn("Found orphan tasks: {} in connector: {}", connectorTaskList, connector);
```
DEEPTHIKORAT: Can we add a metric for this so we can monitor it?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
vmaheshw: Updated the comment as well. get all children of /cluster/connector is a very costly operation. So, it should not be done very frequently.

datastream-server/src/test/java/com/linkedin/datastream/server/zk/TestZkAdapter.java
```diff
@@ -759,6 +759,26 @@ public void testDeleteTasksWithPrefix() {
...
+
+    leftOverTasks = zkClient.getChildren(KeyBuilder.connector(testCluster, connectorType));
+    Assert.assertEquals(leftOverTasks.size(), 2);
+
+    updateInstanceAssignment(adapter, adapter.getInstanceName(), new ArrayList<DatastreamTask>());
+
+    leftOverTasks = zkClient.getChildren(KeyBuilder.connector(testCluster, connectorType));
+    Assert.assertEquals(leftOverTasks.size(), 2);
+
+    adapter.cleanUpOrphanConnectorTasks(true);
```
DEEPTHIKORAT: should we test with false first after updateInstanceAssignments and see 2 left as well?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -722,6 +732,10 @@ protected synchronized void handleEvent(CoordinatorEvent event) {
           performPartitionMovement((Long) event.getEventMetadata());
           break;
 
+        case LEADER_DO_POST_BECOMING_LEADER:
```
DEEPTHIKORAT: It's really hard to come up with a name for this.. not happy with this because it suggests action with DO but doesnt specify the action. I am unable to come up with any better though. Will post if I do.

datastream-server/src/test/java/com/linkedin/datastream/server/zk/TestZkAdapter.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
somandal: Shall we call this something like : POST_BECOME_LEADER instead? I don't like the existing name at all

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1243,6 +1257,13 @@ private void performPartitionMovement(Long notifyTimestamp) {
     return newAssignmentsByInstance;
   }
 
+  void performTaskPostBecomingLeader() {
+    _log.info("performTaskPostBecomingLeader called");
+    int orphanCount = _adapter.cleanUpOrphanConnectorTasks(_config.getZkCleanUpOrphanConnectorTask());
+    _dynamicMetricsManager.createOrUpdateMeter(MODULE, "performTaskPostBecomingLeader",
```
somandal: If you plan to rename "performTaskPostBecomingLeader", can you also add a static constant for this string literal?

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -2443,6 +2450,115 @@ public void testDatastreamDedupeMetadataCopy() throws Exception {
...
+    instance1.start();
+
+    String isLeaderMetricName = "Coordinator.isLeader";
+    Gauge<Integer> isLeader = DynamicMetricsManager.getInstance().getMetric(isLeaderMetricName);
+
+    Assert.assertEquals(isLeader.getValue().intValue(), 1);
+
+    //
+    // create datastream definitions under /testAssignmentBasic/datastream/datastream1
+    //
```
somandal: nit: can we make the comment style consistent in this function? some comments have an extra // before and after, other's don't. And it is not exactly clear if this is only because of differentiating steps? If so, maybe number the steps instead to make it clear that those are step related comments?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: I completely agree with both of you. @somandal POST_BECOME_LEADER looks incomplete and is not aligning with the other 3 leader specific event names. How about LEADER_DO_POST_ELECTION ?

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
vmaheshw: I have followed the comment style used in TestCoordinator file.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: Please correct me if I understand something wrong. This pattern is followed in the entire file for  createOrUpdateMeter.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
somandal: Just an FYI, the event to handle datastream add/drop is also a leader specific event, so I wouldn't read too much into the "LEADER" naming scheme. I also noticed that the names of events talk about the operation performed. Maybe we should name it along those lines as well? LEADER_DO_POST_ELECTION_CLEANUP or something.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
somandal: Just thought it'll be cleaner to have this as a const separately, but sure if you've done it this way for consistency it's your call. I would generally prefer making all string literals constants, but that is just me.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
somandal: Just an FYI, the event to handle datastream add/drop is also a leader specific event, so I wouldn't read too much into the "LEADER" naming scheme. I also noticed that the names of events talk about the operation performed. Maybe we should name it along those lines as well? LEADER_DO_POST_ELECTION_CLEANUP or something.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
somandal: Sure, if you think consistency is more important then leave it as is. In general though, this style of commenting is out of the ordinary from the rest of the code base

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: How about LEADER_DO_CLEANUP_POST_ELECTION?  It sounds more readable.

somandal: Sounds great!

PR #686, opened 2020-02-19
Add diag support to report consumer groups and topic partitions

vishwajith-s:
This patch adds diag support to report consumer groups and topic
partitions in each consumer group.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
@@ -11,6 +11,7 @@ import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -26,11 +27,13 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.utils.URLEncodedUtils;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.codehaus.jackson.type.TypeReference;
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 
@@ -94,6 +97,7 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
 
   enum DiagnosticsRequestType {
     DATASTREAM_STATE,
+    PARTITIONS,
   }
 
   /**
@@ -382,6 +386,10 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
         String response = processDatastreamStateRequest(uri);
         _logger.trace("Query: {} returns response: {}", query, response);
         return response;
+      } else if (path != null && path.equalsIgnoreCase(DiagnosticsRequestType.PARTITIONS.toString())) {
+        String response = processTopicPartitionStatsRequest();
+        _logger.trace("Query: {} returns response: {}", query, response);
+        return response;
       } else {
         _logger.warn("Could not process query {} with path {}", query, path);
       }
@@ -428,6 +436,25 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
         .map(KafkaDatastreamStatesResponse::toJson).orElse(null);
   }
 
+  private String processTopicPartitionStatsRequest() {
+    _logger.info("process topic partitions stats request");
+    List<KafkaTopicPartitionStatsResponse> serializedResponses = new ArrayList<>();
+
+    synchronized (_runningTasks) {
+      _runningTasks.forEach((dataStreamTask, connectorTaskEntry) -> {
+        List<Datastream> datastreams = dataStreamTask.getDatastreams();
+        Set<String> datastreamNames = new HashSet<>();
+        datastreams.forEach(datastream -> datastreamNames.add(datastream.getName()));
+
+        KafkaTopicPartitionTracker tracker = connectorTaskEntry.getConnectorTask().getKafkaTopicPartitionTracker();
+        KafkaTopicPartitionStatsResponse response = new KafkaTopicPartitionStatsResponse(tracker.getConsumerGroupId(),
+          tracker.getTopicPartitions(), datastreamNames);
+        serializedResponses.add(response);
+      });
+    }
+    return JsonUtils.toJson(serializedResponses);
+  }
+
   /**
    * Aggregates the responses from all the instances into a single JSON response.
    * Sample query: /datastream_state?datastream=PizzaDatastream
@@ -454,6 +481,10 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
           && (path.equalsIgnoreCase(DiagnosticsRequestType.DATASTREAM_STATE.toString()))) {
         return JsonUtils.toJson(responses);
       }
+      if (path != null
+          && (path.equalsIgnoreCase(DiagnosticsRequestType.PARTITIONS.toString()))) {
+        return reduceTopicPartitionStatsResponses(responses);
+      }
     } catch (Exception e) {
       _logger.warn("Failed to reduce responses from query {}: {}", query, e.getMessage());
       _logger.debug(String.format("Failed to reduce responses from query %s: %s", query, e.getMessage()), e);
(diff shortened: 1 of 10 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionStatsResponse.java
```diff
@@ -0,0 +1,115 @@
...
+  @JsonSetter("_topicPartitions")
+  public void setTopicPartitions(Set<TopicPartition> topicPartitions) {
+    _topicPartitions = topicPartitions;
+  }
+
+  /**
+   * Serialize to JSON
+   */
+  public static String toJson(List<KafkaTopicPartitionStatsResponse> obj) {
+    ObjectMapper mapper = new ObjectMapper();
```
ahmedahamid: Could you, please, check if you can reuse this object for serialization and deserialization (e.g. declare as a `static final`)? I don't think we need to instantiate and configure one every time this method is invoked. Same goes for `fromJson()`.

datastream-common/src/main/java/com/linkedin/datastream/common/JsonUtils.java
```diff
@@ -96,6 +96,46 @@
...
+    Validate.notNull(object, "null input object");
+    String out = null;
+    try {
+      out = mapper.writeValueAsString(object);
+    } catch (IOException e) {
+      String errorMessage = "Failed to serialize object: " + object;
+      ErrorLogger.logAndThrowDatastreamRuntimeException(LOG, errorMessage, e);
+    }
+    return out;
+  }
```
ahmedahamid: Could you, please, have the existing `fromJson()` and `toJson()` invoke these methods after reconciling all differences, e.g.

```java
  public static <T> T fromJson(String json, TypeReference<T> typeRef) {
    return fromJson(json, typeRef, MAPPER);
  }

  public static <T> T fromJson(String json, TypeReference<T> typeRef, ObjectMapper mapper) {
    Validate.notNull(json, "null JSON string");
    Validate.notNull(typeRef, "null type reference");
    T object = null;
    try {
      object = mapper.readValue(json, typeRef);
    } catch (IOException e) {
      String errorMessage = "Failed to parse json: " + json;
      ErrorLogger.logAndThrowDatastreamRuntimeException(LOG, errorMessage, e);
    }
    return object;
  }

  public static <T> String toJson(T object) {
    return toJson(object, MAPPER);
  }

  public static <T> String toJson(T object, ObjectMapper mapper) {
    Validate.notNull(object, "null input object");
    StringWriter out = new StringWriter();   // Modified to match the old toJson()
    try {
      mapper.writeValue(out, object);        // ditto
    } catch (IOException e) {
      String errorMessage = "Failed to serialize object: " + object;
      ErrorLogger.logAndThrowDatastreamRuntimeException(LOG, errorMessage, e);
    }
    return out.toString();
  }
```

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionStatsResponse.java
```diff
@@ -0,0 +1,115 @@
...
+   * Constructor for KafkaTopicPartitionStatsResponse
+   * @param consumerGroupId identifier for consumer group
+   * @param topicPartitions a map of consumer offsets for topic partitions
+   */
+  public KafkaTopicPartitionStatsResponse(String consumerGroupId, Set<TopicPartition> topicPartitions) {
+    _consumerGroupId = consumerGroupId;
+    _topicPartitions = topicPartitions;
+  }
+
+  @JsonGetter("_consumerGroupId")
```
ahmedahamid: With the auto-field discovery and the custom deserializer, could you please double-check if you need these getter/setter annotations? I have a somewhat strong suspicion we don't really need them.

I think we can do w/o `@JsonGetter`, `@JsonSetter`, and the default ctor in favor of the newer `@JsonCreator` and `@JsonProperty` annotations.
```java
  @JsonCreator
  public KafkaTopicPartitionStatsResponse(
      @JsonProperty("consumerGroupId") String consumerGroupId,
      @JsonProperty("topicPartitions") Set<TopicPartition> topicPartitions) {
    ...
  }
```

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaTopicPartitionStats.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  boolean testProcessTopicPartitionsStatsInternal(KafkaMirrorMakerConnector connector, Set<TopicPartition> expected) {
+    String jsonStr = connector.process("/topicpartition_stats");
+    List<KafkaTopicPartitionStatsResponse> responseList = KafkaTopicPartitionStatsResponse.fromJson(jsonStr);
+
+    Set<TopicPartition> actual = new HashSet<>();
+    responseList.forEach(response -> {
+      actual.addAll(response.getTopicPartitions());
+    });
```
ahmedahamid: nit: You don't need parenthesis for one liner lambdas
```java
    responseList.forEach(response -> actual.addAll(response.getTopicPartitions()));
```

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaTopicPartitionStats.java
```diff
@@ -0,0 +1,148 @@
...
+    // Wait until the partitions from the deleted topic are revoked
+    if (!PollUtils.poll(() -> testProcessTopicPartitionsStatsInternal(connector, expected),
+        KafkaMirrorMakerConnectorTestUtils.POLL_PERIOD_MS,
+        KafkaMirrorMakerConnectorTestUtils.POLL_TIMEOUT_MS)) {
+      Assert.fail("Topic partitions still not revoked");
+    }
+    connector.stop();
+  }
+
+  boolean testProcessTopicPartitionsStatsInternal(KafkaMirrorMakerConnector connector, Set<TopicPartition> expected) {
```
ahmedahamid: `private`

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaTopicPartitionStats.java
```diff
@@ -0,0 +1,148 @@
...
+    String jsonStr = connector.process("/topicpartition_stats");
```
ahmedahamid: Replace w/ `TOPICPARTITION_STATS_QUERY`

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -463,6 +487,38 @@ public String reduce(String query, Map<String, String> responses) {
...
+        responseList = KafkaTopicPartitionStatsResponse.fromJson(json);
+      } catch (Exception e) {
+        _logger.error("Invalid response {} from instance {}", json, instance);
+        return;
+      }
+      if (responseList.size() >= 0) {
+        responseList.forEach(response -> {
+          if (response.getTopicPartitions() == null || response.getConsumerGroupId().isEmpty()) {
+            _logger.warn("Empty topic partition stats map from instance {}. Ignoring the result", instance);
+            return;
```
gauravrkulkarni: should we throw exception here so that caller of the endpoint will know about it?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -463,6 +487,38 @@ public String reduce(String query, Map<String, String> responses) {
     return null;
   }
 
+  private String reduceTopicPartitionStatsResponses(Map<String, String> responses) {
+    Map<String, Map<String, Set<Integer>>> result = new HashMap<>();
```
gauravrkulkarni: Is the endpoint invoked for a particular datastream? If so, what are the cases where we would expect more than one consumer group? 

If the response is supposed to include data for multiple unrelated datastreams, I think the response should include datastream name as well?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -463,6 +487,38 @@ public String reduce(String query, Map<String, String> responses) {
...
+
+    responses.forEach((instance, json) -> {
+      List<KafkaTopicPartitionStatsResponse> responseList;
+      try {
+        responseList = KafkaTopicPartitionStatsResponse.fromJson(json);
+      } catch (Exception e) {
+        _logger.error("Invalid response {} from instance {}", json, instance);
+        return;
+      }
+      if (responseList.size() >= 0) {
```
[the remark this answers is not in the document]
vishwajith-s: Done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionStatsResponse.java
vishwajith-s: I cannot make it as final. But I have made it static.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaTopicPartitionStats.java
vishwajith-s: Thanks! I will retain this as it makes it more clearer.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: I think ignoring the result and proceeding further is good here.  At least we will return a warning and proceed further with best effort. This is a diag tool.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: The end point is not invoked for a particular datastream. It just aggregates the consumer groups across all brooklin instances, datastreams.  

If the response is supposed to include data for multiple unrelated datastreams, I think the response should include datastream name as well?
I believe the intent here is to get enough info to query Kafka or lag monitor.  Since datastream is an internal brooklin concept it will not be reported.

datastream-common/src/main/java/com/linkedin/datastream/common/JsonUtils.java
vishwajith-s: Good point. I should have done this at the start. I cannot combine the ones that takes class and typereference. But the rest is doable.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
gauravrkulkarni: I believe datastream name will be useful, as that's a fundamental unit of operation for brooklin (for example, if one needs to restart a datstream to recover from lags).
Admittedly there are other ways to map datastream to consumer group and some type of group IDs even have datastream names in them, but that can mean going through all the datastreams in a cluster. 
We can discuss this offline and add the datastream in a separate PR if needed.

ahmedahamid:
@vishwajith-s Please, update the wiki pages to reflect the current status of the diag endpoint. In particular, we need to:
1. Remove all Kafka position tracker configs from the Kafka connectors [config section](https://github.com/linkedin/brooklin/wiki/Kafka-Connectors-Shared-Logic#configuration)
2. Update the endpoint query params and response data models in the [Diagnostics section](https://github.com/linkedin/brooklin/wiki/Kafka-Connectors-Shared-Logic#diagnostics)

PR #688, opened 2020-03-05
Fix keys getting discarded before sending the record to the destination

sdomalap:
The record's keys are missing in the destination Kafka topic. This pull request has the fix to make sure that the keys are intact in the producer record.

### DIFF
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
@@ -97,7 +97,7 @@ public class KafkaTransportProvider implements TransportProvider {
 
     Optional<Integer> partition = record.getPartition();
 
-    byte[] keyValue = new byte[0];
+    byte[] keyValue = null;
     byte[] payloadValue = new byte[0];
     if (event instanceof BrooklinEnvelope) {
       BrooklinEnvelope envelope = (BrooklinEnvelope) event;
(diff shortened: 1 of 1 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -119,7 +119,7 @@ public KafkaTransportProvider(DatastreamTask datastreamTask, List<KafkaProducerW
       // If the partition is not specified. We use the partitionKey as the key. Kafka will use the hash of that
       // to determine the partition. If partitionKey does not exist, use the key value.
       keyValue = record.getPartitionKey().isPresent()
-          ? record.getPartitionKey().get().getBytes(StandardCharsets.UTF_8) : null;
+              ? record.getPartitionKey().get().getBytes(StandardCharsets.UTF_8) : keyValue;
```
ahmedahamid: I understand you're trying to minimize changes but I think the code would be slightly clearer/better if we avoid the redundant assignment, i.e.

```java
if (record.getPartitionKey().isPresent()) {
    keyValue = record.getPartitionKey().get().getBytes(StandardCharsets.UTF_8);
}
```

sdomalap: Thanks for the review @ahmedahamid. I agree with your feedback. 
I was trying to make the code look as it was prior to this [commit](https://github.com/linkedin/brooklin/commit/0230f364732d1cafc779c420108067e359fd88f7).

Let me know if you feel strongly about the code style. I can make that change and run a quick round of tests.

ahmedahamid: Thanks for your contribution, @sdomalap

PR #690, opened 2020-03-10
Fix zk cleanup test in TestCoordinator.java

vmaheshw:
Fix the flaky test case. It is failing because post clean up takes some time after becoming leader. 
Adding additional timeout in verify to make sure that cleanUpOrphanConnectorTasks is called within 30 sec of becoming leader. I have to upgrade junit to version4.x because mockito.timeout throws classNotFoundError. Mockito is built with junit V4.x.


datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -2555,7 +2556,7 @@ public void testCoordinatorLeaderCleanupTasksPostElection() throws Exception {
     instance1.stop();
     Assert.assertTrue(PollUtils.poll(() -> instance2.getIsLeader().getAsBoolean(), 100, 30000));
     verify(spyZkAdapter1, times(0)).cleanUpOrphanConnectorTasks(anyBoolean());
-    verify(spyZkAdapter2, times(1)).cleanUpOrphanConnectorTasks(anyBoolean());
+    verify(spyZkAdapter2, timeout(30000).times(1)).cleanUpOrphanConnectorTasks(anyBoolean());
```
somandal: Is there no way to do this using testng instead of adding a junit dependency?

vmaheshw: Updated the comment. Mockito uses junit underneath and is compiled using V4.X. We are getting junit transitively and is pointing to 3.8.1. So, it hits ClassNotFoundError.

somandal: got it.

vishwajith-s: Quick question. How do we make sure its called within 30 seconds? Also how did we come up with the number 30s? Will this not fail in the future due to timeout?

vmaheshw: After becoming leader, coordinator does assignment and then cleanup. Unit-tests are not heavy. So, the operations should finish within 30 sec. 30 sec is used for other operations in the tests as well. For reference, in intellij and gradlew this test finishes within 1 sec.

vishwajith-s: Cool! Interesting. I would prefer if we avoided depending on timeouts in unit tests.  It's fine for now. LGTM.

PR #691, opened 2020-03-11
Cleanup ZkOrphanTasks only when leader assignment is successful and fix CoordinatorEventBlockingQueue to handle duplicate events with metadata.

vmaheshw:
Cleanup ZkOrphanTasks only when leader assignment is successful and fix Coordinator queue to handle duplicate events with metadata.

We added an initial fix to have a new event LEADER_DO_CLEANUP_POST_ELECTION which gets called once after becoming leader. But, the cleanup is dependent on successful completion of LEADER_DO_ASSIGNMENT. If LEADER_DO_ASSIGNMENT fails, LEADER_DO_ASSIGNMENT is added again to the Queue. we have no means to remove LEADER_DO_CLEANUP_POST_ELECTION this from the queue, since the ordering is important. [PR685](https://github.com/linkedin/brooklin/pull/685).

To accomplish this, we will use eventMetadata in CoordinatorEvent to set cleanup to true when coordinator becomes leader and perform cleanup on successful finish of LEADER_DO_ASSIGNMENT.

We also need to fix CoordinatorEventBlockingQueue which allows duplicate entries for Events with metadata. To accomplish this, implementing equals and hashcode in CoordinatorEvent. Now, we will not allow duplicate entries for events with metadata.[PR632](https://github.com/linkedin/brooklin/pull/632). There was a discussion in PR632 to have long-term fix for this. Adding it as part of this PR as well.


datastream-server/src/test/java/com/linkedin/datastream/server/TestCoordinatorEventBlockingQueue.java
```diff
@@ -16,38 +16,20 @@
...
+    eventBlockingQueue.put(CoordinatorEvent.createLeaderDoAssignmentEvent(false));
+    eventBlockingQueue.put(CoordinatorEvent.createLeaderDoAssignmentEvent(true));
+    eventBlockingQueue.put(CoordinatorEvent.createLeaderDoAssignmentEvent(false));
+    eventBlockingQueue.put(CoordinatorEvent.createLeaderDoAssignmentEvent(true));
     eventBlockingQueue.put(CoordinatorEvent.createLeaderPartitionAssignmentEvent("test1"));
     eventBlockingQueue.put(CoordinatorEvent.createLeaderPartitionAssignmentEvent("test1"));
     eventBlockingQueue.put(CoordinatorEvent.createLeaderPartitionAssignmentEvent("test2"));
     eventBlockingQueue.put(CoordinatorEvent.HANDLE_ASSIGNMENT_CHANGE_EVENT);
     eventBlockingQueue.put(CoordinatorEvent.HANDLE_ASSIGNMENT_CHANGE_EVENT);
     eventBlockingQueue.put(CoordinatorEvent.HANDLE_ASSIGNMENT_CHANGE_EVENT);
```
ahmedahamid: Thinking out loud: should we assert on the total number of events _actually_ in the queue after these puts, or whether our takes exhaust them?

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
```diff
@@ -38,19 +38,14 @@ public CoordinatorEventBlockingQueue() {
    */
   public synchronized void put(CoordinatorEvent event) {
     LOG.info("Queuing event {} to event queue", event.getType());
-    if (!_eventMap.containsKey(event.getType())) {
+    if (!_eventSet.contains(event)) {
```
ahmedahamid: I almost reached out to you then thought this would prolly be a lot more useful to ask over here for knowledge sharing and wider visibility (think of future maintainers):
1. Why could `put()` be invoked with duplicate events?
2. Why are we rejecting duplicates?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -375,10 +375,9 @@ public void onBecomeLeader() {
     // when an instance becomes a leader, make sure we don't miss new datastreams and
     // new assignment tasks that was not finished by the previous leader
     _eventQueue.put(CoordinatorEvent.createHandleDatastreamAddOrDeleteEvent());
-    _eventQueue.put(CoordinatorEvent.createLeaderDoAssignmentEvent());
-    // This should be only called after createLeaderDoAssignmentEvent as it will verify/cleanup the orphan task nodes
-    // under connector.
-    _eventQueue.put(CoordinatorEvent.createLeaderDoCleanupPostElectionEvent());
+    // verify/cleanup the orphan task nodes under connector should be called only once after becoming leader.
```
somandal: May just be a good idea to call out why this should only be called once -> that it's an expensive operation.
Someone reading this code later may not realize why without making it explicit.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1001,6 +996,9 @@ private void handleLeaderDoAssignment() {
       instances.add(PAUSED_INSTANCE);
       _adapter.cleanupDeadInstanceAssignments(instances);
       _adapter.cleanupOldUnusedTasks(previousAssignmentByInstance, newAssignmentsByInstance);
+      if (cleanUpOrphanConnectorTasks) {
+        performCleanupTaskPostElection();
```
somandal: can we rename this function now, since there is no longer a "CLEANUP_TASK_POST_ELECTION" event?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -960,7 +955,7 @@ private void deleteTopic(Datastream datastream) {
         .collect(Collectors.toList());
   }
 
-  private void handleLeaderDoAssignment() {
+  private void handleLeaderDoAssignment(boolean cleanUpOrphanConnectorTasks) {
```
somandal: It'll be good to explain what this parameter does

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
```diff
@@ -78,7 +73,7 @@ public synchronized CoordinatorEvent take() throws InterruptedException {
     if (queuedEvent != null) {
       LOG.info("De-queuing event " + queuedEvent.getType());
       LOG.debug("Event queue size: {}", _eventQueue.size());
-      _eventMap.remove(queuedEvent.getType());
+      _eventSet.remove(queuedEvent);
```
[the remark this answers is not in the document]
vmaheshw: poll returns null if queue is empty.

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
vmaheshw: 1. Heartbeat is added periodically. If the Coordinator Event thread is stuck or slow at processing, we might add another heartbeat event.
2. Reason to reject is if the event is already present in the queue, it will do the exact operation twice (which is unnecessary) and delay the pipeline.

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEvent.java
```diff
@@ -63,9 +59,10 @@ private CoordinatorEvent(EventType eventType, Object eventMetadata) {
 
   /**
    * Returns an event that indicates a new assignment needs to be done (this is a leader-specific event).
+   * cleanUpOrphanConnectorTasks should be set to true once when the coordinator becomes leader.
```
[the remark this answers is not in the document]
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
```diff
@@ -38,19 +38,14 @@ public CoordinatorEventBlockingQueue() {
    */
   public synchronized void put(CoordinatorEvent event) {
     LOG.info("Queuing event {} to event queue", event.getType());
-    if (!_eventMap.containsKey(event.getType())) {
+    if (!_eventSet.contains(event)) {
       // only insert if there isn't an event present in the queue with the same name
```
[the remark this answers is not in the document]
vmaheshw: Done.

datastream-server/src/test/java/com/linkedin/datastream/server/TestCoordinatorEventBlockingQueue.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: Done.

PR #694, opened 2020-03-14
Move consumer.close() to a try-catch block in AbstractKafkaBasedConnectorTask

somandal:
The AbstractKafkaBasedConnectorTask calls consumer.close() in the finally block
without checking for exceptions. If consumer.close() throws an exception, it can result
in the remaining cleanup to be ignored. This can have major repercussions for scenarios
in partition-managed, where the lock is not released. So we should encapsulate the
consumer.close() in a try-catch block to prevent this.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -352,7 +352,11 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
     } finally {
       _stoppedLatch.countDown();
       if (null != _consumer) {
-        _consumer.close();
+        try {
+          _consumer.close();
+        } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -352,7 +352,11 @@ public void run() {
     } finally {
       _stoppedLatch.countDown();
       if (null != _consumer) {
-        _consumer.close();
+        try {
+          _consumer.close();
```
gauravrkulkarni: should we have retries for this?

somandal: I don’t think retry is necessary since we are anyways on the shutdown path. Let me know if you have a reason for why you think retry may be helpful

gauravrkulkarni: synced with Sonam offline, it looks like assign can be called on same set of partitions. And since the task will be shut down, we won't use the consumer for polling again even if its not closed.

PR #696, opened 2020-03-24
[WIP] Added timeouts on send and flush calls in KafkaProducerWrapper

jzakaryan:
the contribution guidelines.

### DIFF
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -12,15 +12,18 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
@@ -31,6 +34,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.RateLimiter;
 
+import com.linkedin.datastream.common.CompletableFutureUtils;
 import com.linkedin.datastream.common.DatastreamRuntimeException;
 import com.linkedin.datastream.common.DatastreamTransientException;
 import com.linkedin.datastream.common.ReflectionUtils;
@@ -59,6 +63,9 @@ class KafkaProducerWrapper<K, V> {
 
   private static final int TIME_OUT = 2000;
   private static final int MAX_SEND_ATTEMPTS = 10;
+  private static final int SEND_TIME_OUT = 5000;
+  private static final int FLUSH_TIME_OUT = 10 * 60 * 1000;
+
   private final Logger _log;
   private final long _sendFailureRetryWaitTimeMs;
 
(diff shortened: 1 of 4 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -59,6 +63,9 @@
 
   private static final int TIME_OUT = 2000;
   private static final int MAX_SEND_ATTEMPTS = 10;
+  private static final int SEND_TIME_OUT = 5000;
```
jzakaryan: We need to come up with reasonable numbers here. Maybe test a bit more to understand how the system behaves if we change these.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * An extension of {@link KafkaProducerWrapper} with bounded calls for flush and send
+ */
+class BoundedKafkaProducerWrapper<K, V> extends KafkaProducerWrapper<K, V> {
+  private static final int DEFAULT_SEND_TIME_OUT = 5000;
+  private static final int DEFAULT_FLUSH_TIME_OUT = 10 * 60 * 1000;
+
+  private static final String SEND_TIMEOUT_CONFIG_KEY = "brooklin.server.kafkaProducerWrapper.sendTimeout";
+  private static final String FLUSH_TIMEOUT_CONFIG_KEY = "brooklin.server.kafkaProducerWrapper.flushTimeout";
```
somandal, line 31: This isn't the right way to pass configs. Any configs you want to pass here need to be scoped under the transport provider configs. The KafkaProducerWrapper receives all of the transport provider configs, so you'd want to add these under that scope. We don't need to access the full property name, because as the configs pass through the layers, the relevant prefixes are removed. You can see how other configs are accessed in KafkaProducerWrapper, you'll see they don't have that whole "brooklin.server" prefix. You can access these the same way.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -142,7 +143,7 @@ private void populateDefaultProducerConfigs() {
         DEFAULT_MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION_VALUE);
   }
 
-  private Optional<Producer<K, V>> maybeGetKafkaProducer(DatastreamTask task) {
+  protected Optional<Producer<K, V>> maybeGetKafkaProducer(DatastreamTask task) {
```
somandal, line 146: I don't see you overriding this in your Bounded implementation, why make this protected?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -246,7 +253,7 @@ private synchronized void shutdownProducer() {
     }
   }
 
-  private DatastreamRuntimeException generateSendFailure(Exception exception) {
+  protected DatastreamRuntimeException generateSendFailure(Exception exception) {
```
somandal, line 256: I don't see you overriding this in your Bounded implementation, why make this protected?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -60,7 +61,8 @@
 
   private static final int TIME_OUT = 2000;
   private static final int MAX_SEND_ATTEMPTS = 10;
-  private final Logger _log;
+
+  protected final Logger _log;
```
somandal, line 65: Move this with other protected member variables

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
jzakaryan: @somandal I've discussed this with Ahmed. Will push the fixe soon

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.linkedin.datastream.common.CompletableFutureUtils;
+import com.linkedin.datastream.common.VerifiableProperties;
+
+/**
+ * An extension of {@link KafkaProducerWrapper} with bounded calls for flush and send
+ */
+class BoundedKafkaProducerWrapper<K, V> extends KafkaProducerWrapper<K, V> {
+  private static final int DEFAULT_SEND_TIME_OUT = 5000;
+  private static final int DEFAULT_FLUSH_TIME_OUT = 10 * 60 * 1000;
```
ahmedahamid, line 28: 1. Please encode time units into variable/config names, e.g. `DEFAULT_SEND_TIME_OUT_MS`
2. Use `long`s for timeout configs (e.g. see existing [timeout configs](https://github.com/linkedin/brooklin/wiki/Kafka-Connectors-Shared-Logic#configuration))
3. If you like, you can use `Duration` methods to initialize (e.g. `Duration.ofSeconds(5)`)

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+
+  private static final String SEND_TIMEOUT_CONFIG_KEY = "brooklin.server.kafkaProducerWrapper.sendTimeout";
+  private static final String FLUSH_TIMEOUT_CONFIG_KEY = "brooklin.server.kafkaProducerWrapper.flushTimeout";
+
+  private int _sendTimeout;
+  private int _flushTimeout;
```
ahmedahamid, line 34: final

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.errors.InterruptException;
+
+import com.linkedin.datastream.common.CompletableFutureUtils;
+import com.linkedin.datastream.common.VerifiableProperties;
+
+/**
+ * An extension of {@link KafkaProducerWrapper} with bounded calls for flush and send
+ */
+class BoundedKafkaProducerWrapper<K, V> extends KafkaProducerWrapper<K, V> {
```
ahmedahamid, line 26: `Bounded` is a little vague cause it's easy to confuse with buffering.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -186,56 +187,62 @@ int getTasksSize() {
...
-        throw generateSendFailure(e);
+      } catch (Exception ex) {
+        _log.error("Send failed for partition {} with an exception", producerRecord.partition(), ex);
+        throw generateSendFailure(ex);
       }
     }
   }
 
-  private synchronized void shutdownProducer() {
+  void doSend(Producer<K, V> producer, ProducerRecord<K, V> record, Callback callback) {
```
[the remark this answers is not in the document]
somandal: Sure,  makes sense. Just keep things consistent, i.e. if you need a more visible scope for multiple methods, declare them all as package private, and the rest as private. Don't mix protected and package private without having a very good reason to.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+    producer.send(record, (metadata, exception) -> {
+      if (exception == null) {
+        future.complete(metadata);
+      } else {
+        future.completeExceptionally(new KafkaClientException(metadata, exception));
+      }
+    });
+
+    return future;
+  }
```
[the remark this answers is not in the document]
jzakaryan: @ahmedahamid I just think that cancellation and timeout are semantically different. We may want to cancel the future after timeout in our case, but that's not necessarily true in general. Also, smth may be cancelled without waiting for timeout (based on user input or other external factors). Just thinking out loud. Will see whether I can get rid of the utils.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -186,56 +187,62 @@ int getTasksSize() {
...
-          cause = cause.getCause();
-        }
-        // Set a max_send_attempts for KafkaException as it may be non-recoverable
-        if (numberOfAttempt > MAX_SEND_ATTEMPTS || ((cause instanceof Error || cause instanceof RuntimeException))) {
-          _log.error("Send failed for partition {} with a non retriable exception", producerRecord.partition(), e);
-          throw generateSendFailure(e);
+      } catch (IllegalStateException ex) {
+        // The following exception should be quite rare as most exceptions will be throw async callback
+        _log.warn("Either send is called on a closed producer or broker count is less than minISR, retry in {} ms.",
+            _sendFailureRetryWaitTimeMs, ex);
```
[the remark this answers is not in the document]
jzakaryan: Good catch. This is definitely a bug I introduced. Will revert the entire exception handling piece in `send()`. Better leave it as it is.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
```diff
@@ -0,0 +1,97 @@
...
+
+  private CompletableFuture<RecordMetadata> produceMessage(Producer<K, V> producer, ProducerRecord<K, V> record) {
+    CompletableFuture<RecordMetadata> future = new CompletableFuture<>();
+
+    producer.send(record, (metadata, exception) -> {
+      if (exception == null) {
+        future.complete(metadata);
+      } else {
+        future.completeExceptionally(new KafkaClientException(metadata, exception));
```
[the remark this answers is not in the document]
jzakaryan: There were tons of checks up the callback chain that are dealing with the metadata and exception. You're probably right, I need to see if it's safe to do so and remove `KafkaClientException`. I was kind of forced to introduce it in the first place.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/BoundedKafkaProducerWrapper.java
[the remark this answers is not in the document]
ahmedahamid: > I just think that cancellation and timeout are semantically different.

I didn't mean to suggest we should propagate `CancellationException` to `callback`. I was assuming `doSend()` will still construct a `TimeoutException` in case `future` is cancelled. That does sound a bit roundabout though; it would certainly be better to do `future.completeExceptionally(new TimeoutException(...))` instead of `future.cancel()` if `future.isDone()` is false after timeout.

vmaheshw:
@jzakaryan Do we still need this PR?

PR #697, opened 2020-04-06
Improve partitionV2 logging in DatastreamTaskImpl.

vmaheshw:
Currently, partition as Array List is converted into String list in partitionV2 and partitionV2 is used for assignment. The task will 2000 partitions show up in logs as 1,2,3....,1999,2000. which takes up most of the log file and make it difficult to debug. Summarizing this to [1-2000] will save log space and  allow efficient debugging.


datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
```diff
@@ -436,8 +436,25 @@ public int hashCode() {
   @Override
   public String toString() {
     // toString() is mainly for logging purpose, feel free to modify the content/format
-    return String.format("%s(%s), partitionsV2=%s, partitions=%s, dependencies=%s", getDatastreamTaskName(), _connectorType,
-        String.join(",", _partitionsV2), LogUtils.logNumberArrayInRange(_partitions), _dependencies);
+    // If _partitions is populated, it is replicated into _partitionsV2 for assignment. However,
+    // if _partitions is not populated, we can assume that only _partitionsV2 is populated.
```
somandal, line 440: I think more context is necessary here for when _partitions vs. _partitionsV2 is used. Partition managed uses _partitionsV2 and puts in the full name of the Topic-Partition, due to which the log cannot be reduced in this way.

I'd also question whether we should avoid populating _partitionsV2 altogether for any tasks that are not Partition managed. This can be explored as part of another PR though (maybe open a ticket for this)

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
```diff
@@ -436,8 +436,25 @@ public int hashCode() {
...
+    if (_partitions.size() > 0) {
+      try {
+        List<Integer> partitionList = _partitionsV2.stream().map(Integer::parseInt).collect(Collectors.toList());
+
+        return String.format("%s(%s), partitionsV2=%s, partitions=%s, dependencies=%s", getDatastreamTaskName(),
```
somandal, line 445: The format string is used in 3 places now. Create a local variable/member constant FORMAT_STRING?

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
```diff
@@ -436,8 +436,25 @@ public int hashCode() {
...
+            _connectorType, LogUtils.logNumberArrayInRange(partitionList), LogUtils.logNumberArrayInRange(_partitions),
+            _dependencies);
+      } catch (NumberFormatException e) {
+        LOG.error(e.getMessage());
+        return String.format("%s(%s), partitionsV2=%s, partitions=%s, dependencies=%s", getDatastreamTaskName(),
```
somandal, line 450: Remove the return log from here. It's the same as the one outside the if() block. Let it drop into the return outside the if() block.

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
vmaheshw: Added comments. Also created ticket to investigate.

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
vmaheshw: Redid the logic.

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
vmaheshw: Redid the logic. Format string is in 1 place now.

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
```diff
@@ -436,8 +438,21 @@ public int hashCode() {
   @Override
   public String toString() {
     // toString() is mainly for logging purpose, feel free to modify the content/format
-    return String.format("%s(%s), partitionsV2=%s, partitions=%s, dependencies=%s", getDatastreamTaskName(), _connectorType,
-        String.join(",", _partitionsV2), LogUtils.logNumberArrayInRange(_partitions), _dependencies);
+    // When DatastreamTaskImpl is created using constructor that passes _partitionsV2, _partitions is not populated.
+    // When DatastreamTaskImpl is created using constructor that passes _partitions, _partitionsV2 is also populated.
+    // So, if _partitions is not populated, we can assume that only _partitionsV2 is populated.
+    String partitionsV2FormatLog = String.join(",", _partitionsV2);
+    if (_partitions.size() > 0) {
```
DEEPTHIKORAT, line 445: can this check be `(if (!StringUtils.isBlank(_partitionsV2))` since only v2 is being referenced within the conditional statement?

DEEPTHIKORAT: Thanks for the explanation. ParititonsV2 is topic-partitions or stringified partitions based on low-level or high-level tasks.

PR #698, opened 2020-04-07
Update ArithmeticException to IllegalArgumentException in StickyPartitionAssignmentStrategy.

vmaheshw:
2020/03/19 00:00:00.014 WARN [Coordinator] [Thread-62] [brooklin-service] [] Partition assignment failed, Exception:
java.lang.ArithmeticException: / by zero
at com.linkedin.datastream.server.assignment.StickyPartitionAssignmentStrategy.assignPartitions(StickyPartitionAssignmentStrategy.java:84) ~[datastream-server-26.0.1.jar:?]
at com.linkedin.datastream.server.Coordinator.performPartitionAssignment(Coordinator.java:1037) [datastream-server-26.0.1.jar:?]
at com.linkedin.datastream.server.Coordinator.handleEvent(Coordinator.java:717) [datastream-server-26.0.1.jar:?]
at com.linkedin.datastream.server.Coordinator$CoordinatorEventProcessor.run(Coordinator.java:1643) [datastream-server-26.0.1.jar:?]

2020/03/19 00:00:00.014 INFO [Coordinator] [Thread-62] [brooklin-service] [] END: Handle event type:LEADER_PARTITION_ASSIGNMENT

Fixing the exception by doing proper checks. Whenever the counters are zero, it is okay to retry leader assignment.


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -16,6 +16,7 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.Validate;
```
somandal, line 19: can we use lang3?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -67,6 +68,8 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
 
     LOG.info("old partition assignment info, assignment: {}", currentAssignment);
 
+    Validate.isTrue(currentAssignment.size() > 0, "Zero tasks assigned. Retry leader assignment.");
```
somandal, line 71: just curious, do we know why currentAssignment size was 0? I know saw this happen once on ZK session expiry and looked like the ZKClient was null (had the fix Justin made)

somandal: Also the log message, can we change it to: "Retry leader partition assignment"?

vmaheshw: Unfortunately, I don't have the logs to confirm why. I thought it is still a good fix irrespective of the cause.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -67,6 +68,9 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
 
     LOG.info("old partition assignment info, assignment: {}", currentAssignment);
 
+    Validate.isTrue(currentAssignment.size() > 0,
```
DEEPTHIKORAT, line 71: Isn't this going to raise IllegalArgumentException since for whatever reason we are seeing it be 0? Not sure I understand what the change set out to achieve. Throw a more appropriate exception?

vmaheshw: I feel IllegalArgumentException is appropriate for this api. This change is to not do the assignment if total assignments are zero and redo leader assignment.

somandal: Just to clarify, the behavior of this this won't change right, just the type of exception thrown? Even before this change, an exception would be thrown and the assignment retried?

vmaheshw: Yes, Just Arithmetic Exception is something we want to avoid.

PR #703, opened 2020-04-16
Fix Connector task cleanup, when leader coordinator thread gets interrupted.

vmaheshw:
We are seeing lot of orphan nodes getting created when the cluster is bounced. 
During leader assignment, dead instances are also cleaned up.

Clean up order for dead instances:

1. Remove all nodes under /cluster/instances/assignment
2. Remove node /cluster/instances
3. Remove connector node from /cluster/connector/espresso/<Connector Task>.
4. Orphan node cleanup.

If Thread gets interrupted between 2 and 3 (node is restarted), there will be orphan connector nodes, but nodes under instances/assignments will be cleaned up. Next leader will only check the dead instances and will not find anything about these connector nodes. During cleanup of orphan connector tasks (step 4), it will clean up these nodes. But, if the leader gets interrupted again before performing step4, the cleanup will not be complete. 

Bouncing the cluster with some nodes at a time, will make multiple leader change within short duration. We perform ls during OrphanConnectorCleanupTask. But, if the children size of znode becomes greater than juteMaxBuffer, zkClient will start throwing exception and will not let the orphanConnectorCleanUpTask go through.

To solve this problem:
We will merge cleanup from step 2 and step3 by pre-calculating the unusedTask Set and for every node deleted from /instance/<instanceNumber>/assignments and present in unusedTask Set, right after that we will delete the connector task node from /connector/<connectorName>/<Unused Task>. This will ensure that node present under instance and connector task will get cleanup right one after another, rather than first all nodes deleted under instance and then deleted under connector.

Sample code:
1. for each dead instance node: 
      walk its children nodes:
            delete the node 
            if node in UnusedTaskSet:
                  delete the task node under connector.
          
So, when the leader Coordinator thread gets interrupted and it is currently at step 2, all the  nodes deleted from /instance/assignments will also be cleaned up from /connector except may be the last one (which the next leader can take care of during CleanupOrphanConnectorTasks.)


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -850,13 +846,36 @@ public void setDatastreamTaskStateForKey(DatastreamTask datastreamTask, String k
...
       LOG.info("Cleaning up assignments for dead instances: " + deadInstances);
 
       for (String instance : deadInstances) {
+        // get the old assignment from ZooKeeper
+        Set<String> oldAssignments = new HashSet<>(getInstanceAssignment(instance));
+
+        oldAssignments.forEach(oldAssignment -> {
+          DatastreamTask unusedTask = unusedTasksByName.remove(oldAssignment);
+          removeTaskNodes(instance, oldAssignment);
+          // /connector/<Task> node should always be deleted after /instance/<Task>/.
```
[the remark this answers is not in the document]
vmaheshw: Can you please clarify this again?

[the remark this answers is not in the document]
somandal: updated the comment

vmaheshw:
> Let's sync up first cause I have some fundamental questions/concerns.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -841,22 +838,50 @@ public void setDatastreamTaskStateForKey(DatastreamTask datastreamTask, String k
...
+        oldAssignments.forEach(oldAssignment -> {
+          removeTaskNodes(instance, oldAssignment);
+          // /<cluster>/connectors/<connectorType>/<task> node should always be deleted after
+          // /<cluster>/instances/<task>/.
+          // If a node is present under /instance, the code assumes that the node will be present under
+          // /<cluster>/connectors/<connectorType>/ as well. So, the cleanup order is very important.
+
+          // If the thread gets interrupted after removing node under /instance, but before deleting node
+          // under /<cluster>/connectors/<connectorType>/, next leader will take care of cleaning this orphan node in
+          // cleanUpOrphanConnectorTask call.
```
ahmedahamid, line 877: I like this comment a lot but:
- It doesn't say that `removeTaskNodes()` removes `/<cluster>/instances/<task>/`
- The path isn't exactly `/<cluster>/instances/<task>/`
- It assumes a bit of background

How about we write this as:
```java
        oldAssignments.forEach(oldAssignment -> {
          // We have two znodes to clean up for every datastream task
          //  1. /<cluster>/instances/{instanceId}/assignments/<task>
          //  2. /<cluster>/connectors/<connectorType>/<task>
          // 
          // (1) should always be deleted first. If we were to delete (2) first,
          // and the code got interrupted (e.g. due to shutdown), we would run
          // the risk of having the next leader assign a task with no presence
          // in (2) to a live host, which may cause errors in other pieces of
          // code because datastream task entries in (2) are needed for host-
          // independent operations like locking tasks and check-pointing progress.
          //
          // On the other hand, when we delete (1) before (2), if the code is 
          // interrupted in between the two operations, the next leader will not
          // assign that task to any host; only (2) will become an orphan node 
          // and the next leader will take care of deleting it in cleanUpOrphanConnectorTask().
          
          // Delete (1)
          removeTaskNodes(instance, oldAssignment);
          
          DatastreamTask unusedTask = unusedTasksByName.remove(oldAssignment);
          if (unusedTask != null) {
            // Delete (2)
            deleteConnectorTask(unusedTask.getConnectorType(), unusedTask.getDatastreamTaskName());
            unusedTasks.remove(unusedTask);
          }
        });
```

datastream-server/src/test/java/com/linkedin/datastream/server/zk/TestZkAdapter.java
```diff
@@ -787,6 +788,49 @@ public void testDeleteTasksWithPrefix() {
...
+      if (i % 2 == 0) {
+        unusedTaskSet.add(dsTask);
+      }
+    }
+    updateInstanceAssignment(adapter, adapter.getInstanceName(), tasks1);
+
+    // Send some of the connector tasks as unused tasks to check there are still left over tasks under /connector,
+    // but everything cleaned up under /instance.
+    Assert.assertEquals(unusedTaskSet.size(), totalTasks / 2);
+    adapter.cleanupDeadInstanceAssignmentsAndUnusedTasks(new ArrayList<>(), unusedTaskSet);
```
ahmedahamid, line 821: `Collections.emptyList()`

datastream-server/src/test/java/com/linkedin/datastream/server/zk/TestZkAdapter.java
```diff
@@ -769,7 +770,7 @@ public void testDeleteTasksWithPrefix() {
     leftOverTasks = zkClient.getChildren(KeyBuilder.connector(testCluster, connectorType));
     Assert.assertEquals(leftOverTasks.size(), 2);
 
-    updateInstanceAssignment(adapter, adapter.getInstanceName(), new ArrayList<DatastreamTask>());
+    updateInstanceAssignment(adapter, adapter.getInstanceName(), new ArrayList<>());
```
ahmedahamid, line 773: `Collections.emptyList()`

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -841,22 +838,50 @@ public void setDatastreamTaskStateForKey(DatastreamTask datastreamTask, String k
...
    * reassigned or become safe to discard per strategy requirement.
    * Coordinator is expected to cache the "current" assignment before
    * invoking the assignment strategy and pass the saved assignment
    * to us to figure out the obsolete tasks.
    * @param instances Instances for which nodes are to be cleaned up
+   * @param unusedTasks Unused tasks that can be cleaned up once they
+   *                    are removed from dead instances.
    */
-  public void cleanupDeadInstanceAssignments(List<String> instances) {
+  public void cleanupDeadInstanceAssignmentsAndUnusedTasks(List<String> instances,
```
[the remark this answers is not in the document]
vmaheshw: Along with cleaning up dead instance data this function is also cleaning up remaining unused tasks which are not present in new assignment. so this name may be insufficient.

How about cleanUpDeadInstanceDataAndOtherUnusedTasks ?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -841,22 +838,65 @@ public void setDatastreamTaskStateForKey(DatastreamTask datastreamTask, String k
...
+      List<String> instances) {
+
+    LOG.info("cleanUpDeadInstanceData is called");
+    Set<DatastreamTask> unusedTasks = findOldUnusedTasks(previousAssignmentByInstance, newAssignmentsByInstance);
+
+    Set<String> deadInstances = previousAssignmentByInstance.keySet();
     deadInstances.removeAll(instances);
+
+    LOG.debug("unusedTasks before cleanUp: {}", unusedTasks.stream()
+        .collect(Collectors.toMap(DatastreamTask::getDatastreamTaskName, Function.identity())).keySet());
```
ahmedahamid, line 865: If you're only interested in `keySet()` then maybe we don't need a map. A list will do.
```java
    LOG.debug("unusedTasks before cleanUp: {}", unusedTasks.stream()
        .map(DatastreamTask::getDatastreamTaskName).collect(Collectors.toList()));
```
The same applies to lines 910-911.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -841,22 +838,65 @@ public void setDatastreamTaskStateForKey(DatastreamTask datastreamTask, String k
...
    * @param instances Instances for which nodes are to be cleaned up
    */
-  public void cleanupDeadInstanceAssignments(List<String> instances) {
-    List<String> deadInstances = getAllInstances();
+  public void cleanUpDeadInstanceDataAndOtherUnusedTasks(
+      Map<String, Set<DatastreamTask>> previousAssignmentByInstance,
+      Map<String, List<DatastreamTask>> newAssignmentsByInstance,
+      List<String> instances) {
+
+    LOG.info("cleanUpDeadInstanceData is called");
```
ahmedahamid, line 858: Stale function name. The same applies to line 914.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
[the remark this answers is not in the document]
ahmedahamid: Both names are equally descriptive. Thank you.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -841,22 +838,65 @@ public void setDatastreamTaskStateForKey(DatastreamTask datastreamTask, String k
...
+   * @param previousAssignmentByInstance Previous task assignment
+   * @param newAssignmentsByInstance New task assignment.
    * @param instances Instances for which nodes are to be cleaned up
    */
-  public void cleanupDeadInstanceAssignments(List<String> instances) {
-    List<String> deadInstances = getAllInstances();
+  public void cleanUpDeadInstanceDataAndOtherUnusedTasks(
+      Map<String, Set<DatastreamTask>> previousAssignmentByInstance,
+      Map<String, List<DatastreamTask>> newAssignmentsByInstance,
+      List<String> instances) {
```
ahmedahamid, line 856: Rename to `liveInstances`

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -867,29 +908,56 @@ public void cleanupDeadInstanceAssignments(List<String> instances) {
...
+        oldTasks.stream().filter(x -> !newTasks.contains(x)).collect(Collectors.toSet());
+    LOG.info("Unused tasks to be deleted: {}", unusedTasks);
+    return unusedTasks;
+  }
 
+  /**
+   * Delete the unused connector tasks from zookeeper.
+   * @param unusedTasks Unused tasks.
+   */
+  private void cleanUpOldUnusedTasksFromConnector(Collection<DatastreamTask> unusedTasks) {
```
somandal, line 945: Can you add logs for when this starts and when this completes? Want to be able to figure out how long this process takes

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -867,29 +908,56 @@ public void cleanupDeadInstanceAssignments(List<String> instances) {
         _liveTaskMap.remove(instance);
       }
     }
+    LOG.debug("unusedTasks remaining after dead instances cleanup: {}", unusedTasks.stream()
+        .map(DatastreamTask::getDatastreamTaskName)
+        .collect(Collectors.toList()));
+    // Clean the remaining unusedTasks.
```
somandal, line 914: Can you add a log before this so that we can find out how long the previous phase of cleanup took? I am concerned that we are increasing the time it will take to cleanup the dead instances, which can lead to handleDoLeaderAssignmentChange failures that can wedge the cluster into a bad state requiring a cluster restart. We should have sufficient logging to understand how long each phase takes in case we need to make a call later on about preventing such stages from getting interrupted in the first place.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
vmaheshw: There are start and end logs in the api calling this api which will take care of it. Also, the api inside this has api for each task.

somandal: Adding logs here may be useful for other callers of this function that aren't going through cleanUpDeadInstanceDataAndOtherUnusedTasks()

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
vmaheshw: We are not really increasing the dead instance cleanup time. Just doing the two cleaning process interleaved. I don't really see a value of another debug line in the middle. If we come across a condition when we need it, we can always enable debug logs.

somandal: I know we have run into issues when cleaning up dead instances gets interrupted and I would prefer having a log to know how long that takes. Cleaning up two nodes instead of one can essentially double the time this process takes since you need to make two calls to ZK to delete.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
vmaheshw: "LOG.info("Unused tasks to be deleted: {}", unusedTasks);" This is the log present in the other path as well to detect.

somandal: And is end time also logged for other code paths?

somandal: What if I want to find out how long these take to complete without it getting interrupted? I'm just trying to get data on how long these cleanup phases take on average. Today we only know when each event in the coordinator starts and ends. There is no way to track how long cleanup itself takes. Knowing this may be important to make a call on whether we should minimize interruptions during cleanup or not in the future.

vmaheshw: Added the log that will cover both the conversations.

PR #705, opened 2020-04-29
Avoid accessing consumer in send callback to avoid ConcurrentModificationException in the Kafka connector

somandal:
This change just avoids allowing access to the consumer from different threads to avoid the ConcurrentModificationException we sometimes see. It basically tracks all the TopicPartitions which have seen send failures, and attempts to rewind their offsets to the last checkpoint after flush() or before the next poll.

The flush() enabled mode can still land up committing offsets for TopicPartitions that have seen send failures in case rewind to last checkpoint fails. Addressing this will be done as part of a separate PR.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -109,6 +109,9 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected Consumer<?, ?> _consumer;
   protected final Set<TopicPartition> _consumerAssignment = new HashSet<>();
 
+  // TopicPartitions which have seen exceptions on send. Access to this map must be synchronized.
+  protected final Map<TopicPartition, Exception> _sendFailureTopicPartitionExceptionMap = new HashMap<>();
+
   // Datastream task updates that need to be processed
   protected final Set<DatastreamConstants.UpdateType> _taskUpdates = Sets.newConcurrentHashSet();
   @VisibleForTesting
@@ -229,8 +232,13 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
     for (TopicPartition topicPartition : records.partitions()) {
       for (ConsumerRecord<?, ?> record : records.records(topicPartition)) {
         try {
-          if (_autoPausedSourcePartitions.containsKey(topicPartition)) {
-            _logger.warn("Abort sending as {} is auto-paused, rewind offset", topicPartition);
+          boolean skipRecord;
+          synchronized (_sendFailureTopicPartitionExceptionMap) {
+            skipRecord = _autoPausedSourcePartitions.containsKey(topicPartition)
+                || _sendFailureTopicPartitionExceptionMap.containsKey(topicPartition);
+          }
+          if (skipRecord) {
+            _logger.warn("Abort sending as {} is auto-paused or saw a send failure, rewind offset", topicPartition);
             seekToLastCheckpoint(Collections.singleton(topicPartition));
             break;
           } else {
@@ -257,15 +265,23 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
       seekToLastCheckpoint(Collections.singleton(srcTopicPartition));
       partitionRewound = true;
     } catch (Exception e) {
-      _logger.error("Partition rewind failed due to ", e);
+      _logger.error(String.format("Partition rewind for %s failed due to ", srcTopicPartition), e);
     }
     if (_pausePartitionOnError && (!partitionRewound || !containsTransientException(ex))) {
-      // if doesn't contain DatastreamTransientException and it's configured to pause partition on error conditions,
-      // add to auto-paused set
+      // If partition rewind failed or the exception is not of type DatastreamTransientException and it is configured to
+      // pause partition on error conditions, add it to the auto-paused set
       _logger.warn("Adding source topic partition {} to auto-pause set", srcTopicPartition);
       _autoPausedSourcePartitions.put(srcTopicPartition,
           PausedSourcePartitionMetadata.sendError(start, _pauseErrorPartitionDuration, ex));
       _taskUpdates.add(DatastreamConstants.UpdateType.PAUSE_RESUME_PARTITIONS);
+    }
+  }
+
+  protected void rewindAndPausePartitionsOnSendException() {
+    // For all topic partitions which have seen send exceptions, attempt to rewind them to the last checkpoint
+    synchronized (_sendFailureTopicPartitionExceptionMap) {
+      _sendFailureTopicPartitionExceptionMap.forEach(this::rewindAndPausePartitionOnException);
+      _sendFailureTopicPartitionExceptionMap.clear();
     }
   }
 
@@ -283,10 +299,12 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
       TopicPartition srcTopicPartition, int numBytes, SendCallback sendCallback) {
     _producer.send(datastreamProducerRecord, ((metadata, exception) -> {
       if (exception != null) {
-        String msg = String.format("Detect exception being thrown from callback for src partition: %s while "
-            + "sending, metadata: %s , exception: ", srcTopicPartition, metadata);
+        String msg = String.format("Detected exception being thrown from send callback for source topic-partition: %s "
+            + "with metadata: %s, exception: ", srcTopicPartition, metadata);
         _logger.warn(msg, exception);
-        rewindAndPausePartitionOnException(srcTopicPartition, exception);
+        synchronized (_sendFailureTopicPartitionExceptionMap) {
+          _sendFailureTopicPartitionExceptionMap.put(srcTopicPartition, exception);
+        }
       } else {
         _consumerMetrics.updateBytesProcessedRate(numBytes);
       }
@@ -522,6 +540,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
     if (force || timeSinceLastCommit > _offsetCommitInterval) {
       _logger.info("Trying to flush the producer and commit offsets.");
       _producer.flush();
+      rewindAndPausePartitionsOnSendException();
       try {
         commitWithRetries(consumer, Optional.empty());
         _lastCommittedTime = System.currentTimeMillis();
@@ -676,6 +695,9 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
    * new update type when there is any update to datastream task (in method checkForUpdateTask())
    */
   protected void preConsumerPollHook() {
+    // check if any send failures were seen on the last poll and rewind them before the next poll
+    rewindAndPausePartitionsOnSendException();
+
     // check if any auto-paused partitions need to be resumed
     checkForPartitionsToAutoResume();
 
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -110,6 +110,8 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected final Set<TopicPartition> _consumerAssignment = new HashSet<>();
 
   // TopicPartitions which have seen exceptions on send. Access to this map must be synchronized.
+  // A ConcurrentHashMap is not used here due to the need for having more than one operation performed together as an
+  // atomic block
   private final Map<TopicPartition, Exception> _sendFailureTopicPartitionExceptionMap = new HashMap<>();
 
   // Datastream task updates that need to be processed
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -110,7 +110,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected final Set<TopicPartition> _consumerAssignment = new HashSet<>();
 
   // TopicPartitions which have seen exceptions on send. Access to this map must be synchronized.
-  protected final Map<TopicPartition, Exception> _sendFailureTopicPartitionExceptionMap = new HashMap<>();
+  private final Map<TopicPartition, Exception> _sendFailureTopicPartitionExceptionMap = new HashMap<>();
 
   // Datastream task updates that need to be processed
   protected final Set<DatastreamConstants.UpdateType> _taskUpdates = Sets.newConcurrentHashSet();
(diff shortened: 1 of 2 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -229,8 +232,13 @@ protected void translateAndSendBatch(ConsumerRecords<?, ?> records, Instant read
...
         try {
-          if (_autoPausedSourcePartitions.containsKey(topicPartition)) {
-            _logger.warn("Abort sending as {} is auto-paused, rewind offset", topicPartition);
+          boolean skipRecord;
+          synchronized (_sendFailureTopicPartitionExceptionMap) {
+            skipRecord = _autoPausedSourcePartitions.containsKey(topicPartition)
+                || _sendFailureTopicPartitionExceptionMap.containsKey(topicPartition);
+          }
+          if (skipRecord) {
+            _logger.warn("Abort sending as {} is auto-paused or saw a send failure, rewind offset", topicPartition);
```
vmaheshw, line 241: Do you see a value in clarifying in the warning whether skipping the record because of autoPausedSourcePartition or send failure ?

somandal: Sure, I can make that distinction

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -229,8 +232,14 @@ protected void translateAndSendBatch(ConsumerRecords<?, ?> records, Instant read
...
-          if (_autoPausedSourcePartitions.containsKey(topicPartition)) {
-            _logger.warn("Abort sending as {} is auto-paused, rewind offset", topicPartition);
+          boolean skipRecord;
+          synchronized (_sendFailureTopicPartitionExceptionMap) {
+            skipRecord = _autoPausedSourcePartitions.containsKey(topicPartition)
+                || _sendFailureTopicPartitionExceptionMap.containsKey(topicPartition);
+          }
+          if (skipRecord) {
+            _logger.warn("Abort sending as {} {}, rewind offset", topicPartition,
+                _autoPausedSourcePartitions.containsKey(topicPartition) ? "is auto-paused" : "saw a send failure");
```
vmaheshw, line 242: We should set both the values in sychronized above and use them to display.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -264,9 +264,12 @@ protected void sendDatastreamProducerRecord(DatastreamProducerRecord datastreamP
           ((metadata, exception) -> {
             if (exception != null) {
               _logger.warn(
-                  String.format("Detected exception being throw from callback for src partition: %s while sending producer "
-                      + "record: %s, exception: ", srcTopicPartition, datastreamProducerRecord), exception);
-              rewindAndPausePartitionOnException(srcTopicPartition, exception);
+                  String.format("Detected exception being throw from flushless send callback for source "
+                      + "topic-partition: %s with metadata: %s, exception: ", srcTopicPartition, metadata),
+                  exception);
```
vmaheshw, line 269: I think you should move this in a function in AbstractKafkaBasedConnectorTask and use that function and avoid using the map outside the class. You should also set the visibilty to private in that case.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -283,10 +300,12 @@ protected void sendDatastreamProducerRecord(DatastreamProducerRecord datastreamP
...
     _producer.send(datastreamProducerRecord, ((metadata, exception) -> {
       if (exception != null) {
-        String msg = String.format("Detect exception being thrown from callback for src partition: %s while "
-            + "sending, metadata: %s , exception: ", srcTopicPartition, metadata);
+        String msg = String.format("Detected exception being thrown from send callback for source topic-partition: %s "
+            + "with metadata: %s, exception: ", srcTopicPartition, metadata);
         _logger.warn(msg, exception);
-        rewindAndPausePartitionOnException(srcTopicPartition, exception);
+        synchronized (_sendFailureTopicPartitionExceptionMap) {
+          _sendFailureTopicPartitionExceptionMap.put(srcTopicPartition, exception);
```
vmaheshw, line 307: You can extract this out into another function and reuse between both connectorTasks

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -257,18 +266,26 @@ protected void rewindAndPausePartitionOnException(TopicPartition srcTopicPartiti
...
       _autoPausedSourcePartitions.put(srcTopicPartition,
           PausedSourcePartitionMetadata.sendError(start, _pauseErrorPartitionDuration, ex));
       _taskUpdates.add(DatastreamConstants.UpdateType.PAUSE_RESUME_PARTITIONS);
     }
   }
 
+  protected void rewindAndPausePartitionsOnSendException() {
+    // For all topic partitions which have seen send exceptions, attempt to rewind them to the last checkpoint
+    synchronized (_sendFailureTopicPartitionExceptionMap) {
+      _sendFailureTopicPartitionExceptionMap.forEach(this::rewindAndPausePartitionOnException);
```
vmaheshw, line 284: You should check if the topic-partition is already in the pause list and skip the rewind process.  Reason: suppose for a topic-partition, 10 records are polled and got 1 callback with error before next poll, this topic partition will be rewind and paused and poll will continue. Suppose the other 9 callbacks are coming with delay and come before next 2-3 poll iterations, we will do this rewind multiple-times, even though the partition is already paused and rewind. We should try to skip the same no-op  operation multiple times.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: Not sure I understood, which values?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: Yeah so I did think that, but decided against it. When we add a TopicPartition to the auto-pause list on send failure, we can do it for two reasons: a) the send exception is not transient, b) the rewind failed. For (b) I think it is safer to potentially attempt the rewind on the next send failure if possible, in case the first rewind failed for some transient reason. When we resume TopicPartitions from the auto-pause list, we don't attempt to rewind them then.

The maximum number of rewinds we'll do for a TopicPartition = number of polls between flushes, since on each poll/flush we'll rewind the TopicPartition only once in spite of receiving multiple send callbacks.

We can potentially differentiate between rewind failure vs. non-transient exception and use that to decide whether to rewind in the future. This may need more thought though. The current code is no worse than the existing code in terms of doing no-op rewinds. What do you think? Okay to leave this as is for now?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
somandal: Do you want me to move the comment or the synchronized block or both?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
vmaheshw: Do you want to mark it as TODO then with the explanation?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
vmaheshw: I initially meant just the synchronized block. Whether to move the warning log as well is your discretion.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: Sure, will do, and I'll open a ticket. Will not leave this change hanging forever.

vmaheshw: Can you add TODO and the explanation so that the thought is not lost in the process.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
vmaheshw: I meant: 
```synchronized (_sendFailureTopicPartitionExceptionMap) {
            partitionPaused = _autoPausedSourcePartitions.containsKey(topicPartition)
            sendFailure =   _sendFailureTopicPartitionExceptionMap.containsKey(topicPartition);
          }
          if (partitionPaused || sendFailure) {
            _logger.warn("Abort sending for {} paused: {} sendfailure: {}, rewind offset", topicPartition,
                partitionPaused, sendFailure");```

Comment (somandal) [reply datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java:242]: got it, thanks for clarifying!

Comment (somandal) [reply datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java:242]: done

File: datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java [hunk 6]
```diff
@@ -257,18 +265,26 @@ protected void rewindAndPausePartitionOnException(TopicPartition srcTopicPartiti
       seekToLastCheckpoint(Collections.singleton(srcTopicPartition));
       partitionRewound = true;
     } catch (Exception e) {
-      _logger.error("Partition rewind failed due to ", e);
+      _logger.error(String.format("Partition rewind for %s failed due to ", srcTopicPartition), e);
     }
     if (_pausePartitionOnError && (!partitionRewound || !containsTransientException(ex))) {
-      // if doesn't contain DatastreamTransientException and it's configured to pause partition on error conditions,
-      // add to auto-paused set
+      // If partition rewind failed or the exception is not of type DatastreamTransientException and it is configured to
```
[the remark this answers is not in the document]
somandal: done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -522,6 +540,7 @@ protected void maybeCommitOffsetsInternal(Consumer<?, ?> consumer, boolean force
     if (force || timeSinceLastCommit > _offsetCommitInterval) {
       _logger.info("Trying to flush the producer and commit offsets.");
       _producer.flush();
+      rewindAndPausePartitionsOnSendException();
```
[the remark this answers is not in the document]
somandal: done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
somandal: done, didn't add the comment there.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -109,6 +109,9 @@
   protected Consumer<?, ?> _consumer;
   protected final Set<TopicPartition> _consumerAssignment = new HashSet<>();
 
+  // TopicPartitions which have seen exceptions on send. Access to this map must be synchronized.
+  private final Map<TopicPartition, Exception> _sendFailureTopicPartitionExceptionMap = new HashMap<>();
```
DEEPTHIKORAT, line 113: You might have mentioned this to me before, but isnt a ConcurrentHashMap useful here?

somandal: So there is this one place where I iterate over all the elements and then clear the map. I wanted that block to be synchronized as a whole, which is why I decided to use HashMap instead. That block can be contended with the callbacks on code paths were we skip flush() between calls to poll(). Due to this my worry is that in the middle of iterating, if a new entry gets added due to a send callback completion, what if i miss processing it? Since I clear the map after it, we might miss rewinding some topic partitions.

Documentation for Concurrent collections seem to indicate that there is weak consistency while iterating (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#Weakly):

`Most concurrent Collection implementations (including most Queues) also differ from the usual java.util conventions in that their Iterators and Spliterators provide weakly consistent rather than fast-fail traversal:

they may proceed concurrently with other operations
they will never throw ConcurrentModificationException
they are guaranteed to traverse elements as they existed upon construction exactly once, and may (but are not guaranteed to) reflect any modifications subsequent to construction.`

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -109,6 +109,11 @@
   protected Consumer<?, ?> _consumer;
   protected final Set<TopicPartition> _consumerAssignment = new HashSet<>();
 
+  // TopicPartitions which have seen exceptions on send. Access to this map must be synchronized.
+  // A ConcurrentHashMap is not used here due to the need for having more than one operation performed together as an
```
vmaheshw, line 113: Do we need this extra explanation? It might become a norm every time to put this in comment. I understand this will make the understanding easier, but do we then want to put every time going forward. It might become cumbersome.

somandal: Deepthi pointed out that in the future someone may accidentally change this to a concurrent hashmap, and having a comment explaining why may help prevent that. I think it’s okay to explain when important so that people aren’t left wondering why one data structure is chosen over another. What do you think?

vmaheshw: Sure. sounds good.

PR #706, opened 2020-04-29
Reuse task from dead instances instead of creating new task in Broadcast strategy.

vmaheshw:
The current code creates a new task instance if the task for the datastream group is not already existing in the previous assignment for the same instance. This results in creation of new tasks whenever the cluster is bounced, resulting in extra zk operation.

With this change, we will reuse the task from other dead instances.

The following equation is not sufficient:
`Total Task instance that can be reused = sum of all tasks in dead instances.` 

If the leader was doing an assignment and reused the task from dead instance and updated the new assignment but got interrupted before deleted the dead instances, the next leader will try to reuse the tasks from the dead instance which is already assigned to another live instance by previous leader. This will result in duplicate error during update Assignment.

To avoid this:
`Total Task instance that can be reused  = sum of all tasks in dead instances - tasks in live instances.`

This will ensure that a task is not reused.

Note: This does not mean that during rebouncing the cluster, there will be no new task created. Because if the cluster has N nodes, the next leader may only see M live instances where M<N and it may clean-up the dead instances before the rest of the nodes come back online. But, this is still an improvement from no task re-use.


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -59,17 +60,27 @@ public BroadcastStrategy(Optional<Integer> maxTasks) {
...
     Map<String, Set<DatastreamTask>> newAssignment = new HashMap<>();
 
+    Set<DatastreamTask> tasksAvailableToReuse = new HashSet<>();
     // Make a copy of the current assignment, since the strategy modifies it during calculation
     Map<String, Set<DatastreamTask>> currentAssignmentCopy = new HashMap<>(currentAssignment.size());
-    currentAssignment.forEach((k, v) -> currentAssignmentCopy.put(k, new HashSet<>(v)));
+    currentAssignment.forEach((k, v) -> {
+      currentAssignmentCopy.put(k, new HashSet<>(v));
+      // building the full list and then removing the reused tasks rather than building the list from dead tasks
+      // to avoid assignment of same task from dead instance to two live instances in case of interrupt exception.
```
somandal, line 69: "removing the reused tasks"? Did you mean removing the tasks assigned to current live instances? 

> to avoid assignment of same task from dead instance to two live instances in case of interrupt exception.

So that part of the comment is a bit confusing here. What got interrupted? This will only make sense to people who already know the details about the problem you are solving. Instead can you modify this to talk about interrupted cleanup of ZK nodes due to shutdown during leader assignment?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -93,6 +107,13 @@ public BroadcastStrategy(Optional<Integer> maxTasks) {
     return newAssignment;
   }
 
+  private DatastreamTask getOrCreateDatastreamTask(List<DatastreamTask> reuseTasksPerDg, DatastreamGroup dg) {
+    if (reuseTasksPerDg.size() > 0) {
+      return reuseTasksPerDg.remove(0);
+    }
+    return new DatastreamTaskImpl(dg.getDatastreams());
```
somandal, line 114: You can modify this to be:

`return ((reuseTasksPerDg.size() > 0) ? reuseTasksPerDg.remove(0) : new DatastreamTaskImpl(dg.getDatastreams()));`

You can also inline this where it is called if you'd like.

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
```diff
@@ -279,4 +280,21 @@ public void testRebalanceTasksWhenNewInstanceIsAdded() {
...
+    BroadcastStrategy strategy = new BroadcastStrategy(Optional.empty());
+    Map<String, Set<DatastreamTask>> assignment =
+        strategy.assign(datastreams, Arrays.asList(instances), new HashMap<>());
+    instances = new String[]{"instance1", "instance2", "instance4"};
+
+    Map<String, Set<DatastreamTask>> newAssignment = strategy.assign(datastreams, Arrays.asList(instances), assignment);
+    Set<DatastreamTask> oldAssignmentTasks = assignment.values().stream().flatMap(Set::stream).collect(Collectors.toSet());
+    Set<DatastreamTask> newAssignmentTasks = newAssignment.values().stream().flatMap(Set::stream).collect(Collectors.toSet());
+    LOG.info("{}", oldAssignmentTasks);
+    LOG.info("{}", newAssignmentTasks);
```
somandal, line 297: Do you need these logs in the test?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
ahmedahamid: Strongly agree.

There's a whole story behind reusing tasks assigned to dead instances. I'd suggest expanding the comment even further to explain:
- why we reuse tasks assigned to dead instances to begin with
- caveats we consider during this reuse, like the possibility of having the same task assigned to more than one instance (and why that might happen to begin with)

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -59,17 +60,27 @@ public BroadcastStrategy(Optional<Integer> maxTasks) {
...
+    currentAssignment.forEach((k, v) -> {
+      currentAssignmentCopy.put(k, new HashSet<>(v));
+      // building the full list and then removing the reused tasks rather than building the list from dead tasks
+      // to avoid assignment of same task from dead instance to two live instances in case of interrupt exception.
+      tasksAvailableToReuse.addAll(v);
+    });
 
     for (String instance : instances) {
       newAssignment.put(instance, new HashSet<>());
       currentAssignmentCopy.putIfAbsent(instance, new HashSet<>());
```
ahmedahamid, line 75: Not your fault: this isn't very efficient; `new HashSet` isn't needed if `instance` is not absent. This could be improved to:
```java
 currentAssignmentCopy.computeIfAbsent(instance, i -> new HashSet<>());
```
If we do that, we could also improve line 76 by avoiding the extra `get()`:
```java
Set<DatastreamTask> instanceTasks = currentAssignmentCopy.computeIfAbsent(instance, i -> new HashSet<>());
tasksAvailableToReuse.removeAll(instanceTasks);
```

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -59,17 +60,27 @@ public BroadcastStrategy(Optional<Integer> maxTasks) {
...
+      tasksAvailableToReuse.removeAll(currentAssignmentCopy.get(instance));
     }
 
     int instancePos = 0;
     for (DatastreamGroup dg : datastreams) {
+      List<DatastreamTask> reuseTasksPerDg = tasksAvailableToReuse.stream().filter(x ->
+          x.getTaskPrefix().equals(dg.getTaskPrefix())).collect(Collectors.toList());
```
ahmedahamid, line 82: Would it be better if we built a map of `taskPrefix` -> `List<DatastreamTask>` out of `tasksAvailableToReuse` before the loop on 80 instead of doing this? 

I think that'll be simpler, let alone it'll spare us having to filter `tasksAvailableToReuse` over and over for every `dg`.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -78,7 +89,10 @@ public BroadcastStrategy(Optional<Integer> maxTasks) {
             .stream()
             .filter(x -> x.getTaskPrefix().equals(dg.getTaskPrefix()))
             .findFirst()
-            .orElse(new DatastreamTaskImpl(dg.getDatastreams()));
+            .orElse(null);
+        if (foundDatastreamTask == null) {
+          foundDatastreamTask = getOrCreateDatastreamTask(reuseTasksPerDg, dg);
+        }
```
ahmedahamid, line 95: You can make this part of the previous expression by replacing `orElse(null)` with:
```java
.orElseGet(() -> getOrCreateDatastreamTask(reuseTasksPerDg, dg));
```

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -93,6 +107,13 @@ public BroadcastStrategy(Optional<Integer> maxTasks) {
     return newAssignment;
   }
 
+  private DatastreamTask getOrCreateDatastreamTask(List<DatastreamTask> reuseTasksPerDg, DatastreamGroup dg) {
+    if (reuseTasksPerDg.size() > 0) {
+      return reuseTasksPerDg.remove(0);
```
ahmedahamid, line 112: Removing the last one is cheaper.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -93,6 +107,13 @@ public BroadcastStrategy(Optional<Integer> maxTasks) {
     return newAssignment;
   }
 
+  private DatastreamTask getOrCreateDatastreamTask(List<DatastreamTask> reuseTasksPerDg, DatastreamGroup dg) {
+    if (reuseTasksPerDg.size() > 0) {
```
ahmedahamid, line 111: Can be more succinctly expressed as `if (!reuseTasksPerDg.isEmpty())`

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
```diff
@@ -279,4 +280,21 @@ public void testRebalanceTasksWhenNewInstanceIsAdded() {
...
+    String[] instances = new String[]{"instance1", "instance2", "instance3"};
+    List<DatastreamGroup> datastreams = generateDatastreams("ds", 5);
+    BroadcastStrategy strategy = new BroadcastStrategy(Optional.empty());
+    Map<String, Set<DatastreamTask>> assignment =
+        strategy.assign(datastreams, Arrays.asList(instances), new HashMap<>());
+    instances = new String[]{"instance1", "instance2", "instance4"};
+
+    Map<String, Set<DatastreamTask>> newAssignment = strategy.assign(datastreams, Arrays.asList(instances), assignment);
+    Set<DatastreamTask> oldAssignmentTasks = assignment.values().stream().flatMap(Set::stream).collect(Collectors.toSet());
+    Set<DatastreamTask> newAssignmentTasks = newAssignment.values().stream().flatMap(Set::stream).collect(Collectors.toSet());
```
ahmedahamid, line 295: nit: Names are slightly inconsistent; we get `oldAssignmentTasks` from `assignment` but `newAssignmentTasks` from `newAssignment`.

How about `firstAssignment` -> `firstAssignmentTasks` and `secondAssignment` -> `secondAssignmentTasks`?

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
```diff
@@ -279,4 +280,21 @@ public void testRebalanceTasksWhenNewInstanceIsAdded() {
 
     Assert.assertEquals(newAssignment.get(instance4).size(), expectedNumTasksPerInstance);
   }
+
+  @Test
+  public void testDontCreateNewTasksWhenInstanceChanged() {
```
ahmedahamid, line 285: Rename to something along the lines of: `testReuseDeadInstanceTasks`?

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
ahmedahamid: +1

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
```diff
@@ -279,4 +280,21 @@ public void testRebalanceTasksWhenNewInstanceIsAdded() {
...
+    Map<String, Set<DatastreamTask>> assignment =
+        strategy.assign(datastreams, Arrays.asList(instances), new HashMap<>());
+    instances = new String[]{"instance1", "instance2", "instance4"};
+
+    Map<String, Set<DatastreamTask>> newAssignment = strategy.assign(datastreams, Arrays.asList(instances), assignment);
+    Set<DatastreamTask> oldAssignmentTasks = assignment.values().stream().flatMap(Set::stream).collect(Collectors.toSet());
+    Set<DatastreamTask> newAssignmentTasks = newAssignment.values().stream().flatMap(Set::stream).collect(Collectors.toSet());
+    LOG.info("{}", oldAssignmentTasks);
+    LOG.info("{}", newAssignmentTasks);
+    Assert.assertTrue(oldAssignmentTasks.containsAll(newAssignmentTasks));
```
ahmedahamid, line 298: 1. Should we make this assert stricter: assert they're actually the exact same set of tasks?

```java
Assert.assertEquals(oldAssignmentTasks, newAssignmentTasks);
```

2. Should we also assert that no task was assigned to more than one instance?

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
vmaheshw: I kept the names inline with rest of the tests in the class for uniformity. Please let me know if you feel otherwise.

ahmedahamid: Understood. No problem. Feel free to keep them as is.

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
vmaheshw: kept the test name same as StickyMulticastStrategy.

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
vmaheshw: Done.

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestBroadcastStrategy.java
ahmedahamid: No problem

PR #714, opened 2020-05-22
Fix producer.close() deadlock by moving actual producer.close() call to separate thread in KafkaProducerWrapper

somandal:
We have a deadlock scenario which looks like:

- The instance is shutdown, and all the KafkaProducerWrappers have their close() API called. This in turn results in calling the KafkaProducer's close() call. The close() API is synchronized
- We try to close the KafkaProducer (shutdownProducer()) on any send failures, and this callback is called from the KafkaProducer's Sender (I/O) thread.
- The KafkaProducerWrapper's shutdownProducer() is synchronized
- KafkaProducer's close() waits for a  Timeout duration for the close() to complete, otherwise it proceeds to force close. The force close path tries to do an `ioThread.join()` without a timeout.
- Due to the SendCallback firing with an Exception, and since SendCallbacks are processed as part of the ioThread, the callback is  stuck waiting for the lock on the shutdownProducer() method.
- The `ioThread.join()` is unable to complete because the ioThread is deadlocked waiting for the synchronized lock on shutdownProducer(), but the lock is held by the close() call.

This PR addresses the above by spawning the "producer.close()" part in a separate thread. It also removes the synchronized methods in favor of a synchronized lock object to avoid dealing with issues where the callers of the KafkaProducerWrapper APIs themselves try to lock the object. Synchronization with other methods may still be a problem, but some fixes have been added to force kill stuck tasks, etc which should reduce such deadlock scenarios.

### DIFF
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -19,6 +19,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerConfig;
@@ -104,7 +105,7 @@ class KafkaProducerWrapper<K, V> {
   private final DynamicMetricsManager _dynamicMetricsManager;
   private final String _metricsNamesPrefix;
 
-  // A lock used to synchronized access to operations performed on the _kafkaProducer object
+  // A lock used to synchronize access to operations performed on the _kafkaProducer object
   private final Object _producerLock = new Object();
 
   // An executor to spawn threads to close the producer.
@@ -129,7 +130,7 @@ class KafkaProducerWrapper<K, V> {
     _dynamicMetricsManager.registerGauge(_metricsNamesPrefix, AGGREGATE, PRODUCER_COUNT, PRODUCER_GAUGE);
 
     _clientId = transportProviderProperties.getProperty(ProducerConfig.CLIENT_ID_CONFIG);
-    if (_clientId == null || _clientId.isEmpty()) {
+    if (StringUtils.isEmpty(_clientId)) {
       _log.warn("Client ID is either null or empty");
     }
 
@@ -276,10 +277,10 @@ class KafkaProducerWrapper<K, V> {
   private DatastreamRuntimeException generateSendFailure(Exception exception) {
     _dynamicMetricsManager.createOrUpdateMeter(_metricsNamesPrefix, AGGREGATE, PRODUCER_ERROR, 1);
     if (exception instanceof IllegalStateException) {
-      _log.warn("send failed transiently with exception: ", exception);
+      _log.warn("Send failed transiently with exception: ", exception);
       return new DatastreamTransientException(exception);
     } else {
-      _log.warn("send failed, restart producer, exception: ", exception);
+      _log.warn("Send failed with a non-transient exception. Shutting down producer, exception: ", exception);
       shutdownProducer();
       return new DatastreamRuntimeException(exception);
     }
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -108,7 +108,7 @@ class KafkaProducerWrapper<K, V> {
   private final Object _producerLock = new Object();
 
   // An executor to spawn threads to close the producer.
-  private final ExecutorService _producerCloseExecutorService = Executors.newFixedThreadPool(1,
+  private final ExecutorService _producerCloseExecutorService = Executors.newSingleThreadExecutor(
       new ThreadFactoryBuilder().setNameFormat("KafkaProducerWrapperClose-%d").build());
 
   KafkaProducerWrapper(String logSuffix, Properties props) {
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -13,8 +13,11 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 
 import org.apache.kafka.clients.producer.Callback;
@@ -31,10 +34,12 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.RateLimiter;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import com.linkedin.datastream.common.DatastreamRuntimeException;
 import com.linkedin.datastream.common.DatastreamTransientException;
 import com.linkedin.datastream.common.ReflectionUtils;
+import com.linkedin.datastream.common.ThreadUtils;
 import com.linkedin.datastream.common.VerifiableProperties;
 import com.linkedin.datastream.kafka.factory.KafkaProducerFactory;
 import com.linkedin.datastream.kafka.factory.SimpleKafkaProducerFactory;
@@ -60,6 +65,8 @@ class KafkaProducerWrapper<K, V> {
 
   private static final int TIME_OUT = 2000;
   private static final int MAX_SEND_ATTEMPTS = 10;
+  private static final Duration PRODUCER_CLOSE_EXECUTOR_SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
+
   private final Logger _log;
   private final long _sendFailureRetryWaitTimeMs;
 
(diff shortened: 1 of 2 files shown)


datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaProducerWrapper.java
```diff
@@ -141,8 +153,17 @@ void verifyFlush(int numExpected) {
...
-    void verifyClose(int numExpected) {
+    void verifyClose(int numExpected) throws NoSuchMethodException {
+      // Producer close is invoked in a separate thread. Must wait for the thread to get scheduled and call close
+      Method method = Producer.class.getMethod("close", long.class, TimeUnit.class);
+      PollUtils.poll(() -> {
+        Collection<Invocation> invocations = mockingDetails(_mockProducer).getInvocations();
+        long count = invocations.stream().filter(invocation -> invocation.getMethod().equals(method)).count();
+        return count == numExpected;
+      }, 1000, 10000);
       verify(_mockProducer, times(numExpected)).close(anyLong(), any(TimeUnit.class));
```
ahmedahamid, line 164: Please, feel free to ignore this comment.

I generally dislike polling in tests for several reasons. I'll just mention the most relevant ones here:
1. It's a source of flakiness (I know you used a timeout of 10s for that reason)
2. When we do `verifyClose(0)`, `count == numExpected` will return `true` immediately, a racey behavior that does not affect this test because it has a subsequent assertion that verifies `shutdownProducer()` itself hasn't been called.
    ```java
    Assert.assertEquals(_numShutdownProducerCalls, numExpected)
    ```
    but this begs the question: do we have to assert on `_mockProducer.produce()` in the first place?

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaProducerWrapper.java
```diff
@@ -141,8 +153,17 @@ void verifyFlush(int numExpected) {
...
+      Assert.assertEquals(_numShutdownProducerCalls, numExpected);
+      _numShutdownProducerCalls = 0;
```
ahmedahamid, line 166: Unlike the other two `verify*()` methods, this one resets its call count (`_numShutdownProducerCalls`). This is a little confusing because the call counts passed to the other methods in `testFlushInterrupt()` are all cumulative.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -98,6 +105,13 @@
   private final DynamicMetricsManager _dynamicMetricsManager;
   private final String _metricsNamesPrefix;
 
+  // A lock used to synchronized access to operations performed on the _kafkaProducer object
+  private final ReentrantLock _producerLock = new ReentrantLock();
```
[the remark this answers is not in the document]
somandal: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -98,6 +105,13 @@
...
+
+  // An executor to spawn threads to close the producer.
+  private final ExecutorService _producerCloseExecutorService = Executors.newFixedThreadPool(10,
```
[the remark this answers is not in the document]
somandal: changed to 1

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -163,22 +177,27 @@ int getTasksSize() {
   }
 
   /**
-   * Must be synchronized to avoid creating duplicate producers when multiple concurrent
+   * Must be protected by a lock to avoid creating duplicate producers when multiple concurrent
    * sends are in-flight and _kafkaProducer has been set to null as a result of previous
    * producer exception.
    */
```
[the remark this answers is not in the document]
somandal: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -221,61 +240,87 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
+    }
+
+    // This may be called from the send callback. The callbacks are called from the sender thread, and must complete
+    // quickly to avoid delaying/blocking the sender thread. Thus schedule the actual producer.close() on a separate
+    // thread
     if (producer != null) {
-      producer.close(TIME_OUT, TimeUnit.MILLISECONDS);
-      NUM_PRODUCERS.decrementAndGet();
+      _producerCloseExecutorService.submit(() -> {
+        _log.debug("KafkaProducerWrapper: Closing the Kafka Producer");
```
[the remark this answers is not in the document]
somandal: done

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaProducerWrapper.java
somandal: As discussed offline, it does make sense to leave this as is since we want to test both that shutdownProducer is called and that mockProducer's close() is called.

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaProducerWrapper.java
somandal: As discussed offline, leaving this as is since mockProducers can be created and removed during the test

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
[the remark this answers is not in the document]
ahmedahamid: `newSingleThreadExecutor`?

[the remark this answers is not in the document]
somandal: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -221,47 +234,63 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
+      _producerCloseExecutorService.submit(() -> {
+        _log.info("KafkaProducerWrapper: Closing the Kafka Producer");
+        producer.close(TIME_OUT, TimeUnit.MILLISECONDS);
+        NUM_PRODUCERS.decrementAndGet();
+        _log.info("KafkaProducerWrapper: Kafka Producer is closed");
+      });
     }
   }
 
   private DatastreamRuntimeException generateSendFailure(Exception exception) {
```
DEEPTHIKORAT, line 276: nit: not sure why this method called `generateSendFailure`. Should this be renamed to `handleSendFailure`?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -98,6 +104,13 @@
   private final DynamicMetricsManager _dynamicMetricsManager;
   private final String _metricsNamesPrefix;
 
+  // A lock used to synchronized access to operations performed on the _kafkaProducer object
```
DEEPTHIKORAT, line 107: nit: A lock used to "synchronize" access to

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -98,6 +104,13 @@
...
+  private final Object _producerLock = new Object();
+
+  // An executor to spawn threads to close the producer.
+  private final ExecutorService _producerCloseExecutorService = Executors.newSingleThreadExecutor(
+      new ThreadFactoryBuilder().setNameFormat("KafkaProducerWrapperClose-%d").build());
```
DEEPTHIKORAT, line 112: I am guessing we can now use setName without the %d with singleThreadExecutor?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -117,7 +130,7 @@
 
     _clientId = transportProviderProperties.getProperty(ProducerConfig.CLIENT_ID_CONFIG);
     if (_clientId == null || _clientId.isEmpty()) {
```
DEEPTHIKORAT, line 132: nit: Should we use StringUtils ?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -221,47 +234,63 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
 
   private DatastreamRuntimeException generateSendFailure(Exception exception) {
     _dynamicMetricsManager.createOrUpdateMeter(_metricsNamesPrefix, AGGREGATE, PRODUCER_ERROR, 1);
     if (exception instanceof IllegalStateException) {
-      _log.warn("sent failure transiently, exception: ", exception);
+      _log.warn("send failed transiently with exception: ", exception);
       return new DatastreamTransientException(exception);
     } else {
-      _log.warn("sent failure, restart producer, exception: ", exception);
+      _log.warn("send failed, restart producer, exception: ", exception);
```
DEEPTHIKORAT, line 282: nit: should this read: "send failed with a non-transient exception. Shutting down producer" to match what is being done in the method.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
somandal: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
somandal: as discussed, there is no `setName` API for the  ThreadFactoryBuilder,  so leaving this as is.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
somandal: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
somandal: Went back and forth in my head on this. `throws handleSendFailure` looks very weird to me. So leaving this as is.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
somandal: done

PR #721, opened 2020-06-15
Remove leading and trailing whitespaces from datastream name on creation

ahmedahamid:


datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -761,6 +761,14 @@ public CreateResponse create(Datastream datastream) {
         metadataMap.put(DatastreamMetadataConstants.IS_USER_MANAGED_DESTINATION_KEY, "true");
       }
 
+      String datastreamName = datastream.getName();
+      String trimmedDatastreamName = (datastreamName == null) ? datastreamName : datastreamName.trim();
```
somandal, line 765: can  datastreamName be null?

datastream-server-restli/src/test/java/com/linkedin/datastream/server/dms/TestDatastreamResources.java
```diff
@@ -559,11 +559,19 @@ public void testCreateDatastream() throws Exception {
     Set<String> missingFields = new HashSet<>();
 
     // happy path
-    Datastream fullDatastream = generateDatastream(1);
+    Datastream fullDatastream = generateDatastream(0);
     CreateResponse response = resource.create(fullDatastream);
     Assert.assertNull(response.getError());
     Assert.assertEquals(response.getStatus(), HttpStatus.S_201_CREATED);
 
+    Datastream whitespaceDatastream = generateDatastream(1);
```
DEEPTHIKORAT, line 567: nit and feel free to ignore" add this as its own test. easy to spot perhaps?

ahmedahamid: You're right. I'll add a comment to make it stand out instead because this test method contains all the datastream creation test cases.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
ahmedahamid: Removed the null check. Turns out this is already covered by this call performed at the beginning of this method:
```
DatastreamUtils.validateNewDatastream(datastream);
```

PR #723, opened 2020-06-22
Fix hash and equals for DatstreamTaskImpl

vmaheshw:
DatastreamTaskImpl has ArrayList for partitions and partitionsV2. The ordering can be a issue in comparison if the list is not sorted, event though the list is same, just differently ordered.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
@@ -11,6 +11,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -88,6 +89,8 @@ public class DatastreamTaskImpl implements DatastreamTask {
   private DatastreamEventProducer _eventProducer;
   private String _transportProviderName;
   private SerDeSet _destinationSerDes = new SerDeSet(null, null, null);
+  //precalculated hash
+  private int _hashValue;
 
   /**
    * Constructor for DatastreamTaskImpl.
@@ -99,6 +102,7 @@ public class DatastreamTaskImpl implements DatastreamTask {
     _partitions = new ArrayList<>();
     _partitionsV2 = new ArrayList<>();
     _dependencies = new ArrayList<>();
+    computeHash();
   }
 
   /**
@@ -147,6 +151,7 @@ public class DatastreamTaskImpl implements DatastreamTask {
     }
     LOG.info("Created new DatastreamTask " + this);
     _dependencies = new ArrayList<>();
+    computeHash();
   }
 
 
@@ -178,6 +183,7 @@ public class DatastreamTaskImpl implements DatastreamTask {
 
     _dependencies = new ArrayList<>();
     _dependencies.add(predecessor.getDatastreamTaskName());
+    computeHash();
   }
 
     /**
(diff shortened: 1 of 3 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
```diff
@@ -99,6 +102,7 @@ public DatastreamTaskImpl() {
     _partitions = new ArrayList<>();
     _partitionsV2 = new ArrayList<>();
     _dependencies = new ArrayList<>();
+    computeHash();
```
ahmedahamid, line 105: While I understand your concern, it's fairly daunting and error-prone to have to call `computeHash()` in (almost) every single method that could mutate this object (but not all of them). Precomputing hashcode is mostly useful when the type is immutable. I think it'd be more straightforward to just compute the hashcode on demand, especially that most — if not all of — Brooklin's code is not CPU bound anyways.

datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java
```diff
@@ -109,4 +110,26 @@ public void testErrorFromZkJson() throws Exception {
     DatastreamTaskImpl task2 = DatastreamTaskImpl.fromJson(json);
     task2.getDatastreams();
   }
+
+  @Test
+  public void testDatastreamTaskComparison() {
```
ahmedahamid, line 115: Not asking you to test every single field but it would be useful to add comprehensive tests for the general contract of `equals()` and `hashcode()` for this type, i.e.

#### equals()
> The equals method implements an equivalence relation on non-null object references:
> - It is reflexive: for any non-null reference value x, x.equals(x) should return true.
> - It is symmetric: for any non-null reference values x and y, x.equals(y) should return true if and only if y.equals(x) returns true.
> - It is transitive: for any non-null reference values x, y, and z, if x.equals(y) returns true and y.equals(z) returns true, then x.equals(z) should return true.
> - It is consistent: for any non-null reference values x and y, multiple invocations of x.equals(y) consistently return true or consistently return false, provided no information used in equals comparisons on the objects is modified.
> - For any non-null reference value x, x.equals(null) should return false.

#### hashcode()
> The general contract of hashCode is:
> - Whenever it is invoked on the same object more than once during an execution of a Java application, the hashCode method must consistently return the same integer, provided no information used in equals comparisons on the object is modified. This integer need not remain consistent from one execution of an application to another execution of the same application.
> - If two objects are equal according to the equals(Object) method, then calling the hashCode method on each of the two objects must produce the same integer result.
> - It is not required that if two objects are unequal according to the equals(Object) method, then calling the hashCode method on each of the two objects must produce distinct integer results. However, the programmer should be aware that producing distinct integer results for unequal objects may improve the performance of hash tables.

I understand that accounting for each one of the points above could take a lot of effort. Feel free to pick the ones that feel more critical (e.g. `hashcode()`'s contract item #2)

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
somandal: +1

vmaheshw: Done.

datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java
vmaheshw: Done

PR #725, opened 2020-06-25
Add support for Producer flush() with a configurable timeout

somandal:
This PR adds support for Kafka Producer flush with a timeout which can be configured via config. It also involves some refactoring.

### DIFF
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -60,15 +60,30 @@ class KafkaProducerWrapper<K, V> {
   @VisibleForTesting
   static final String PRODUCER_COUNT = "producerCount";
 
+  // Default producer configuration for no data loss pipeline.
+  private static final String DEFAULT_PRODUCER_ACKS_CONFIG_VALUE = "all";
+  private static final String DEFAULT_MAX_BLOCK_MS_CONFIG_VALUE = String.valueOf(Integer.MAX_VALUE);
+  private static final String DEFAULT_MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION_VALUE = "1";
+
+  private static final long DEFAULT_SEND_FAILURE_RETRY_WAIT_MS = Duration.ofSeconds(5).toMillis();
+  private static final int DEFAULT_PRODUCER_FLUSH_TIMEOUT_MS = Integer.MAX_VALUE;
+  private static final Double DEFAULT_RATE_LIMITER = 0.1;
+
+  private static final String CFG_SEND_FAILURE_RETRY_WAIT_MS = "send.failure.retry.wait.time.ms";
+  private static final String CFG_KAFKA_PRODUCER_FACTORY = "kafkaProducerFactory";
+  private static final String CFG_RATE_LIMITER_CFG = "producerRateLimiter";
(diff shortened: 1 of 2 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -60,15 +60,30 @@
   @VisibleForTesting
```
vmaheshw, line 60: Since you did the refactoring, can you please take of this as well ? May be move down after all private static final.

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaProducerWrapper.java
```diff
@@ -150,7 +151,7 @@ void verifySend(int numExpected) {
     }
 
     void verifyFlush(int numExpected) {
-      verify(_mockProducer, times(numExpected)).flush();
+      verify(_mockProducer, times(numExpected)).flush(anyInt(), any(TimeUnit.class));
```
vmaheshw, line 154: Do you see a value in adding a test to ensure that flush timeout is honored?

vmaheshw: One example is here https://github.com/linkedin/brooklin/pull/676/files where it can throw timeoutException. See if this is useful.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -60,15 +60,30 @@
...
+  private static final String CFG_SEND_FAILURE_RETRY_WAIT_MS = "send.failure.retry.wait.time.ms";
+  private static final String CFG_KAFKA_PRODUCER_FACTORY = "kafkaProducerFactory";
+  private static final String CFG_RATE_LIMITER_CFG = "producerRateLimiter";
+  private static final String CFG_PRODUCER_FLUSH_TIMEOUT_MS = "producerFlushTimeoutMs";
+
   private static final AtomicInteger NUM_PRODUCERS = new AtomicInteger();
   private static final Supplier<Integer> PRODUCER_GAUGE = NUM_PRODUCERS::get;
 
-  private static final int TIME_OUT = 2000;
+  private static final int CLOSE_TIME_OUT_MS = 2000;
```
ahmedahamid, line 80: Not your fault nit: `TIMEOUT`

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaProducerWrapper.java
somandal: As discussed, adding a test similar to what you had in that PR is hard because we are relying on Kafka's internal timeout  mechanism, unlike in your earlier PR where the `KafkaProducerWrapper` itself was responsible for timing out the  flush. Instead, added a test to  validate that we handle TimeoutException as expected.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
somandal: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
somandal: done

PR #728, opened 2020-07-13
Add isLeader check for Leader only events

vmaheshw:
Adding check while processing events which are supposed to be processed by leader only. This is to ensure that the follower does not process the leader events in case of session expiry.


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -777,6 +777,10 @@ private boolean isDeletingOrExpired(Datastream stream) {
    * either. Also, expired streams are excluded from any future task assignments.
    */
   private void handleDatastreamAddOrDelete() {
+    if (!_adapter.isLeader()) {
+      _log.warn("skipping handleDatastreamAddOrDelete isLeader {}", _adapter.isLeader());
+      return;
+    }
```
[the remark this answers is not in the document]
vmaheshw: done.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -663,7 +663,11 @@ private void assignSerdes(DatastreamTaskImpl datastreamTask) {
 
   protected synchronized void handleEvent(CoordinatorEvent event) {
     _log.info("START: Handle event " + event.getType() + ", Instance: " + _adapter.getInstanceName());
-
+    boolean isLeader = _adapter.isLeader();
+    if (skipLeaderEvent(event.getType(), isLeader)) {
```
ahmedahamid, line 667: nit: I think this'd be more intuitive/straightforward if expressed as:
```java
if (isLeaderEvent(event) && !isLeader) {
  _log.info(...);
  return;
}
```

vmaheshw: done

PR #732, opened 2020-07-20
Support for Kafka broker message format bump for passthrough BMM

somandal:
This PR adds support for the Kafka broker message format bump for passthrough by extracting the magic from the PassthroughConsumerRecord. This magic is stored in the metadata of the BrooklinEnvelope after consumption. A utility class is added for abstracting out conversion of byte to String and back to byte. The transport layer is responsible for extracting this metadata field and constructing the right ProducerRecord with headers if the metadata is present. This is done in the KafkaTransportProvider class.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
@@ -26,6 +26,7 @@ import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.clients.consumer.PassThroughConsumerRecord;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.record.TimestampType;
 import org.slf4j.Logger;
@@ -48,6 +49,8 @@ import com.linkedin.datastream.connectors.kafka.KafkaConnectionString;
 import com.linkedin.datastream.connectors.kafka.KafkaDatastreamStatesResponse;
 import com.linkedin.datastream.connectors.kafka.PausedSourcePartitionMetadata;
 import com.linkedin.datastream.connectors.kafka.TopicPartitionUtil;
+import com.linkedin.datastream.kafka.KafkaDatastreamMetadataConstants;
+import com.linkedin.datastream.kafka.KafkaPassthroughRecordMagicConverter;
 import com.linkedin.datastream.kafka.factory.KafkaConsumerFactory;
 import com.linkedin.datastream.metrics.BrooklinCounterInfo;
 import com.linkedin.datastream.metrics.BrooklinGaugeInfo;
@@ -241,6 +244,16 @@ public class KafkaMirrorMakerConnectorTask extends AbstractKafkaBasedConnectorTa
     String offsetStr = String.valueOf(offset);
     metadata.put(KAFKA_ORIGIN_OFFSET, offsetStr);
     metadata.put(BrooklinEnvelopeMetadataConstants.EVENT_TIMESTAMP, String.valueOf(eventsSourceTimestamp));
+    if (Boolean.TRUE.toString()
+        .equals(_datastream.getMetadata().get(KafkaDatastreamMetadataConstants.USE_PASSTHROUGH_COMPRESSION))) {
+      // If passthrough mode is enabled, we need to create a Kafka header on the transport side for supporting
+      // Kafka broker message format bump. The magic byte contains details about the message format for the passthrough
+      // record and it needs to be preserved and set via the Kafka headers to ensure that the correct message format
+      // can be negotiated.
+      PassThroughConsumerRecord<?, ?> passThroughConsumerRecord = (PassThroughConsumerRecord<?, ?>) fromKafka;
+      metadata.put(KafkaPassthroughRecordMagicConverter.PASS_THROUGH_MAGIC_VALUE,
+          KafkaPassthroughRecordMagicConverter.convertMagicToString(passThroughConsumerRecord.magic()));
+    }
     BrooklinEnvelope envelope = new BrooklinEnvelope(fromKafka.key(), fromKafka.value(), null, metadata);
     DatastreamProducerRecordBuilder builder = new DatastreamProducerRecordBuilder();
     builder.addEvent(envelope);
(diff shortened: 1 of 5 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -241,6 +244,16 @@ protected DatastreamProducerRecord translate(ConsumerRecord<?, ?> fromKafka, Ins
...
     metadata.put(KAFKA_ORIGIN_OFFSET, offsetStr);
     metadata.put(BrooklinEnvelopeMetadataConstants.EVENT_TIMESTAMP, String.valueOf(eventsSourceTimestamp));
+    if (Boolean.TRUE.toString()
+        .equals(_datastream.getMetadata().get(KafkaDatastreamMetadataConstants.USE_PASSTHROUGH_COMPRESSION))) {
+      // If passthrough mode is enabled, we need to create a Kafka header on the transport side for supporting
+      // Kafka broker message format bump. The magic byte contains details about the message format for the passthrough
+      // record and it needs to be preserved and set via the Kafka headers to ensure that the correct message format
+      // can be negotiated.
+      PassThroughConsumerRecord<?, ?> passThroughConsumerRecord = (PassThroughConsumerRecord<?, ?>) fromKafka;
+      metadata.put(KafkaPassthroughRecordMagicConverter.PASS_THROUGH_MAGIC_VALUE,
```
MayureshGharat, line 254: This patch LGTM. Any reason, you want to convert this to a String and not use it as it is?

MayureshGharat:
LGTM. Left a minor comment/question.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
somandal: @MayureshGharat we convert ConsumerRecord to a `DatastreamProducerRecord`, and `DatastreamProdcuerRecord` has a list  of `BrooklinEnvelope`. `BrooklinEnvelope` has a generic field which stores all "metadata" and is a `Map<String, String>` . If I didn't convert the magic to String to store in this generic "metadata", I'd have to introduce a new field to store the magic as a `byte`. Which isn't a bad option, but this will only be used by BMM passthrough. 

@DEEPTHIKORAT and @jzakaryan can you folks also comment on whether using "metadata" and doing string conversion makes more sense, or it's okay to introduce this `byte` field to store the magic? The only use case would be BMM passthrough.

jzakaryan: Here's my two cents. @somandal I wouldn't introduce a "magic" byte as a separate field, because it only makes sense in the scope of BMM passthrough, and it would be useless and confusing out of that limited scope.
The only precaution with the current approach would be the conversion overhead, but since it's just being done once for a batch, it should be fine.

somandal: @jzakaryan thank you!

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -241,6 +244,16 @@ protected DatastreamProducerRecord translate(ConsumerRecord<?, ?> fromKafka, Ins
     String offsetStr = String.valueOf(offset);
     metadata.put(KAFKA_ORIGIN_OFFSET, offsetStr);
     metadata.put(BrooklinEnvelopeMetadataConstants.EVENT_TIMESTAMP, String.valueOf(eventsSourceTimestamp));
+    if (Boolean.TRUE.toString()
```
DEEPTHIKORAT, line 247: nit: would it make sense to have a member field for isPassThrough to avoid checking this for each translate?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaPassthroughRecordMagicConverter.java
```diff
@@ -0,0 +1,64 @@
...
+      throw new DatastreamRuntimeException(String.format("Cannot convert magic byte %02X to String", magic), e);
+    }
+  }
+
+  /**
+   * Convert the converted magic String back into a byte array
+   */
+  public static byte[] convertMagicStringToByteArray(String magic) {
+    try {
+      return magic.getBytes(CHARSET);
```
DEEPTHIKORAT, line 42: Can it return a single byte instead? since we are using ISO-8 we can be sure not to get multiple bytes. Makes the APIs look consistent in the encode and decode with taking 1 byte and returning 1 byte.

DEEPTHIKORAT: oh nvm, looks like RecordHeader does require a byte array.

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaPassthroughRecordMagicConverter.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Tests for {@link KafkaPassthroughRecordMagicConverter}.
+ */
+@Test
+public class TestKafkaPassthroughRecordMagicConverter {
+
+  @Test
+  public void testMagicConversion() {
+    String magicValue0 = KafkaPassthroughRecordMagicConverter.convertMagicToString(RecordBatch.MAGIC_VALUE_V0);
```
DEEPTHIKORAT, line 21: nit and feel free to ignore, but this one might read better as a for loop.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
somandal: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaPassthroughRecordMagicConverter.java
somandal: yeah, that's why I decided to return a byte[] instead of byte.

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaPassthroughRecordMagicConverter.java
somandal: done

PR #733, opened 2020-07-21
Log and explicitly send 404 in get datastream rest call

DEEPTHIKORAT:
`DatastreamResources` get call does not log an error when stream is not
found. This is not consistent with other APIs. Make it consistent with
the rest of the APIs which allows for also logging in case stream is not
found and can be helpful with debugging. Change also removes the logging
of time it takes for the get for consistency. It is not a batch operation and
doesn't seem like it warrants a log for execution time.

### DIFF
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -666,21 +666,22 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
 
   @Override
   public Datastream get(String name) {
+    Datastream stream = null;
     try {
-      LOG.info("Get datastream called for datastream {}", name);
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, GET_CALL, 1);
-      Instant startTime = Instant.now();
-      Datastream stream = _store.getDatastream(name);
-      LOG.info("Get datastream call took {} ms", Duration.between(startTime, Instant.now()).toMillis());
-      return stream;
+      stream = _store.getDatastream(name);
     } catch (Exception e) {
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, CALL_ERROR, 1);
       _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_500_INTERNAL_SERVER_ERROR,
           "Get datastream failed for datastream: " + name, e);
     }
 
-    // Returning null will automatically trigger a 404 Not Found response
-    return null;
+    if (stream == null) {
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_404_NOT_FOUND,
+          "Datastream not found: " + name);
+    }
+
+    return stream;
   }
 
   @SuppressWarnings("deprecated")
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -668,6 +668,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
   public Datastream get(String name) {
     Datastream stream = null;
     try {
+      LOG.info("Get datastream called for datastream {}", name);
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, GET_CALL, 1);
       stream = _store.getDatastream(name);
     } catch (Exception e) {
--- a/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
+++ b/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
@@ -2194,7 +2194,15 @@ public class TestCoordinator {
     Assert.assertEquals(createResponse.getStatus(), HttpStatus.S_201_CREATED);
 
     // Poll up to 30s for stream1 to get deleted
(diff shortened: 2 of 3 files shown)


datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -666,21 +666,22 @@ public UpdateResponse delete(String datastreamName) {
 
   @Override
   public Datastream get(String name) {
+    Datastream stream = null;
     try {
-      LOG.info("Get datastream called for datastream {}", name);
```
somandal, line 670: nit: maybe we should keep this log? That Get was called? I know that I've used this log for other operations we perform (stop, resume, pause, etc), and it may be helpful to know if the server is receiving Get requests or not if we see some failures.

DEEPTHIKORAT: Sorry, wasn't intended. Fixed it.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -666,21 +666,23 @@ public UpdateResponse delete(String datastreamName) {
...
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, CALL_ERROR, 1);
       _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_500_INTERNAL_SERVER_ERROR,
           "Get datastream failed for datastream: " + name, e);
     }
 
-    // Returning null will automatically trigger a 404 Not Found response
-    return null;
+    if (stream == null) {
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_404_NOT_FOUND,
+          "Datastream not found: " + name);
```
ahmedahamid, line 682: I scoured Rest.li's docs and source code to see if throwing an exception was idiomatic for 404s. The javadocs of this method in of the parent types of `DatastreamResources`, i.e. `CollectionResource`, states that `null` is expected in case the resource isn't found:
```
  /**
   * Gets this resource.
   *
   * @param key resource key
   * @return null if resource was not found
   */
  V get(K key);
```
Throwing a `RestLiServiceException` seems to be intended for unexpected server errors. 

I think we can just adjust logging in the current code to print whether a datastream was found or not.

DEEPTHIKORAT: Makes sense. I reverted the change to throw. But what counts for unexpected server errors? For example the `delete` API doc says
```  /**
   * Deletes this resource.
   *
   * @param key resource key
   * @return the status or null if for a default, 204 No Content response
   */
``` 
Does that mean, other APIs also should be returning a null when the look up of the stream to operate on fails?

ahmedahamid: > But what counts for unexpected server errors?

Getting a network exception, interrupted exception, failing to find a required file on disk, no permission to access resource ... etc — situations where your web-app is running into an unexpected problem.

> Does that mean, other APIs also should be returning a null when the look up of the stream to operate on fails?

Rest.li seems to interpret `null` differently for the different HTTP verbs. In the case of delete for instance, returning `null` actually indicates success. [204](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) is a success code. This is why the method you referenced offers this convenience facility of returning `null` if you want to say deletion was successful, instead of having to construct an `UpdateResponse` with a 204 status code yourself.

Accordingly, our `DatastreamResources.delete()` should just return an `UpdateRespose` with a 404 status code if a datastream isn't found; there's no service error here.

ahmedahamid: I looked into Rest.li's source code and [here](https://github.com/linkedin/rest.li/blob/master/restli-server/src/main/java/com/linkedin/restli/internal/server/response/RestLiResponseHandler.java#L181) is what I found:
- Rest.li translates a `null` return value into a 200 (OK) response for methods whose HTTP method is `POST`
- It also translates a `null` return value into a 404 (NOT FOUND) response for methods whose HTTP method is `GET`

However, it does not translate `null` for methods whose HTTP method is `DELETE` (e.g. `DatastreamResources.delete()`). So, we can't return `null` to indicate success from a delete method. We'd have to return `UpdateResponse(HttpStatus.S_404_NOT_FOUND)`.

But everything we've discussed so far still holds.

ahmedahamid: Just documenting a conversation you and I had offline in case anyone refers to this thread in the future. 

You brought it to my attention that the the same piece of code I referenced did actually throw a `RestLiServiceException` when translating `null` from `GET` methods into 400 ([code](https://github.com/linkedin/rest.li/blob/master/restli-server/src/main/java/com/linkedin/restli/internal/server/response/RestLiResponseHandler.java#L202)). I looked further and found that Rest.li translates all thrown `RestLiServiceException`s ([code](https://github.com/linkedin/rest.li/blob/30d525f4943bb2f136cab7891daa85a2120add4f/restli-server/src/main/java/com/linkedin/restli/internal/server/response/ErrorResponseBuilder.java#L169)) for all HTTP verbs according to the status codes they carry, which invalidates the assumption I was making that these exceptions were reserved for internal server errors. 

Out of consistency with our existing `DatastreamResources.delete()`, which throws if a datastream isn't found, we decided to keep the current logic in your PR as is.

Please, excuse the back and forth. I am learning about Rest.li as we go.

PR #734, opened 2020-07-22
Add Future Utils Library

vmaheshw:
We need a utility to abort the future after a specified duration. This is a general purpose library to add time bound to any Future. This will help in cancelling the connector task Future within specified duration.

### DIFF
--- a/datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
+++ b/datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
@@ -6,38 +6,68 @@
 package com.linkedin.datastream.common;
 
 import java.time.Duration;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.lang.NullArgumentException;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+
+/**
+ * Utilities to work with {@link Future}
+ */
 public class FutureUtils {
 
-  private static final ScheduledExecutorService SCHEDULED_EXECUTOR_SERVICE = Executors.newScheduledThreadPool(100);
+  private static final ScheduledExecutorService SCHEDULED_EXECUTOR_SERVICE = Executors.newScheduledThreadPool(4,
+      new ThreadFactoryBuilder().setDaemon(true).setNameFormat("FutureUtils-%d").build());
 
+  private FutureUtils() {
+
+  }
   /**
-   * This function waits till timeout for Future to complete, otherwise, it cancels it.
-   * @param futureToAwait Future to complete/stop
-   * @param timeout time to wait before aborting future
-   * @return CompletableFuture for this operation
+   * Check if {@code future} is done after {@code after}. Otherwise, cancels {@code future}.
+   * It returns {@link CompletableFuture} which should be used directly for light-weight operations to avoid blocking the
+   * executor thread, or call {@link CompletableFuture#supplyAsync(Supplier)} or {@link CompletableFuture#supplyAsync(Supplier, Executor)}
+   * which uses another thread to execute and does not block the main executor thread.
+   * @param future {@code future} to check/cancel
+   * @param after time to wait before aborting {@code future}
+   * @return {@link CompletableFuture} completed with computed result of {@code future}, if {@code future} was done before {@code after}.
+   *         {@link CompletableFuture} completed exceptionally with {@link InterruptedException}, if {@code future} was interrupted.
+   *         {@link CompletableFuture} completed exceptionally with cause of Exception, if {@code future} hit {@link ExecutionException}.
+   *         {@link CompletableFuture} cancelled, if {@code future} thread was cancelled.
+   * @throws NullArgumentException {@link NullArgumentException} is Thrown if future is Null
    */
-  public static CompletableFuture<?> awaitFutureOrCancel(Future<?> futureToAwait, Duration timeout) {
-    CompletableFuture<?> result = new CompletableFuture<>();
+  public static <T> CompletableFuture<T> getIfDoneOrCancel(Future<T> future, Duration after) throws
+                                                                                             NullArgumentException {
+    if (future == null) {
+      throw new NullArgumentException("future");
+    }
+
+    CompletableFuture<T> result = new CompletableFuture<>();
 
     SCHEDULED_EXECUTOR_SERVICE.schedule(() -> {
-      if ((futureToAwait == null) || futureToAwait.isDone()) {
-        result.complete(null);
-      } else {
-        boolean cancel = futureToAwait.cancel(true);
-        if (!cancel) {
-          result.cancel(true);
-        } else {
-          result.complete(null);
+      if (future.isDone()) {
+        try {
+          T t = future.get();
+          result.complete(t);
+        } catch (InterruptedException e) {
+          result.completeExceptionally(e);
+        } catch (ExecutionException e) {
+          result.completeExceptionally(e.getCause());
+        } catch (CancellationException e) {
+          result.cancel(false);
         }
+      } else {
+        future.cancel(true);
+        result.cancel(false);
       }
-    }, timeout.toMillis(), TimeUnit.MILLISECONDS);
+    }, after.toMillis(), TimeUnit.MILLISECONDS);
     return result;
   }
 }
--- a/datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
+++ b/datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
@@ -1,5 +1,5 @@
 /**
- *  Copyright 2019 LinkedIn Corporation. All rights reserved.
+ *  Copyright 2020 LinkedIn Corporation. All rights reserved.
  *  Licensed under the BSD 2-Clause License. See the LICENSE file in the project root for license information.
  *  See the NOTICE file in the project root for additional information regarding copyright ownership.
  */
@@ -14,7 +14,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
-import org.apache.commons.lang.NullArgumentException;
+import org.apache.commons.lang3.Validate;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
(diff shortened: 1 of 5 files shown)


datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private static final ScheduledExecutorService SCHEDULED_EXECUTOR_SERVICE = Executors.newScheduledThreadPool(100);
+
+  /**
+   * This function waits till timeout for Future to complete, otherwise, it cancels it.
+   * @param futureToAwait Future to complete/stop
+   * @param timeout time to wait before aborting future
+   * @return CompletableFuture for this operation
+   */
+  public static CompletableFuture<?> awaitFutureOrCancel(Future<?> futureToAwait, Duration timeout) {
```
ahmedahamid, line 26: I know our discussions have been mostly focused on the specific use-case that motivated the introduction of this API. However, we need to consider how such a fundamental utility will be used in general.

For instance:
1. If I'm calling this API, wouldn't I naturally find it useful if it offered me a way to access the result of `futureToAwait` in case it completed successfully? This should be fairly straightforward for this API to accomplish because all it takes in this case is calling `futureToAwait.get()`, sparing the caller this effort.

2. Wouldn't it be useful if the returned `CompletableFuture` communicated the success/failure/cancellation of `futureToAwait` instead? One of the annoying things about `Future` in Java is the lack of composability; there's no straightforward way to chain other futures to it. However, this isn't the case with `CompletableFuture`. So, as a caller, I would imagine composing the returned future to react to success and cancellation/failure of the future I passed. This would allow me to do:
```java
    CompletableFuture<?> result = awaitFutureOrCancel(future, timeout);
    result
        .exceptionally()   // future didn't complete successfully (threw or got cancelled)
        .thenRun();        // future completed successfully
```

Here's a proposed impl that addresses these two issues:
```java
  public static <T> CompletableFuture<T> awaitFutureOrCancel(Future<T> future, Duration timeout) {
    CompletableFuture<T> result = new CompletableFuture<>();

    SCHEDULED_EXECUTOR_SERVICE.schedule(() -> {
      if (future.isDone()) {
        try {
          T t = future.get();
          result.complete(t);
        } catch (InterruptedException e) {
          result.completeExceptionally(e);
        } catch (ExecutionException e) {
          result.completeExceptionally(e.getCause());
        } catch (CancellationException e) {
          result.cancel(false);
        }
      } else {
        future.cancel(true);
        result.cancel(false);
      }
    }, timeout.toMillis(), TimeUnit.MILLISECONDS);
    return result;
  }
```

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+    CompletableFuture<?> result = new CompletableFuture<>();
+
+    SCHEDULED_EXECUTOR_SERVICE.schedule(() -> {
+      if ((futureToAwait == null) || futureToAwait.isDone()) {
```
ahmedahamid, line 30: I'd rather disallow null `futureToAwait`

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
ahmedahamid, line 26: This comment is purely about naming.

- I think it'd be better to rename `futureToAwait` to just `future`
- If you decide to go with an implementation that propagates back the future's result, I'd consider giving the method a different name, one that reflects that it is meant to get the future's result (if possible) after a specific duration (e.g. `getIfDoneOrCancel`?)
- `timeout` is a misleading name for that param cause it implies we may check/finish sooner. We need a name that convey that this is how much we'll wait (should we call it `after`?)

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class FutureUtils {
+
+  private static final ScheduledExecutorService SCHEDULED_EXECUTOR_SERVICE = Executors.newScheduledThreadPool(100);
+
+  /**
+   * This function waits till timeout for Future to complete, otherwise, it cancels it.
+   * @param futureToAwait Future to complete/stop
+   * @param timeout time to wait before aborting future
+   * @return CompletableFuture for this operation
```
ahmedahamid, line 24: The javadocs can be improved for greater accuracy so users don't have to grok the code themselves.

- The description can be something like: 
```java
/**
 * Check if {@code future} is done after {@code timeout} ... (and the rest of the story)
```
- `Future to complete/stop` - check/cancel?
- `before aborting future` - before checking actually
- We'd need to document when and how the returned future is completed successfully, completed w/ an exception, or cancelled.

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+  public static CompletableFuture<?> awaitFutureOrCancel(Future<?> futureToAwait, Duration timeout) {
+    CompletableFuture<?> result = new CompletableFuture<>();
+
+    SCHEDULED_EXECUTOR_SERVICE.schedule(() -> {
+      if ((futureToAwait == null) || futureToAwait.isDone()) {
+        result.complete(null);
+      } else {
+        boolean cancel = futureToAwait.cancel(true);
+        if (!cancel) {
+          result.cancel(true);
```
ahmedahamid, line 35: Change to `false`; `true` makes little sense for `CompletableFuture`

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+   * @return CompletableFuture for this operation
+   */
+  public static CompletableFuture<?> awaitFutureOrCancel(Future<?> futureToAwait, Duration timeout) {
+    CompletableFuture<?> result = new CompletableFuture<>();
+
+    SCHEDULED_EXECUTOR_SERVICE.schedule(() -> {
+      if ((futureToAwait == null) || futureToAwait.isDone()) {
+        result.complete(null);
+      } else {
+        boolean cancel = futureToAwait.cancel(true);
```
ahmedahamid, line 33: nit: inline

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        if (!cancel) {
```
ahmedahamid, line 34: nit: invert if/else for simplicity

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * This function waits till timeout for Future to complete, otherwise, it cancels it.
+   * @param futureToAwait Future to complete/stop
+   * @param timeout time to wait before aborting future
+   * @return CompletableFuture for this operation
+   */
+  public static CompletableFuture<?> awaitFutureOrCancel(Future<?> futureToAwait, Duration timeout) {
+    CompletableFuture<?> result = new CompletableFuture<>();
+
+    SCHEDULED_EXECUTOR_SERVICE.schedule(() -> {
```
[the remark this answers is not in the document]
ahmedahamid: True. The problem here's there's no easy way to compose that separate timeout future with the provided `Future` since it's not a `CompletableFuture`.

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+
+public class FutureUtils {
+
+  private static final ScheduledExecutorService SCHEDULED_EXECUTOR_SERVICE = Executors.newScheduledThreadPool(100);
```
ahmedahamid, line 18: It would be useful to give this executor a thread factory so we can give its threads meaningful names.

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,43 @@
+/**
+ *  Copyright 2019 LinkedIn Corporation. All rights reserved.
```
vishwajith-s, line 2: Copyright need to be updated.

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+
+/**
+ * Utilities to work with {@link Future}
+ */
+public class FutureUtils {
+
+  private static final ScheduledExecutorService SCHEDULED_EXECUTOR_SERVICE = Executors.newScheduledThreadPool(4,
```
vishwajith-s, line 27: Can we have this number 4 as a static int?

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,73 @@
...
+   * @param future {@code future} to check/cancel
+   * @param after time to wait before aborting {@code future}
+   * @return {@link CompletableFuture} completed with computed result of {@code future}, if {@code future} was done before {@code after}.
+   *         {@link CompletableFuture} completed exceptionally with {@link InterruptedException}, if {@code future} was interrupted.
+   *         {@link CompletableFuture} completed exceptionally with cause of Exception, if {@code future} hit {@link ExecutionException}.
+   *         {@link CompletableFuture} cancelled, if {@code future} thread was cancelled.
+   * @throws NullArgumentException {@link NullArgumentException} is Thrown if future is Null
+   */
+  public static <T> CompletableFuture<T> getIfDoneOrCancel(Future<T> future, Duration after) throws
+                                                                                             NullArgumentException {
```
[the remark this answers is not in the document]
vishwajith-s: +1

datastream-utils/src/main/java/com/linkedin/datastream/common/FutureUtils.java
```diff
@@ -0,0 +1,73 @@
...
+    if (future == null) {
+      throw new NullArgumentException("future");
+    }
```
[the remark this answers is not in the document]
vishwajith-s: +1

PR #735, opened 2020-07-22
Added diag endpoint for consumer offsets for bmm

jzakaryan:
This pull request builds on top of the existing diagnostics infrastructure and adds support for consumer offsets per topic partition.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
@@ -101,6 +101,7 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
   enum DiagnosticsRequestType {
     DATASTREAM_STATE,
     PARTITIONS,
+    CONSUMER_OFFSETS
   }
 
   /**
@@ -401,6 +402,10 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
         String response = processTopicPartitionStatsRequest();
         _logger.trace("Query: {} returns response: {}", query, response);
         return response;
+      } else if (path != null && path.equalsIgnoreCase(DiagnosticsRequestType.CONSUMER_OFFSETS.toString())) {
+        String response = processConsumerOffsetsRequest();
+        _logger.trace("Query: {} returns response: {}", query, response);
+        return response;
       } else {
         _logger.warn("Could not process query {} with path {}", query, path);
       }
@@ -466,6 +471,21 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
     return JsonUtils.toJson(serializedResponses);
   }
 
+  private String processConsumerOffsetsRequest() {
+    _logger.info("process consumer stats request");
+    List<KafkaConsumerOffsetsResponse> serializedResponses = new ArrayList<>();
+
+    synchronized (_runningTasks) {
+      _runningTasks.forEach((datastreamTask, connectorTaskEntry) -> {
+        KafkaTopicPartitionTracker tracker = connectorTaskEntry.getConnectorTask().getKafkaTopicPartitionTracker();
+        KafkaConsumerOffsetsResponse response = new KafkaConsumerOffsetsResponse(tracker.getConsumerOffsets(),
+            tracker.getConsumerGroupId());
+        serializedResponses.add(response);
+      });
+    }
+    return JsonUtils.toJson(serializedResponses);
+  }
+
   /**
    * Aggregates the responses from all the instances into a single JSON response.
    * Sample query: /datastream_state?datastream=PizzaDatastream
@@ -491,10 +511,12 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
       if (path != null
           && (path.equalsIgnoreCase(DiagnosticsRequestType.DATASTREAM_STATE.toString()))) {
         return JsonUtils.toJson(responses);
-      }
-      if (path != null
+      } else if (path != null
           && (path.equalsIgnoreCase(DiagnosticsRequestType.PARTITIONS.toString()))) {
         return KafkaConnectorDiagUtils.reduceTopicPartitionStatsResponses(responses, _logger);
+      } else if (path != null
+          && (path.equalsIgnoreCase(DiagnosticsRequestType.CONSUMER_OFFSETS.toString()))) {
+        return KafkaConnectorDiagUtils.reduceConsumerOffsetsResponses(responses, _logger);
       }
     } catch (Exception e) {
       _logger.warn("Failed to reduce responses from query {}: {}", query, e.getMessage());
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectorDiagUtils.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectorDiagUtils.java
@@ -79,20 +79,21 @@ public class KafkaConnectorDiagUtils {
       }
 
       responseList.forEach(response -> {
-        if (response.getConsumerOffsets() == null || StringUtils.isBlank(response.getConsumerGroupId())) {
+        if (response.getConsumerOffsets() == null || response.getConsumerOffsets().isEmpty()) {
           logger.warn("Empty consumer offset map from instance {}. Ignoring the result", instance);
-          return;
+        } else if (StringUtils.isBlank(response.getConsumerGroupId())) {
+          logger.warn("Invalid consumer group id from instance {}, Ignoring the result", instance);
+        } else {
+          KafkaConsumerOffsetsResponse reducedResponse = result.computeIfAbsent(response.getConsumerGroupId(),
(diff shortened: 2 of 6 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
```diff
@@ -68,12 +74,48 @@ public void onPartitionsRevoked(@NotNull Collection<TopicPartition> topicPartiti
...
+  public void onPartitionsPolled(@NotNull ConsumerRecords<?, ?> consumerRecords) {
+    Collection<TopicPartition> topicPartitions = consumerRecords.partitions();
+
+    topicPartitions.forEach(topicPartition -> {
+      List<? extends ConsumerRecord<?, ?>> partitionRecords = consumerRecords.records(topicPartition);
+      ConsumerRecord<?, ?> lastRecord = partitionRecords.get(partitionRecords.size() - 1);
+
+      Map<Integer, Long> partitionOffsetMap = _consumerOffsets.putIfAbsent(topicPartition.topic(),
+          new ConcurrentHashMap<>());
+      partitionOffsetMap.put(topicPartition.partition(), lastRecord.offset());
```
jzakaryan, line 103: Here I'm making an assumption that record collection is ordered by offset, and also that it can't be empty. Need to ensure it's true.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectorDiagUtils.java
```diff
@@ -61,4 +61,41 @@ public static String reduceTopicPartitionStatsResponses(Map<String, String> resp
...
+    Map<String, KafkaConsumerOffsetsResponse> result = new HashMap<>();
+
+    responses.forEach((instance, json) -> {
+      List<KafkaConsumerOffsetsResponse> responseList;
+      try {
+        responseList = JsonUtils.fromJson(json, new TypeReference<List<KafkaConsumerOffsetsResponse>>() {
+        });
+      } catch (Exception e) {
+        logger.error("Invalid response {} from instance {}", json, instance);
+        return;
```
vishwajith-s, line 78: Aren't we supposed to return an empty String? Does this compile?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
vishwajith-s: How are we testing these changes?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
```diff
@@ -68,12 +74,48 @@ public void onPartitionsRevoked(@NotNull Collection<TopicPartition> topicPartiti
         }
       }
     });
+
+    // Remove consumer offsets for partitions that have been revoked
```
somandal, line 78: I think this code won't be called if `_enablePartitionAssignment` is true (since this doesn't use consumer group management). These `onPartitionAssigned()/onPartitionRevoked()` callbacks are Kafka callbacks invoked specifically for consumer group management only. For `onPartitionAssigned()`, we manually set up some of the state in `consumerSubscribe()` function for `_enablePartitionAssignment`. I don't see any handling there to revoke TopicPartitions which are no longer assigned for the partition tracker. We need to fix this for the topic manager related stuff in that function too.

Feel free to open a separate bug to handle this correctly (and feel free to take it up or assign it to me). Wanted to capture this potential gap that we may have (needs proper investigation to ensure there is no other code path which is taking care of this already) so that we don't forget to look into it.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -101,6 +101,7 @@ public Thread newThread(@NotNull Runnable r) {
   enum DiagnosticsRequestType {
     DATASTREAM_STATE,
     PARTITIONS,
+    CONSUMER_OFFSETS
```
somandal, line 104: This may be more of a question for @vishwajith-s and @ahmedahamid. What's the recommendation for reusing an existing `DiagnosticsRequestType` vs. adding a new one? I see that some request types return more than one piece of information (such as returning all partitions, paused partitions, auto-paused partitions, and inflight messages as part of a single request).

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectorDiagUtils.java
jzakaryan: Surprisingly, trying to return any String doesn't compile here. That's probably Java related trickery. Not sure why.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
jzakaryan: Will open a ticket to handle that case

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
jzakaryan: I added unit tests.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
```diff
@@ -46,6 +51,7 @@ public KafkaTopicPartitionTracker(String consumerGroupId) {
    * @param topicPartitions the topic partitions which have been assigned
    */
   public void onPartitionsAssigned(@NotNull Collection<TopicPartition> topicPartitions) {
```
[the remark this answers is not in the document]
jzakaryan: Added a comment explaining why I remove offsets when partitions are revoked.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaConsumerOffsets.java
```diff
@@ -0,0 +1,208 @@
...
+    Assert.assertEquals(cg2Response.getConsumerOffsets().get(topic2).keySet().size(), 2); // cg2 consumes 2 partitions for topic 2
+
+    KafkaConsumerOffsetsResponse cg3Response = responseList.stream().
+        filter(r -> r.getConsumerGroupId().equals(consumerGroup3)).findAny().orElse(null);
+    Assert.assertNotNull(cg3Response);
+    Assert.assertEquals(cg3Response.getConsumerOffsets().keySet().size(), 1); // cg3 consumes only topic 2
+    Assert.assertEquals(cg3Response.getConsumerOffsets().get(topic2).size(), 1); // cg3 consumes 1 partition for topic 2
+  }
+
+  private boolean testConsumerOffsetsAreUpdated(KafkaMirrorMakerConnector connector) {
```
somandal, line 175: Feel free to ignore this suggestion, but can we take TOPIC_COUNT, PARTITION_COUNT, and PARTITION_MESSAGE_COUNT as input parameters? Reason I ask is that we may want to add more tests in this area in the future, and it'll be nice to make this a future proof utility.

somandal (dismissed):

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaConsumerOffsets.java
jzakaryan: For now that can easily be configured by changing the values of constants and rebuilding. If we see the need to add more tests here we can easily parameterize those.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectorDiagUtils.java
```diff
@@ -61,4 +61,41 @@ public static String reduceTopicPartitionStatsResponses(Map<String, String> resp
...
+      }
+
+      responseList.forEach(response -> {
+        if (response.getConsumerOffsets() == null || StringUtils.isBlank(response.getConsumerGroupId())) {
```
vishwajith-s, line 82: Should we ignore all the responses from an instance if one result is bad?

jzakaryan: I think you're right @vishwajith-s . Will make changes to just log a warning message and not ignore the results from the rest of instances.

vishwajith-s (approved):

vishwajith-s (approved):

somandal (approved):

PR #736, opened 2020-08-04
Revert "Support for Kafka broker message format bump for passthrough BMM"

somandal:
This reverts the changes done as part of "Support for Kafka broker message format bump for passthrough BMM".
The Kafka clients version which made changes for passthrough support introduced a large performance regression.
This performance regression is seen for passthrough, even though the changes on our end to support passthrough BMM weren't even in.
Also bumping up the LiKafkaVersion to depend on a version where the change has been reverted on Kafka's end.

the contribution guidelines.


gradle/dependency-versions.gradle
```diff
@@ -1,5 +1,5 @@
 ext {
-    LIKafkaVersion = "1.0.41"
+    LIKafkaVersion = "1.0.52"
```
jzakaryan, line 2: @somandal this was not part of the original commit that you're reverting. Just checking if it's intentional.

somandal: Yes this is intentional. Kafka reverted the changes on their end in this version. I just want to bump this up for any OSS users that might be using passthrough so that they don't run into a performance regression.

PR #737, opened 2020-08-06
Added kafka headers in BrooklinEnvelope

jzakaryan:
Right now we're dropping headers that come with Kafka events when consuming from source topic and producing to destination. This poses a problem if downstream systems rely on those headers. Essentially headers are extra metadata that come with events. This pull requests adds a field in BrooklinEnvelope to store those headers with the purpose of setting them in ProducerRecord for transit.

### DIFF
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
@@ -11,7 +11,7 @@ import java.util.Optional;
 
 import org.apache.avro.reflect.Nullable;
 import org.apache.commons.lang.Validate;
-import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.Headers;
 
 
 /**
@@ -30,7 +30,7 @@ public class BrooklinEnvelope {
 
   private Map<String, String> _metadata;
 
-  private Iterable<Header> _headers;
+  private Headers _headers;
 
   /**
    * Construct a BrooklinEnvelope using record key, value, and metadata
@@ -63,13 +63,13 @@ public class BrooklinEnvelope {
    * @param metadata Additional metadata to associate with the change event
    */
   public BrooklinEnvelope(@Nullable Object key, @Nullable Object value, @Nullable Object previousValue,
-      @Nullable Iterable<Header> headers, Map<String, String> metadata) {
+      @Nullable Headers headers, Map<String, String> metadata) {
     Validate.notNull(metadata, "metadata cannot be null");
     setKey(key);
     setValue(value);
     setPreviousValue(previousValue);
-    setMetadata(metadata);
     setHeaders(headers);
+    setMetadata(metadata);
   }
 
   /**
@@ -95,14 +95,14 @@ public class BrooklinEnvelope {
   /**
    * Get the Kafka headers
    */
-  public Iterable<Header> getHeaders() {
+  public Headers getHeaders() {
     return _headers;
   }
 
   /**
    * Set the Kafka headers
    */
-  public void setHeaders(Iterable<Header> headers) {
+  public void setHeaders(Headers headers) {
     _headers = headers;
   }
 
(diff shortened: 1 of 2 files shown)


datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
```diff
@@ -29,14 +30,16 @@
 
   private Map<String, String> _metadata;
 
+  private Iterable<Header> _headers;
```
somandal, line 33: can we instead use `Headers` which is an Iterable of Header? This is what the ConsumerRecord's getHeaders() returns:

`    /**
     * The headers
     */
    public Headers headers() {
        return headers;
    }`

And headers looks like this:

`public interface Headers extends Iterable<Header>`

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
```diff
@@ -48,11 +51,25 @@ public BrooklinEnvelope(Object key, Object value, Map<String, String> metadata)
...
+   * @param metadata Additional metadata to associate with the change event
+   */
+  public BrooklinEnvelope(@Nullable Object key, @Nullable Object value, @Nullable Object previousValue,
+      @Nullable Iterable<Header> headers, Map<String, String> metadata) {
     Validate.notNull(metadata, "metadata cannot be null");
     setKey(key);
     setValue(value);
     setPreviousValue(previousValue);
     setMetadata(metadata);
+    setHeaders(headers);
```
somandal, line 72: nit: can we keep order of function parameters the same as order of setting them? so here move this before setMetadata(metadata)?

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
```diff
@@ -75,6 +92,20 @@ public void setPreviousValue(Object previousValue) {
     _previousValue = previousValue instanceof Optional ? ((Optional<?>) previousValue).orElse(null) : previousValue;
   }
 
+  /**
+   * Get the Kafka headers
+   */
+  public Iterable<Header> getHeaders() {
```
somandal, line 98: if we do change this to store `Headers` directly, we can return `Headers` here too.

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
```diff
@@ -75,6 +92,20 @@ public void setPreviousValue(Object previousValue) {
...
+    return _headers;
+  }
+
+  /**
+   * Set the Kafka headers
+   */
+  public void setHeaders(Iterable<Header> headers) {
```
somandal, line 105: if we do change this to store `Headers` directly, we can pass `Headers` here as the argument.

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
```diff
@@ -48,11 +51,25 @@ public BrooklinEnvelope(Object key, Object value, Map<String, String> metadata)
...
+
+  /**
+   * Construct a {@link BrooklinEnvelope} using record key, value, headers and metadata
+   * @param key The record key (e.g. primary key)
+   * @param value The new record value
+   * @param previousValue The old record value
+   * @param headers Kafka headers to associate with the change event
+   * @param metadata Additional metadata to associate with the change event
+   */
+  public BrooklinEnvelope(@Nullable Object key, @Nullable Object value, @Nullable Object previousValue,
```
ahmedahamid, line 65: Just out of curiosity, how are we planning to fit headers within `DatastreamEvent` and `LiDatastreamEvent`?

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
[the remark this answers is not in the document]
ahmedahamid: Just saw Sonam's [comment](https://github.com/linkedin/brooklin/pull/737#discussion_r466656876), I like her suggestion more actually.

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
somandal: @ahmedahamid we don't need to actually. The ProducerRecord<> takes headers as an argument, so we don't need to modify our own `LiDatastreamEvent` or `DatastreamEvent` AFAIK

PR #740, opened 2020-08-11
Add metric to count number of Nanny Task restarts for BMM

vishwajith-s:

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.utils.URLEncodedUtils;
 import org.apache.kafka.clients.consumer.Consumer;
@@ -47,6 +48,10 @@ import com.linkedin.datastream.common.DatastreamUtils;
 import com.linkedin.datastream.common.DiagnosticsAware;
 import com.linkedin.datastream.common.JsonUtils;
 import com.linkedin.datastream.common.ThreadUtils;
+import com.linkedin.datastream.metrics.BrooklinMeterInfo;
+import com.linkedin.datastream.metrics.BrooklinMetricInfo;
+import com.linkedin.datastream.metrics.DynamicMetricsManager;
+import com.linkedin.datastream.metrics.MetricsAware;
 import com.linkedin.datastream.server.DatastreamTask;
 import com.linkedin.datastream.server.api.connector.Connector;
 import com.linkedin.datastream.server.api.connector.DatastreamValidationException;
@@ -72,6 +77,11 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
   private static final Duration POST_CANCEL_TASK_TIMEOUT = Duration.ofSeconds(5);
   private static final Duration SHUTDOWN_EXECUTOR_SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
   static final Duration MIN_DAEMON_THREAD_STARTUP_DELAY = Duration.ofMinutes(2);
+
+  private static final String NUM_TASK_RESTARTS = "numTaskRestarts";
+
+  private final DynamicMetricsManager _dynamicMetricsManager;
+  private final String _metricsPrefix;
 
   protected final String _connectorName;
   protected final KafkaBasedConnectorConfig _config;
@@ -112,12 +122,18 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
    * @see KafkaBasedConnectorConfig
    */
   public AbstractKafkaConnector(String connectorName, Properties config, GroupIdConstructor groupIdConstructor,
-      String clusterName, Logger logger) {
+      String clusterName, Logger logger, String metricsPrefix) {
     _connectorName = connectorName;
     _logger = logger;
     _clusterName = clusterName;
     _config = new KafkaBasedConnectorConfig(config);
     _groupIdConstructor = groupIdConstructor;
+    _dynamicMetricsManager = DynamicMetricsManager.getInstance();
+    _metricsPrefix = metricsPrefix;
+  }
+
+  protected static String generateMetricsPrefix(String connectorName, String simpleClassName) {
+    return StringUtils.isBlank(connectorName) ? simpleClassName : connectorName + "." + simpleClassName;
   }
 
   protected abstract AbstractKafkaBasedConnectorTask createKafkaBasedConnectorTask(DatastreamTask task);
@@ -229,6 +245,7 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
         _logger.warn("Creating a new connector task for the datastream task {}", datastreamTask);
         _runningTasks.put(datastreamTask, createKafkaConnectorTask(datastreamTask));
       });
+      _dynamicMetricsManager.createOrUpdateCounter(_metricsPrefix, NUM_TASK_RESTARTS, deadDatastreamTasks.size());
     }
   }
 
@@ -552,5 +569,13 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
     public AbstractKafkaBasedConnectorTask getConnectorTask() {
       return _task;
     }
+  }
+
+  @Override
+  public List<BrooklinMetricInfo> getMetricInfos() {
+    List<BrooklinMetricInfo> metrics = new ArrayList<>();
+    metrics.add(new BrooklinMeterInfo(buildMetricName(_metricsPrefix + MetricsAware.KEY_REGEX + NUM_TASK_RESTARTS)));
+
+    return Collections.unmodifiableList(metrics);
   }
 }
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
@@ -5,6 +5,7 @@
  */
 package com.linkedin.datastream.connectors.kafka;
 
+import com.linkedin.datastream.metrics.BrooklinGaugeInfo;
 import java.net.URI;
 import java.nio.charset.Charset;
 import java.time.Duration;
@@ -48,10 +49,8 @@ import com.linkedin.datastream.common.DatastreamUtils;
 import com.linkedin.datastream.common.DiagnosticsAware;
 import com.linkedin.datastream.common.JsonUtils;
 import com.linkedin.datastream.common.ThreadUtils;
-import com.linkedin.datastream.metrics.BrooklinMeterInfo;
 import com.linkedin.datastream.metrics.BrooklinMetricInfo;
 import com.linkedin.datastream.metrics.DynamicMetricsManager;
-import com.linkedin.datastream.metrics.MetricsAware;
 import com.linkedin.datastream.server.DatastreamTask;
 import com.linkedin.datastream.server.api.connector.Connector;
 import com.linkedin.datastream.server.api.connector.DatastreamValidationException;
@@ -79,10 +78,10 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
   static final Duration MIN_DAEMON_THREAD_STARTUP_DELAY = Duration.ofMinutes(2);
 
   private static final String NUM_TASK_RESTARTS = "numTaskRestarts";
-
-  private final DynamicMetricsManager _dynamicMetricsManager;
+  private long _numTaskRestarts = 0;
   private final String _metricsPrefix;
 
+  protected final DynamicMetricsManager _dynamicMetricsManager;
   protected final String _connectorName;
   protected final KafkaBasedConnectorConfig _config;
   protected final GroupIdConstructor _groupIdConstructor;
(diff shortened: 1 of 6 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -73,6 +77,11 @@
   private static final Duration SHUTDOWN_EXECUTOR_SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
   static final Duration MIN_DAEMON_THREAD_STARTUP_DELAY = Duration.ofMinutes(2);
 
+  private static final String MODULE = AbstractKafkaConnector.class.getSimpleName();
```
somandal, line 80: Since this is an abstract class which needs to be extended by actual Connectors, can we take a metrics prefix as input and use that to construct the metrics instead? This way, the metrics will be more meaningful, and it'll be easier to differentiate between different connectors that use this as their base.

You can look at `AbstractKafkaBasedConnectorTask` to see how connector tasks that extend it construct and pass their metrics prefix. We do want both the extending class name and the connector name to be part of the metrics prefix. This is because we have many connectors which use the same class.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -553,4 +564,12 @@ public AbstractKafkaBasedConnectorTask getConnectorTask() {
       return _task;
     }
   }
+
+  @Override
+  public List<BrooklinMetricInfo> getMetricInfos() {
```
somandal, line 569: Since this is an abstract class, you need to fix all the extenders of this class to call super.getMetricsInfos() in their override of getMetricsInfos(), along with their existing code to return all metrics, otherwise this metric won't be picked up by them. Any in-house Connectors need to be fixed too.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -64,7 +68,7 @@
  *  <li>Restarting stalled {@link DatastreamTask}s</li>
  * </ul>
  */
-public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAware {
+public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAware, MetricsAware {
```
somandal, line 71: `Connector` already implements `MetricsAware`, so you don't need to here

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -229,6 +239,7 @@ protected void restartDeadTasks() {
         _logger.warn("Creating a new connector task for the datastream task {}", datastreamTask);
         _runningTasks.put(datastreamTask, createKafkaConnectorTask(datastreamTask));
       });
+      _dynamicMetricsManager.createOrUpdateCounter(MODULE, NUM_TASK_RESTARTS, deadDatastreamTasks.size());
```
somandal, line 242: @ahmedahamid is there a recommendation on when to use a gauge vs. a counter?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -553,4 +564,12 @@ public AbstractKafkaBasedConnectorTask getConnectorTask() {
       return _task;
     }
   }
+
+  @Override
+  public List<BrooklinMetricInfo> getMetricInfos() {
+    List<BrooklinMetricInfo> metrics = new ArrayList<>();
+    metrics.add(new BrooklinMeterInfo(buildMetricName(NUM_TASK_RESTARTS)));
```
somandal, line 571: Once you move to using the metrics prefix, you'll need to modify this. Plus you should use a regex for matching:

`<metrics prefix> + MetricsAware.KEY_REGEX + <metrics name>`

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: Done.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: Done.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -229,6 +245,7 @@ protected void restartDeadTasks() {
         _logger.warn("Creating a new connector task for the datastream task {}", datastreamTask);
         _runningTasks.put(datastreamTask, createKafkaConnectorTask(datastreamTask));
       });
+      _dynamicMetricsManager.createOrUpdateCounter(_metricsPrefix, NUM_TASK_RESTARTS, deadDatastreamTasks.size());
```
vishwajith-s, line 248: Do we need to add regex here? Is this right?

ahmedahamid:
@vishwajith-s Please, add new commits to the branch instead of force pushing. GitHub has no way of showing us incremental changes if you force push.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
ahmedahamid, line 248: Sonam left me a question about counters vs. gauges but GitHub won't allow me to comment on it for some reason — they're roughly comparable but [this](https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Health+Check+Sensor+Basics#HealthCheckSensorBasics-CountervsGauge) ingraphs wiki page seems to suggest we should be using gauges for values that would increase and decrease over time (which seems to be the case for the metric this PR is adding). A similar [comment](https://github.com/dropwizard/metrics/issues/1043#issuecomment-272706512) has been made in response to the same question on the repo of the metrics library we're using.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -553,4 +570,12 @@ public AbstractKafkaBasedConnectorTask getConnectorTask() {
       return _task;
     }
   }
+
+  @Override
+  public List<BrooklinMetricInfo> getMetricInfos() {
+    List<BrooklinMetricInfo> metrics = new ArrayList<>();
+    metrics.add(new BrooklinMeterInfo(buildMetricName(_metricsPrefix + MetricsAware.KEY_REGEX + NUM_TASK_RESTARTS)));
```
ahmedahamid, line 577: This should be `BrooklinCounterInfo` if the metric is a counter.

somandal: @ahmedahamid instead of using buildMetricName() here, which prepends the class, can we instead do what other  similar metrics do? One example from `KafkaMirrorMakerConnectorTask`:

`    metrics.add(new BrooklinMeterInfo(generateMetricsPrefix(connectorName, CLASS_NAME) + MetricsAware.KEY_REGEX
        + TASK_LOCK_ACQUIRE_ERROR_RATE));`

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -553,4 +570,12 @@ public AbstractKafkaBasedConnectorTask getConnectorTask() {
       return _task;
     }
   }
+
+  @Override
+  public List<BrooklinMetricInfo> getMetricInfos() {
```
somandal, line 575: @ahmedahamid and @vishwajith-s I just noticed that callers of this pass the "connectorName" when fetching metrics from the connectorTask. Should we also only return metrics for a given "connectorName" here?

somandal: @vishwajith-s had a chat with @ahmedahamid about this offline. We not need to pass the "connectorName" here, as the only metric available is the one we added as part of this PR and it is keyed based on the "connectorName". For the tasks, it looks like the method needed to be static for some reason, or it was an oversight, which is why the "connectorName" is passed.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
somandal: Discussed with @ahmedahamid offline, and we don't need a MetricsAware.KEY_REGEX here, since we don't have any dynamic element in the metric name. For the connector task metrics on the other hand, we do have the datastream or topic name as a dynamic part.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -73,6 +77,11 @@
   private static final Duration SHUTDOWN_EXECUTOR_SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
   static final Duration MIN_DAEMON_THREAD_STARTUP_DELAY = Duration.ofMinutes(2);
 
+  private static final String MODULE = AbstractKafkaConnector.class.getSimpleName();
+  private static final String NUM_TASK_RESTARTS = "numTaskRestarts";
+
+  private final DynamicMetricsManager _dynamicMetricsManager;
```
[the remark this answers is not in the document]
vishwajith-s: Done.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
ahmedahamid: Here's what I found: https://github.com/linkedin/brooklin/pull/740#discussion_r469408157. GitHub started allowing me to leave responses on comments after I posted my own review.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: As discussed made this a Gauge.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: @somandal discussed this with Sonam. We do not need Regex here as we do not have any variable parts here. But I made this a Gauge.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: Thanks for the detailed explanation. I appreciate it.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
vishwajith-s: @somandal with the recent changes, this may not be required.

vishwajith-s: connector name is being passed as part of metrics prefix

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -118,6 +127,10 @@ public AbstractKafkaConnector(String connectorName, Properties config, GroupIdCo
     _clusterName = clusterName;
     _config = new KafkaBasedConnectorConfig(config);
     _groupIdConstructor = groupIdConstructor;
+    _dynamicMetricsManager = DynamicMetricsManager.getInstance();
+    _metricsPrefix = StringUtils.isBlank(connectorName) ? this.getClass().getSimpleName()
+        : connectorName + "." + this.getClass().getSimpleName();
+    _dynamicMetricsManager.registerGauge(_metricsPrefix, NUM_TASK_RESTARTS, () -> _numTaskRestarts);
```
ahmedahamid, line 133: `_numTaskRestarts` will be accessed by different threads (for reading and writing). I suggest marking it `volatile`.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestAbstractKafkaConnector.java
```diff
@@ -50,6 +62,16 @@ public void testConnectorRestartCalled() {
...
 
+    // Verify metric for nanny task restarts get incremented
+    Gauge<Long> metric = DynamicMetricsManager.getInstance()
+        .getMetric("test" + "." + TestKafkaConnector.class.getSimpleName() + "." + "numTaskRestarts");
+    Assert.assertNotNull(metric);
+    Assert.assertEquals(metric.getValue(), Long.valueOf(1));
+
+    // Verify that metrics created through DynamicMetricsManager match those returned by getMetricInfos() given the
+    // connector name of interest.
+    MetricsTestUtils.verifyMetrics(connector, DynamicMetricsManager.getInstance());
```
ahmedahamid, line 73: By default, this verification won't include metrics that do not start with the class name of the `metricsAware` object in question, i.e. it won't consider the metric `test.TestKafkaConnector.numTaskRestarts` because the it doesn't start with `TestKafkaConnector`. To have it include that metric, you'll need to pass in a predicate that explicitly filters such metrics, e.g. `s -> s.startsWith("test")`

PR #741, opened 2020-08-13
Eagerly register a couple of Coordinator KeyedMeter metrics for easier alerting

somandal:
Some important alerts for events such as handleLeaderDoAssignment_numRetries have been disabled since these metrics are lazily created, leading to noise for alerting purposes. Have such metrics be eagerly registered so that they always have a value and enable easier alerting.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -1867,9 +1867,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
       private final boolean _isEagerlyRegistered;
 
       KeyedMeter(String key, String name) {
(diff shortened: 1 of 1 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1856,10 +1864,18 @@ public String getName() {
 
       private final String _key;
       private final String _name;
+      private final boolean _isEagerlyRegistered;
 
       KeyedMeter(String key, String name) {
         _key = key;
         _name = name;
+        _isEagerlyRegistered = false;
```
ahmedahamid, line 1872: Delegate to `this(key, name, false)`

somandal: done

PR #745, opened 2020-08-18
Stop all the tasks on Session Expiry

vmaheshw:
On session expired, the coordinator should stop processing any messages from CoordinatorEventBlockingQueue and update the connector to stop all the running tasks within a debounce timer. Coordinator should ensure that all the states are cleaned up so that the task can be assigned to another instance.


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -415,6 +441,74 @@ public void onPartitionMovement(Long notifyTimestamp) {
...
+  /**
+   * {@inheritDoc}
+   * Stop all the tasks and wait for new session to connect.
+   */
+  @Override
+  public void onSessionExpired() {
+    if (_shutdown) {
+      return;
+    }
+    stopEventThread();
```
somandal, line 454: In the past we've seen some issues with handleLeaderDoAssignment() getting interrupted during cleanup and causing bad state such that we couldn't do future leader assignments anymore for BMM. This would require a full cluster restart to fix the state. We don't know the root cause yet, so we don't know how to  fix this scenario. Won't only calling stopEventThread() make such occurrences more frequent? If for some reason we cannot give more time to the eventThread to stop naturally, then how do we avoid that problem?

Have we added the relevant debug logs we needed for that issue (I recall it being an action item in one of the ZK session expiry documents that listed out ZK related issues)? If not, feel free to assign that item to me so that if it happens again I'll have something to debug with. Don't want to make that problem more frequent without appropriate logs in place. I've wasted 2-3 debugging sessions on it in the past to find out that it was related to early killing of handleLeaderDoAssignment, and after the timeout was increased, we stopped seeing it (so far). Which is why I haven't had a chance to root cause it since. Unit test don't help either.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -415,6 +441,74 @@ public void onPartitionMovement(Long notifyTimestamp) {
...
+        .collect(Collectors.toList());
+
+    // Wait till all the futures are complete or timeout.
+    ExecutorService threadPoolExecutor = Executors.newFixedThreadPool(1);
+    threadPoolExecutor.submit(() -> {
+      Instant start = Instant.now();
+      try {
+        getAssignmentsFuture(assignmentChangeFutures, start);
+      } catch (Exception e) {
+        // if it's timeout then we will retry
```
somandal, line 487: Where is the retry logic? I don't see any retries happening.
In general, what should the protocol be for TimeoutException? And other exceptions like interrupt? Can this get interrupted?

Basically, we need the tasks to stop, what if they don't? E.g. assignment taking too long so we cancel the futures, or our thread gets interrupted. The debounce timer assumption for the persistent lock changes is that the tasks will be brought down in 30 seconds.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -622,8 +714,8 @@ private boolean getCustomCheckpointing(DatastreamTask task) {
     boolean customCheckpointing = _connectors.get(task.getConnectorType()).isCustomCheckpointing();
 
     Datastream datastream = task.getDatastreams().get(0);
-    if (datastream.hasMetadata()
-        && datastream.getMetadata().containsKey(DatastreamMetadataConstants.CUSTOM_CHECKPOINT)) {
+    if (datastream.hasMetadata() && datastream.getMetadata() != null &&
```
somandal, line 717: Can we use Objects.requireNonNull() in place of `datastream.getMetadata() != null`?

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -2597,6 +2595,36 @@ public void testCoordinatorLeaderCleanupTasksPostElection() throws Exception {
...
+            DummyTransportProviderAdminFactory.PROVIDER_NAME, new Properties()));
+
+    TestHookConnector connector1 = new TestHookConnector("connector1", testConnectorType);
+    instance1.addConnector(testConnectorType, connector1, new BroadcastStrategy(Optional.empty()), false,
+        new SourceBasedDeduper(), null);
+    instance1.start();
+
+    ZkClient zkClient = new ZkClient(_zkConnectionString);
+    DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, testConnectorType, datastreamName);
+    //verify the assignment
```
somandal, line 2616: nit: space after '//'

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -415,6 +441,74 @@ public void onPartitionMovement(Long notifyTimestamp) {
...
+   * Stop all the tasks and wait for new session to connect.
+   */
+  @Override
+  public void onSessionExpired() {
+    if (_shutdown) {
+      return;
+    }
+    stopEventThread();
+
+    _leaderDatastreamAddOrDeleteEventScheduled.set(false);
```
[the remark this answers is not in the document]
vmaheshw: right.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -1880,9 +1933,11 @@ public void testSimpleAssignmentStrategyIndependent() throws Exception {
     instance1.stop();
     instance2.stop();
     zkClient.close();
+    instance1.getDatastreamCache().getZkclient().close();
+    instance2.getDatastreamCache().getZkclient().close();
   }
 
-  @Test
+  @Test //(enabled = false)
```
somandal, line 1940: Remove "//(enabled = false)"

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -415,6 +444,90 @@ public void onPartitionMovement(Long notifyTimestamp) {
...
+        .filter(Objects::nonNull)
+        .collect(Collectors.toList());
+
+    onDatastreamChange(new ArrayList<>());
+    // Shutdown the event producer to stop any further production of records.
+    // Event producer shutdown sequence does not need to wait for onAssignmentChange to complete.
+    // This will ensure that even if any task thread does not respond to thread interruption, it will
+    // still not be able to produce any records to destination.
+    for (DatastreamTask task : _assignedDatastreamTasks.values()) {
+      ((EventProducer) task.getEventProducer()).shutdown(true);
```
somandal, line 492: Just leaving this comment here for tracking. As discussed offline, this is not sufficient as this only shutdown the producer for currently assigned tasks. Previously assigned tasks may have been unassigned from a producer and that producer will never get closed. The fix is involved so will be done as a separate PR. From our discussion it sounded like we will need to modify the interface for the TransportProvider, feel free to assign these to me, I'd be happy to help.

Problematic race condition:

> We create a pool of producers which is shared by all tasks for a given connecter -> Kafka broker mapping. There may have been a sequence of previously assigned tasks that are assigned producers that aren't assigned to any of the current tasks. Shouldn't we close all producers to be safe?
> For example, say we create 10 producers per connector -> Kafka broker mapping.
> initialize task t1 -> producer p1
> initialize task t2 -> producer p2
> initialize task t3 -> producer p3
> assignment at this point in time: t1, t2, t3, unused producers from p4-10
> uninitialize task t2 -> which will unassign t2 from producer p2's task list
> assignment at this point in time: t1, t3, unused producers from p4-10, and p2 should also be unused, but wasn't closed
> ZK session expiry occurs
> We'll call task.getEventProducer()).shutdown(true) for t1 and t3, which will close producer p1 and p3. Producer p2 may still be running (send is taking a long time) in this scenario, since t2 was only removed from the task list of p2 and the producer was not actually shutdown.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1793,6 +1917,7 @@ private void registerGaugeMetrics() {
           .put(MAX_PARTITION_COUNT_IN_TASK, MAX_PARTITION_COUNT::get)
           .put(NUM_PAUSED_DATASTREAMS_GROUPS, PAUSED_DATASTREAMS_GROUPS::get)
           .put(IS_LEADER, () -> _coordinator.getIsLeader().getAsBoolean() ? 1 : 0)
+          .put(ZK_SESSION_EXPIRED, () -> _coordinator.isZkSessionExpired() ? 1 : 0)
```
somandal, line 1920: Thanks for adding this!

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -422,8 +422,10 @@ private void checkpoint(int partition, String checkpoint) {
   /**
    * Shuts down the event producer by flushing the checkpoints and closing the transport provider
    */
-  public void shutdown() {
-    _checkpointProvider.flush();
+  public void shutdown(boolean skipCheckpoint) {
+    if (!skipCheckpoint) {
+      _checkpointProvider.flush();
+    }
```
somandal, line 428: You may want to add a comment on why we may want to skip checkpoint commit. (though this only affects if using the framework level checkpointing, which we don't use internally)

Feel free to update the comment where you pass 'skipCheckpoint' as 'true' instead of here if it feel like a better place to discuss this aspect.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1765,6 +1770,10 @@ void onSessionExpired() {
     // cancel the lock clean up
     _orphanLockCleanupFuture.cancel(true);
     onBecomeFollower();
+    if (_listener != null) {
+      _listener.onSessionExpired();
+    }
+    // currently it will try to disconnect and fail. TODO: fix the connect and listen to handleNewSession.
```
somandal, line 1776: nit: do you still need this comment? Won't the exiting comment be sufficient?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -317,7 +316,6 @@ void close(DatastreamTask task) {
         shutdownProducer();
       }
     }
-    ThreadUtils.shutdownExecutor(_producerCloseExecutorService, PRODUCER_CLOSE_EXECUTOR_SHUTDOWN_TIMEOUT, _log);
```
somandal, line 320: Can you open a bug for fixing this and assign it to me? I've sent you an email about it too, so wondering if that approach looks sound? If so, I can take this up after we finish basic low-level certification.

somandal (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
vmaheshw: I have a TODO marker which makes it explicit.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: done.

somandal (approved):

ahmedahamid (approved):

PR #748, opened 2020-08-27
Add a Future implementation that wraps a Thread

ahmedahamid:


datastream-utils/src/main/java/com/linkedin/datastream/common/ThreadFuture.java
```diff
@@ -0,0 +1,105 @@
+/**
+ *  Copyright 2019 LinkedIn Corporation. All rights reserved.
```
vmaheshw, line 2: nit: 2020 instead of 2019

datastream-utils/src/main/java/com/linkedin/datastream/common/ThreadFuture.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.lang3.Validate;
+
+
+/**
+ * A {@link Future} impl that wraps a {@link Thread}.
```
vmaheshw, line 16: Can you please document the benefit of converting thread to future instead of directly checking the thread status?

somandal: +1

datastream-utils/src/main/java/com/linkedin/datastream/common/ThreadFuture.java
somandal: same nit for the test file

datastream-utils/src/main/java/com/linkedin/datastream/common/ThreadFuture.java
ahmedahamid: There's no inherent benefit. We're just adapting one API to another, i.e. we have a utility (`FutureUtils`) that checks if a `Future` is done after a `timeout` and we're using that to verify if connector tasks stop within a specific duration. We need that same logic for situations when we working with threads not futures.

My initial intent was to keep `ThreadFuture` private within the MySQL connector module but you suggested it may be useful as a general purpose utility.

PR #749, opened 2020-09-09
Fix stickyMulticast strategy to avoid assigning the same task to multiple instances

vmaheshw:
Currently, whenever the leader is updating the assignments, it first adds the assignments to the instances and then remove the assignments to ensure that we don't lose the task nodes in case of server crash/thread interruption.

The behavior we are seeing that if the new assignment update has some task moving from one live instance-1 to another live instance-2 and after adding the task in the live-instance-2 assignment if there is a thread interruption and does not let the leader to remove the task from live instance-1 assignment, the next leader does not check if there is any duplicate in the task between two live instances which results in exception in leaderDoAssignment.

Solving this problem by removing the duplicate assignments from new assignment.


datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
```diff
@@ -282,6 +282,23 @@ public void testDontCreateNewTasksWhenInstanceChanged() {
     Assert.assertTrue(oldAssignmentTasks.containsAll(newAssignmentTasks));
   }
 
+  @Test
+  public void testSameTaskIsNotAssignedToAnyTwoInstances() {
```
ahmedahamid, line 286: nit: `ToMoreThanOneInstance`

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
```diff
@@ -282,6 +282,23 @@ public void testDontCreateNewTasksWhenInstanceChanged() {
...
+    String[] instances = new String[]{"instance1", "instance2", "instance3"};
+    int numDatastreams = 5;
+    List<DatastreamGroup> datastreams = generateDatastreams("ds", numDatastreams);
+    StickyMulticastStrategy strategy = new StickyMulticastStrategy(Optional.empty(), Optional.empty());
+    Map<String, Set<DatastreamTask>> assignment =
+        strategy.assign(datastreams, Arrays.asList(instances), new HashMap<>());
+    assignment.get("instance1").addAll(assignment.get("instance2"));
```
ahmedahamid, line 293: Please add a comment explaining why the test is doing this step.

ahmedahamid (dismissed):

somandal (approved):

somandal (approved):

ahmedahamid (approved):

PR #750, opened 2020-09-10
Fix waitForTaskRelease and orphan lock cleanup

vmaheshw:
Fixing two issues in this PR.

1. During waitForTaskRelease, the recently changed code added hierarchy in the lock path, but the changes in the child node was no fixed to lockPrefix from lockroot. This results in the waitForTaskRelease for the timeout - debounce timer even if the lock was already released by other owner. 

2. During orphan lock cleanup, we should ignore the connectors that do not have locks.


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1196,7 +1200,7 @@ private void waitForTaskRelease(DatastreamTask task, long timeoutMs, String lock
     CountDownLatch busyLatch = new CountDownLatch(1);
 
     String lockNode = lockPath.substring(lockPath.lastIndexOf('/') + 1);
-    String lockRootPath = KeyBuilder.datastreamTaskLockRoot(_cluster, task.getConnectorType());
+    String lockPrefixPath = KeyBuilder.datastreamTaskLockPrefix(_cluster, task.getConnectorType(), task.getTaskPrefix());
```
somandal, line 1203: 🤦‍♀️ sorry for missing this in the review.

somandal (approved):

ahmedahamid (approved):

PR #751, opened 2020-09-14
shutdown producer in KafkaProducerWrapper during task assign/unassign call.

vmaheshw:
The producer should be not have any pending inflight sends post the task is unassigned. KafkaProducerWrapper should not allow any more send operation for the task after it is unassigned. In the current code, the tasks share the producer and if the producer is initialized, it does not check whether the task is allowed to send or not. Adding a check to stop that. Since multiple tasks share the same KafkaProducerWrapper, when one task is unassigned, we should shutdown the producer to ensure that there are no pending inflight sends. Kafka allows a longer time for pending inflight sends. So, the sends are not aborted or forced to complete, the destination can see  wrong ordering of the messages. Flushing the kafka producer will not achieve this. The only way is to shutdown the producer which internally force close the kafka producer after a specified timeout.

In our current code, the producer is closed only for the existing tasks. So, if there are tasks which were just unassigned, the corresponding producer will never be forced shutdown. This is fine so far because the shutdown was called only during the coordinator stop. But, with recent change to handle session expiry, all the producers are expected to be shutdown before reconnecting the new session. So, it is important to shutdown the producers of the tasks which were recently unassigned as well.  Adding the shutdown logic while `unassignTask` will take care of this.

A new producer should not be created till the older producer is marked as closed. This is to avoid out-of-order record generation. Also, made the synchronization blocks light, so that the no thread holds the lock for a long time. 

Moved the kafka producer flush outside the synchronized block, because it is okay for two threads sharing kafka producer to call `flush()` at the same time.

During send call, we will check if the task is part of the current assigned task list, which will internally calculate the hash every time. To optimize and avoid repetition of this operation, we will cache the hash in `DatastreamTaskImpl`.

Making an optimization to call all the tasks getting unassigned in one shot to make sure that the producer is closed only once. Adding a new interface in TransportProviderAdmin to handle it. This will give TransportProviderAdmins more power to optimize logic if there is any sharing involved.

Also, reducing the shutdown timer in `AbstractKafkaConnector` to be in alignment with 30 second debounce timer.

### DIFF
--- a/datastream-directory/src/main/java/com/linkedin/datastream/server/DirectoryTransportProviderAdmin.java
+++ b/datastream-directory/src/main/java/com/linkedin/datastream/server/DirectoryTransportProviderAdmin.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.time.Duration;
+import java.util.List;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
@@ -37,6 +38,10 @@ public class DirectoryTransportProviderAdmin implements TransportProviderAdmin {
 
   @Override
   public void unassignTransportProvider(DatastreamTask task) {
+  }
+
+  @Override
+  public void unassignTransportProvider(List<DatastreamTask> taskList) {
   }
 
   @Override
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -172,12 +172,12 @@ class KafkaProducerWrapper<K, V> {
   }
 
   private Optional<Producer<K, V>> maybeGetKafkaProducer(DatastreamTask task) {
-    Producer<K, V> producer = _kafkaProducer;
     if (!_tasks.contains(task)) {
       _log.warn("Task {} has been unassigned for producer, abort the send", task);
       return Optional.empty();
     }
 
+    Producer<K, V> producer = _kafkaProducer;
     if (producer == null) {
       try {
         producer = initializeProducer(task);
@@ -268,7 +268,7 @@ class KafkaProducerWrapper<K, V> {
             }
           });
         } else {
-          throw new DatastreamRuntimeException("kafka producer not available for the task");
+          throw new DatastreamRuntimeException(String.format("kafka producer not available for the task: %s", task));
         }
 
         retry = false;
@@ -369,23 +369,23 @@ class KafkaProducerWrapper<K, V> {
       _producerLock.unlock();
     }
 
-    try {
-      if (producer != null) {
-          producer.flush(_producerFlushTimeoutMs, TimeUnit.MILLISECONDS);
-      }
-    } catch (InterruptException | TimeoutException e) {
-      // The KafkaProducer object should not be reused on an interrupted flush
+    if (producer != null) {
       try {
-        _producerLock.lock();
-        if (producer == _kafkaProducer) {
-          _log.warn("Kafka producer flush interrupted/timed out, closing producer {}.", producer);
-          shutdownProducer();
-        } else {
-          _log.warn("Kafka producer flush interrupted/timed out, producer {} already closed.", producer);
+        producer.flush(_producerFlushTimeoutMs, TimeUnit.MILLISECONDS);
+      } catch (InterruptException | TimeoutException e) {
+        // The KafkaProducer object should not be reused on an interrupted flush
+        try {
+          _producerLock.lock();
+          if (producer == _kafkaProducer) {
+            _log.warn("Kafka producer flush interrupted/timed out, closing producer {}.", producer);
+            shutdownProducer();
+          } else {
+            _log.warn("Kafka producer flush interrupted/timed out, producer {} already closed.", producer);
+          }
+          throw e;
+        } finally {
+          _producerLock.unlock();
         }
-        throw e;
-      } finally {
-        _producerLock.unlock();
       }
     }
   }
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -315,9 +315,9 @@ class KafkaProducerWrapper<K, V> {
         if (_kafkaProducer != null) {
           _kafkaProducer.flush(_producerFlushTimeoutMs, TimeUnit.MILLISECONDS);
         }
-      } catch (InterruptException e) {
+      } catch (InterruptException | TimeoutException e) {
         // The KafkaProducer object should not be reused on an interrupted flush
-        _log.warn("Kafka producer flush interrupted, closing producer {}.", _kafkaProducer);
+        _log.warn("Kafka producer flush interrupted/timed out, closing producer {}.", _kafkaProducer);
         shutdownProducer();
         throw e;
       }
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -115,12 +115,14 @@ class KafkaProducerWrapper<K, V> {
 
   // A lock used to synchronize access to operations performed on the _kafkaProducer object
   private final Lock _producerLock = new ReentrantLock();
+  // A condition to wait on before creating a new producer
   private final Condition _waitOnNoProducerClose = _producerLock.newCondition();
 
   // An executor to spawn threads to close the producer.
   private final ExecutorService _producerCloseExecutorService = Executors.newSingleThreadExecutor(
       new ThreadFactoryBuilder().setNameFormat("KafkaProducerWrapperClose-%d").build());
-  private boolean closeInProgress = false;
+
+  private boolean _closeInProgress = false;
 
   KafkaProducerWrapper(String logSuffix, Properties props) {
     this(logSuffix, props, null);
@@ -180,7 +182,7 @@ class KafkaProducerWrapper<K, V> {
       try {
         producer = initializeProducer(task);
       } catch (InterruptedException e) {
-        e.printStackTrace();
+        _log.warn("Got interrupted while trying to initialize the producer for task {}", task);
       }
     }
     return Optional.ofNullable(producer);
@@ -201,7 +203,7 @@ class KafkaProducerWrapper<K, V> {
 
       // whenever a task is unassigned the kafka producer should be shutdown to ensure that
       // there are no in-flight sends. Any further send will not work.
-      if (taskPresent && _kafkaProducer != null && !closeInProgress) {
+      if (taskPresent && _kafkaProducer != null && !_closeInProgress) {
         shutdownProducer();
       }
     } finally {
@@ -224,7 +226,8 @@ class KafkaProducerWrapper<K, V> {
     try {
       _producerLock.lock();
 
-      while (closeInProgress) {
+      // make sure there is no close in progress.
+      while (_closeInProgress) {
         boolean closeCompleted = _waitOnNoProducerClose.await(CLOSE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
         if (!closeCompleted) {
           _log.debug("Close not completed. Retry");
@@ -308,11 +311,12 @@ class KafkaProducerWrapper<K, V> {
     Producer<K, V> producer;
     try {
       _producerLock.lock();
-      if (_kafkaProducer == null || closeInProgress) {
+      // if there is no producer or the producer close is in progress, return.
+      if (_kafkaProducer == null || _closeInProgress) {
         return;
       }
       producer = _kafkaProducer;
-      closeInProgress = true;
+      _closeInProgress = true;
       // Nullify first to prevent subsequent send() to use
       // the current producer which is being shutdown.
       _kafkaProducer = null;
@@ -320,7 +324,6 @@ class KafkaProducerWrapper<K, V> {
       // This may be called from the send callback. The callbacks are called from the sender thread, and must complete
       // quickly to avoid delaying/blocking the sender thread. Thus schedule the actual producer.close() on a separate
       // thread
-
       _producerCloseExecutorService.submit(() -> {
         _log.info("KafkaProducerWrapper: Closing the Kafka Producer");
         try {
@@ -330,7 +333,7 @@ class KafkaProducerWrapper<K, V> {
         } finally {
           try {
             _producerLock.lock();
-            closeInProgress = false;
+            _closeInProgress = false;
             _waitOnNoProducerClose.signalAll();
           } finally {
             _producerLock.unlock();
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
@@ -5,21 +5,6 @@
  */
 package com.linkedin.datastream.kafka;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.RateLimiter;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import com.linkedin.datastream.common.DatastreamRuntimeException;
-import com.linkedin.datastream.common.DatastreamTransientException;
-import com.linkedin.datastream.common.ReflectionUtils;
-import com.linkedin.datastream.common.VerifiableProperties;
-import com.linkedin.datastream.kafka.factory.KafkaProducerFactory;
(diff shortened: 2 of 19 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -177,7 +182,12 @@ void assignTask(DatastreamTask task) {
   }
 
   void unassignTask(DatastreamTask task) {
-    _tasks.remove(task);
+    boolean taskPresent = _tasks.remove(task);
+    // whenever a task is unassigned the kafka producer should be shutdown to ensure that
+    // there are no in-flight sends. Any further send will not work.
```
somandal, line 187: nit: "Any further sends by this task will not work"

Also, do you think we should add a small sentence about how this affects existing tasks that share the same producer? i.e. their inflight records will attempt to flushed, but if it doesn't complete, they'll see send failures, but this is okay since they'll be able to create a new producer on the next send?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -166,6 +166,11 @@ private void populateDefaultProducerConfigs() {
 
   private Optional<Producer<K, V>> maybeGetKafkaProducer(DatastreamTask task) {
     Producer<K, V> producer = _kafkaProducer;
+    if (!_tasks.contains(task)) {
```
somandal, line 169: Does adding a comment here make sense, that if the task is not assigned, production should be disallowed?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -177,7 +182,12 @@ void assignTask(DatastreamTask task) {
   }
 
   void unassignTask(DatastreamTask task) {
-    _tasks.remove(task);
+    boolean taskPresent = _tasks.remove(task);
+    // whenever a task is unassigned the kafka producer should be shutdown to ensure that
+    // there are no in-flight sends. Any further send will not work.
+    if (taskPresent) {
+      shutdownProducer();
```
somandal, line 189: One thing to note here is that since shutdownProducer() requires obtaining the lock, this operation can get stuck waiting on this lock. Operations such as flush can take a while (up to 15 minutes for BMM with the current flush timeout). Once concern might be that unassignTask() is called from the onAssignmentChange() threads, right? And have a limited time of 30-60 seconds within which it needs to complete?

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
```diff
@@ -89,6 +89,7 @@
   private DatastreamEventProducer _eventProducer;
   private String _transportProviderName;
   private SerDeSet _destinationSerDes = new SerDeSet(null, null, null);
+  private int _hashCode;
```
somandal, line 92: Can you add  a comment here or somewhere in this file that if a new field is added along with a setter, or any function that modifies the values, the setter/function must call `computeHashCode()`?

somandal (commented):
Have one concern about unassignTask() blocking, let's discuss this?
Rest lgtm!

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -710,10 +710,17 @@ private DatastreamTask getDatastreamTask(String taskName) {
     });
   }
 
-  private void uninitializeTask(DatastreamTask t) {
-    TransportProviderAdmin tpAdmin = _transportProviderAdmins.get(t.getTransportProviderName());
-    tpAdmin.unassignTransportProvider(t);
-    _cpProvider.unassignDatastreamTask(t);
+  private void uninitializeTask(List<DatastreamTask> t) {
```
somandal, line 713: not you fault, but can you change the input variable to be "tasks" instead of "t". It's much easier to read.

somandal (approved):

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: I feel the code is self explainatory.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -254,67 +298,92 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
+      _producerLock.unlock();
+    }
+
+    try {
+      if (producer != null) {
+          producer.flush(_producerFlushTimeoutMs, TimeUnit.MILLISECONDS);
+      }
+    } catch (InterruptException | TimeoutException e) {
+      // The KafkaProducer object should not be reused on an interrupted flush
+      if (producer == _kafkaProducer) {
```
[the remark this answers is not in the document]
vmaheshw: The reason to skip here is to avoid closing the other producer. I initially thought of passing producer to the method, but that will imply making the producer final to be passed from send callback and I wanted to avoid it from garbage collection point of view and producer is a heavyweight object in general.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: handled it.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -254,67 +298,92 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
+          _log.info("KafkaProducerWrapper: Kafka Producer is closed");
+        } finally {
+          try {
+            _producerLock.lock();
+            closeInProgress = false;
+            _waitOnNoProducerClose.signalAll();
+          } finally {
+            _producerLock.unlock();
+          }
+        }
```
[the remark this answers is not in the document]
vmaheshw: I moved the finally block in another method to improve readability.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
```diff
@@ -172,6 +173,24 @@ public void unassignTransportProvider(DatastreamTask task) {
...
+      if (_transportProviders.containsKey(task)) {
+        KafkaTransportProvider transportProvider = _transportProviders.remove(task);
+        producers.addAll(transportProvider.getProducers());
+        //transportProvider.getProducers().forEach(p -> p.unassignTask(task));
+      } else {
+        LOG.warn("Trying to unassign already unassigned transport provider for task {}.", task);
+      }
+    }
+
+    producers.forEach(p -> p.unassignTask(taskList));
```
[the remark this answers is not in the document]
vmaheshw: Yes, that is true.

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaProducerWrapper.java
```diff
@@ -94,37 +95,104 @@ private void testFlushBehaviorOnException(Class<? extends Throwable> exceptionCl
...
+    producerWrapper.verifyClose(1, 1);
+
+    // Closing the producer's task. Since this is the only task, the producer should be closed
+    producerWrapper.close(task);
+    producerWrapper.verifySend(1);
+    producerWrapper.verifyFlush(0);
+    producerWrapper.verifyClose(1, 1);
+    Assert.assertEquals(producerWrapper.getNumCreateKafkaProducerCalls(), 1);
+
+    producerWrapper.assignTask(task);
```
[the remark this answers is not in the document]
vmaheshw: added.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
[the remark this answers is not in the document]
somandal: I wasn't implying always passing the `producer` object to `shutdownProducer()`. I was thinking that you can keep `shutdownProducer()` as is, but add:

`
void shutdownProducer(KafkaProducer producer) {
    _producerLock.lock()
    try {
        if (producer == _kafkaProducer) {
             shutdownProducer()
       }
    } finally {
        _producerLock.unlock()
    }
}
`

This new method would be called from flush() only. The send callback path will just call the existing `shutdownProducer()` directly, since this method acquires the lock anyways.

I'd like to avoid access to `_kafkaProducer` outside the lock as much as possible. Makes the code much cleaner and easier to reason about. Otherwise one will always need to remember that another code path can potentially have changed the  `_kafkaProducer` object from underneath us by the time we actually call `shutdownProducer()`, implying that we've just closed a newly created producer, etc.

somandal (dismissed):
lgtm!

datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
DEEPTHIKORAT, line 92: Add a comment explaining the hashcode field as it is no obvious.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -166,8 +173,17 @@ private void populateDefaultProducerConfigs() {
 
   private Optional<Producer<K, V>> maybeGetKafkaProducer(DatastreamTask task) {
     Producer<K, V> producer = _kafkaProducer;
```
DEEPTHIKORAT, line 175: extreme nit: move this after the below check.

datastream-directory/src/main/java/com/linkedin/datastream/server/DirectoryTransportProviderAdmin.java
```diff
@@ -39,6 +40,10 @@ public TransportProvider assignTransportProvider(DatastreamTask task) {
   public void unassignTransportProvider(DatastreamTask task) {
   }
 
+  @Override
+  public void unassignTransportProvider(List<DatastreamTask> taskList) {
```
DEEPTHIKORAT, line 44: nit: Would a Set rather than a List be better to use here?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -166,8 +173,17 @@ private void populateDefaultProducerConfigs() {
...
+    if (!_tasks.contains(task)) {
```
DEEPTHIKORAT, line 176: Does access to _tasks not need to be synchronized?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -219,13 +258,18 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
       try {
         ++numberOfAttempt;
-        maybeGetKafkaProducer(task).ifPresent(p -> p.send(producerRecord, (metadata, exception) -> {
-          if (exception == null) {
-            onComplete.onCompletion(metadata, null);
-          } else {
-            onComplete.onCompletion(metadata, generateSendFailure(exception));
-          }
-        }));
+        Optional<Producer<K, V>> producer = maybeGetKafkaProducer(task);
```
DEEPTHIKORAT, line 261: nit: Make this line 
```Producer<K, V> producer = maybeGetKafkaProducer(task).get()``` instead  so that the check below also reads better as producer == null.

File: datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java [hunk 14]
```diff
@@ -254,67 +298,103 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
-    synchronized (_producerLock) {
+    try {
+      _producerLock.lock();
+      // if there is no producer or the producer close is in progress, return.
+      if (_kafkaProducer == null || _closeInProgress) {
+        return;
+      }
       producer = _kafkaProducer;
+      _closeInProgress = true;
       // Nullify first to prevent subsequent send() to use
```
DEEPTHIKORAT, line 317: is this comment still relevant now that producer access is protected with lock? The code is not harmful, but may be the comment can go if this is not relevant as it might cause some confusion on the intent of the code.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -254,67 +298,103 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
-    synchronized (_producerLock) {
+    Producer<K, V> producer;
+    try {
+      _producerLock.lock();
+      producer = _kafkaProducer;
+    } finally {
+      _producerLock.unlock();
+    }
+
+    try {
```
DEEPTHIKORAT, line 372: sorry I might be missing the reason the code looks like this but should it read
```
    if (producer != null) {
       try {
          producer.flush (...)
       } catch (..) {
         ...
       }
  } ```

File: datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java [hunk 16]
```diff
@@ -254,67 +298,103 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
...
+          producer.flush(_producerFlushTimeoutMs, TimeUnit.MILLISECONDS);
+      }
+    } catch (InterruptException | TimeoutException e) {
+      // The KafkaProducer object should not be reused on an interrupted flush
       try {
-        if (_kafkaProducer != null) {
-          _kafkaProducer.flush(_producerFlushTimeoutMs, TimeUnit.MILLISECONDS);
+        _producerLock.lock();
+        if (producer == _kafkaProducer) {
+          _log.warn("Kafka producer flush interrupted/timed out, closing producer {}.", producer);
```
DEEPTHIKORAT, line 381: Is the interruption/timeout typical? Wondering about the use of log.warn which are not too noticeable in logs.. wondering if log.error might be better, but I know it raises false alarms in thinking something is wrong when it is being handled here fine.

DEEPTHIKORAT (approved):
lgtm, just small nits and a couple of questions

datastream-directory/src/main/java/com/linkedin/datastream/server/DirectoryTransportProviderAdmin.java
vmaheshw: Keeping aligned with the other interfaces used in connector that pass bunch of task object, uses list and not set.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: _tasks is a concurrent Hash map and based on its usage, it really does not have to be in synchronized block. We have added other checks as well to catch the misuse. for eg, every send checks if the task is present in _tasks or not.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: I want to avoid calling `get()` without `isPresent()` check. Java 11 has introduced a new api `ifPresentOrElse()` that will make this api cleaner.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: Actually I feel the comment helps in understanding the intent, even though we have made sure that the producer close in progress will block the send anyway. In the send() if the producer is not null, it does not check for close in progress.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: This will be useful in identifying the producer stuck on flush issues. Yes, you are correct. The reason to not add it as ERROR  is to avoid false alarm. This is a self-recovering scenario which should not need extra attention. So, warns will be sufficient to  identify if this issue happened.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: done.

somandal (approved):

DEEPTHIKORAT (approved):

PR #758, opened 2020-10-01
Added committed offsets in KafkaTopicPartitionTracker and diag endpoint response

jzakaryan:
the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
@@ -16,6 +16,7 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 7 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
```diff
@@ -32,8 +33,8 @@
   private final String _consumerGroupId;
 
   private final Map<String, Set<Integer>> _topicPartitions = new ConcurrentHashMap<>();
-
-  private final Map<String, Map<Integer, Long>> _consumerOffsets = new ConcurrentHashMap<>();
+  private final Map<String, Map<Integer, Long>> _consumedOffsets = new ConcurrentHashMap<>();
```
jzakaryan, line 36: I was considering adding a separate `ConsumerOffsets` class to hold both consumed offset and committed offset and use that in the nested dictionary. But synchronizing the writes to the nested dictionaries was a not easy. `onPartitionsPolled` and `onOffsetsCommitted` calls would hold locks on the dictionaries and it would make this code bug prone and non-efficient. So I thought it would be better for the client-side to merge these dictionaries (if necessary for presentation purposes).

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
```diff
@@ -75,22 +76,30 @@ public void onPartitionsRevoked(@NotNull Collection<TopicPartition> topicPartiti
...
     // here is that another host may handle these partitions due to rebalance, and we don't want to have duplicate
     // consumer offsets for affected partitions (even though the ones with larger offsets wins).
+    removeOffsetsForTopicPartition(topicPartitions, _consumedOffsets);
+
+    // Remove committed offsets for partitions that have been revoked.
+    removeOffsetsForTopicPartition(topicPartitions, _committedOffsets);
+  }
+
+  private void removeOffsetsForTopicPartition(@NotNull Collection<TopicPartition> topicPartitions,
+      Map<String, Map<Integer, Long>> committedOffsets) {
```
vishwajith-s, line 89: Nitpick. May be you need to call this just offsets since its being used by both _consumedOffsets and _committedOffsets

vishwajith-s (approved):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaTopicPartitionTracker.java
somandal: +1 make this generic. We also intend to add end offsets to this list.

somandal (approved):
Was just wondering if it makes sense to add a test where flushless is not enabled, so that we can validate that "committedOffsets" is empty but "consumedOffsets" isn't, and it doesn't break anything?

PR #762, opened 2020-10-08
Fix TopicManager related metrics registration in getMetricsInfo in KafkaMirrorMakerConnectorTask

vmaheshw:
The regex used to register counter and gauge metrics are overlapping which results in the metrics matching both the regex. the metrics should match exactly one regex, otherwise it results in periodic errors in the logs. Since getMetricsInfo is called in the beginning and TopicManager is instantiated per task and hooked in the code via factory, there is no direct way to add and override static getMetricsInfo method. So, for every metric type, we will add a unique prefix which will be used by the Implementers of TopicManager interface to register and unregister metrics


somandal (approved):

jzakaryan (commented):
LGTM

jzakaryan (approved):

PR #763, opened 2020-10-08
Add metrics to track event processing time and time spent between polls

somandal:
the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -526,7 +526,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
     }
 
     if (_enableAdditionalMetrics) {
-      _consumerMetrics.updateEventProcessingTimeMs(processingTimeMillis);
+      _consumerMetrics.updatePerEventProcessingTimeMs(records.count() == 0 ? 0 : processingTimeMillis / records.count());
     }
   }
 
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -103,7 +103,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected final boolean _pausePartitionOnError;
   protected final Duration _pauseErrorPartitionDuration;
   protected final long _processingDelayLogThresholdMillis;
-  protected final boolean _enableAdditionalHistogramMetrics;
+  protected final boolean _enableAdditionalMetrics;
   protected final Optional<Map<Integer, Long>> _startOffsets;
 
   protected volatile String _taskName;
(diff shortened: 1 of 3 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -102,7 +103,7 @@
   protected final boolean _pausePartitionOnError;
   protected final Duration _pauseErrorPartitionDuration;
   protected final long _processingDelayLogThresholdMillis;
-  protected final boolean _enablePollDurationMillisMetric;
+  protected final boolean _enableAdditionalHistogramMetrics;
```
vmaheshw, line 106: can you make this name generic and avoid using Histogram

somandal: done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
```diff
@@ -40,6 +40,10 @@
   public static final String NUM_TOPICS = "numTopics";
   // keeps track of how long it takes to return from poll()
   public static final String POLL_DURATION_MS = "pollDurationMs";
+  // keeps track of how long processing takes between polls
+  public static final String TIME_SPENT_BETWEEN_POLLS_MS = "timeSpentBetweenPollsMs";
+  // keeps track of how long processing and sending all events received on poll takes
+  public static final String EVENT_PROCESSING_TIME_MS = "eventProcessingTimeMs";
```
vmaheshw, line 46: The name is slightly misleading. should there be a total somewhere in the name to be clear that it is not per event Processing time, but the total processing time of all the events received in one poll? One more suggestion, should we average here and add (total time/ total events), so it will give clear idea how much time we are spending to process per record. There is one more metric that might be needed how much time the producer send call took even though we know it is asynchronous. We can add that in future, if we see eventProcessingTime to be the culprint.

somandal: done, converted it to perEventProcessingTimeMs. For producer send, we already track time since send() completes and the send callback completes in the send latency metrics (EventProducer). This processing time only includes a) time taken to translate the record and b) time blocked on send() (we can block if the producer buffer is full or if we're waiting on metadata update)

vmaheshw (approved):
LGTM

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -515,9 +520,14 @@ protected void processRecords(ConsumerRecords<?, ?> records, Instant readTime) {
...
     translateAndSendBatch(records, readTime);
 
-    if (System.currentTimeMillis() - readTime.toEpochMilli() > _processingDelayLogThresholdMillis) {
+    long processingTimeMillis = System.currentTimeMillis() - readTime.toEpochMilli();
+    if (processingTimeMillis > _processingDelayLogThresholdMillis) {
       _consumerMetrics.updateProcessingAboveThreshold(1);
     }
+
+    if (_enableAdditionalMetrics) {
+      _consumerMetrics.updatePerEventProcessingTimeMs(records.count() == 0 ? 0 : processingTimeMillis / records.count());
```
vishwajith-s, line 529: I dont think we will encounter a case where records count is 0 cause processRecords is called if records is not empty. Saying that its a good check to have.

vishwajith-s (approved):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: Yeah I did notice that, but just added the check to be extra careful. Will leave this as is for now. Thanks!

PR #765, opened 2020-10-13
Add ref count in metrics registeration/deregistration.

vmaheshw:
DynamicMetricManager currently returns the existing metrics if exists or new metrics when trying to register metrics based on the full key name. So, if the same metrics is registered by 2 class/threads, DynamicMetricsManager is not aware of the ref count. So, when anyone of the class/thread de-register the shared metrics, the metrics is de-registered and no longer emitted. 
Currently BrooklinMetrics has a ref-count check for aggregate metrics to make sure the de-registration happens when all the registration references have de-registered. This is how aggregate metrics was working. The assumption may be that we will not see this conflict in any other case.

For KafkaMirrorMakerConnectorTask, the metrics are registered using datastream name and not the task name, so there is a possibility of the conflict. So, in the PR, I'm adding the ref count logic in DynamicMetricsManager itself. I'm also cleaning up BrooklinMetrics which has the logic for aggregate metrics, as we don't need to special case it. I'm keeping the existing interfaces for backward compatibility and can be cleaned up in future if required.


datastream-common/src/main/java/com/linkedin/datastream/metrics/DynamicMetricsManager.java
```diff
@@ -44,11 +44,16 @@
   // This is created solely for the createOrUpdate APIs, not by registerMetric because the former can be called
   // repeatedly to update the metric whereas the latter is typically only called once per metric during initialization.
   private final ConcurrentHashMap<String, ConcurrentHashMap<String, ConcurrentHashMap<String, Metric>>> _indexedMetrics;
+
+  // Map to maintain the ref count for the registered metric. When the ref count becomes zero or not the key is not present,
```
somandal, line 48: nit: zero or not the key is not present, -> zero or if the key is not present,

datastream-common/src/main/java/com/linkedin/datastream/metrics/BrooklinMetrics.java
```diff
@@ -5,19 +5,11 @@
...
-import java.util.HashMap;
-import java.util.Map;
-import java.util.concurrent.atomic.AtomicInteger;
-
-
 /**
  * Base class for metric categories with support for metric deregistration.
- * It uses reference counting for deregistering aggregate metrics which can
- * only happen when all keyed metrics of the same name have been deregistered.
+ * Underlying DynamicMetricsManager will take care of ref-count.
```
somandal, line 10: nit:  reword: Underlying DynamicMetricsManager will take care of maintaining ref-counts.

somandal (dismissed):
lgtm.
Is there a way to add tests for that race condition, and validate that this fixes it?

somandal (approved):

datastream-common/src/main/java/com/linkedin/datastream/metrics/DynamicMetricsManager.java
```diff
@@ -161,7 +166,7 @@ private void updateCache(String simpleClassName, String key, String metricName,
     String fullMetricName = MetricRegistry.name(simpleName, key, metricName);
 
     Metric metric = getMetric(fullMetricName, metricClass);
-
+    _registeredMetricRefCount.compute(fullMetricName, (localKey, val) -> (val == null) ? 1 : val + 1);
```
vishwajith-s, line 169: Nitpick: Can we move this to line 167?

vmaheshw: the reason to keep it here, because getMetric can throw exception if there is a name conflict with other metric type.

datastream-common/src/main/java/com/linkedin/datastream/metrics/DynamicMetricsManager.java
```diff
@@ -241,11 +246,13 @@ private void updateCache(String simpleClassName, String key, String metricName,
   public void unregisterMetric(String simpleName, String key, String metricName) {
     validateArguments(simpleName, metricName);
     String fullMetricName = MetricRegistry.name(simpleName, key, metricName);
-
+    _registeredMetricRefCount.computeIfPresent(fullMetricName, (localKey, val) -> (val == 1) ? null : val - 1);
```
vishwajith-s, line 249: Nitpick: Why do we need computeIfPresent here? Shouldn't we be using compute instead, unless someone is calling unregister without registering a metric?

vmaheshw: Because MetricRegister.remove handles unregistration without registering a metric.

`/**
     * Removes the metric with the given name.
     *
     * @param name the name of the metric
     * @return whether or not the metric was removed
     */
    public boolean remove(String name) {
        final Metric metric = metrics.remove(name);
        if (metric != null) {
            onMetricRemoved(name, metric);
            return true;
        }
        return false;
    }`

datastream-common/src/main/java/com/linkedin/datastream/metrics/BrooklinMetrics.java
```diff
@@ -5,19 +5,11 @@
...
-import java.util.concurrent.atomic.AtomicInteger;
-
-
 /**
  * Base class for metric categories with support for metric deregistration.
- * It uses reference counting for deregistering aggregate metrics which can
- * only happen when all keyed metrics of the same name have been deregistered.
+ * Underlying DynamicMetricsManager will take care of maintaining ref-counts.
  */
 public abstract class BrooklinMetrics {
```
vishwajith-s, line 12: We have an empty class just to deregister aggregate metrics now. Is it possible to explicitly call  deregister Aggregates in a backward compatible way?

vishwajith-s (commented):
Just some minor nitpicks.

datastream-common/src/main/java/com/linkedin/datastream/metrics/BrooklinMetrics.java
vmaheshw: As mentioned in the description, "I'm keeping the existing interfaces for backward compatibility and can be cleaned up in future if required." Since, the changes are in other components, I'm keeping the changes here. Once I will migrate that code, I will remove this method from here in the future.

vishwajith-s (approved):

PR #767, opened 2020-10-16
Fix the BMM event processing time metric to be in nanoseconds instead of milliseconds

somandal:
Deployed this code in our certification environment and saw that the metric for event processing time was showing up as 0 in the 50 percentile graphs, and very tiny numbers in the 99 percentile graph. Reading some documentation indicated using System.nanoTime() for doing elapsed time calculations as it doesn't use the wall clock (which can lead to issues due to leap seconds, etc). Modified the metric to use nanoTime() instead. To be clear, the actual precision of this  depends on what is provided by the OS, but for our purposes we don't need exact precision, we just need meaningful graphs.

I have decided to keep readTime in processRecords() as is. The reason for this is twofold. First, we do calculation in a couple of locations using this, and some of these involve API changes to translate(). Secondly, nanoTime() by itself is not useful, it is only useful when subtracting from another nanoTime() measurement. Thus the following code snippet cannot be used:

```
    long eventsSourceTimestamp =
        fromKafka.timestampType() == TimestampType.LOG_APPEND_TIME ? fromKafka.timestamp() : readTime.toEpochMilli();
```

Since we do not have control over how translate() uses readTime, it is not a good idea to modify it to only take the readNanoTime.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -376,8 +376,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
           _kafkaTopicPartitionTracker.onPartitionsPolled(records);
 
           Instant readTime = Instant.now();
-          long readTimeInNanos = System.nanoTime();
-          processRecords(records, readTime, readTimeInNanos);
+          processRecords(records, readTime, System.nanoTime());
           recordsPolled = records.count();
         }
         maybeCommitOffsets(_consumer, false);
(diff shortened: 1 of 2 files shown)


DEEPTHIKORAT (dismissed):
lgtm.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -376,7 +376,8 @@ public void run() {
           _kafkaTopicPartitionTracker.onPartitionsPolled(records);
 
           Instant readTime = Instant.now();
-          processRecords(records, readTime);
+          long readTimeInNanos = System.nanoTime();
+          processRecords(records, readTime, readTimeInNanos);
```
vishwajith-s, line 380: Nitpick. Can't we just pass System.nanoTime() to processRecords directly here cause I dont see we logging this variable in any other way.

vishwajith-s (dismissed):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: done

vishwajith-s (approved):

DEEPTHIKORAT (approved):

PR #768, opened 2020-10-16
Add a configurable feature to include the datastream name in the consumer client.id for BMM

somandal:
Currently for BMM the consumer client.id must be specified in the configs at the cluster + connector name level. This means that if multiple datastreams are created with the same connector type, they will all get the same client.id. This PR adds a configurable feature to append the datastream name to the client.id in the configs. The advantage of this is that Kafka consumer metrics (created based on client.id) can be created separately for every datastream rather than putting them all into the same bucket. This can help with debugging where we need Kafka consumer client metrics.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorConfig.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorConfig.java
@@ -32,6 +32,7 @@ public class KafkaBasedConnectorConfig {
   public static final String CONFIG_PAUSE_PARTITION_ON_ERROR = "pausePartitionOnError";
   public static final String CONFIG_PAUSE_ERROR_PARTITION_DURATION_MILLIS = "pauseErrorPartitionDurationMs";
   public static final String ENABLE_ADDITIONAL_METRICS = "enableAdditionalMetrics";
+  public static final String INCLUDE_DATASTREAM_NAME_IN_CONSUMER_CLIENT_ID = "includeDatastreamNameInConsumerClientId";
   public static final String DAEMON_THREAD_INTERVAL_SECONDS = "daemonThreadIntervalInSeconds";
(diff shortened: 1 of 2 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorConfig.java
```diff
@@ -32,6 +32,7 @@
   public static final String CONFIG_PAUSE_PARTITION_ON_ERROR = "pausePartitionOnError";
   public static final String CONFIG_PAUSE_ERROR_PARTITION_DURATION_MILLIS = "pauseErrorPartitionDurationMs";
   public static final String ENABLE_ADDITIONAL_METRICS = "enableAdditionalMetrics";
+  public static final String INCLUDE_DATASTREAM_NAME_IN_CONSUMER_CLIENT_ID = "includeDatastreamNameInConsumerClientId";
```
vishwajith-s, line 35: Do we need to make this a configurable? Why not add it without the config?

vishwajith-s (approved):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorConfig.java
somandal: Want to make this configurable for two reasons:

1. Kafka quotas are currently enforced based on client.id. Kafka is in the process of migrating away from this model, but I don't want to break the client.ids until we are sure that the migration is complete and successful. So was thinking we can enable this on a per cluster basis.
2. I don't expect us to hit metrics limitations in BMM, but want to make sure we have a knob to turn this off in case we start having problems with metrics.

Once we roll this out everywhere and all looks good, I can make a new PR to remove the config. How does that sound?

celiakung (approved):

PR #770, opened 2020-10-20
Handle the new session after session expiry

vmaheshw:
This is the final change to handle new session after session expiry. In this change, we have re-initialized all the local states, listeners, event threads and made the node re-join the cluster.


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1378,13 +1395,17 @@ public void releaseTask(DatastreamTaskImpl task) {
...
-          task.getDatastreamTaskName());
+    try {
+      String owner = _zkclient.ensureReadData(lockPath);
+      if (!owner.equals(_instanceName)) {
+        LOG.warn("{} does not have the lock on {}-{}/{}", _instanceName, task.getConnectorType(), task.getTaskPrefix(),
+            task.getDatastreamTaskName());
+        return;
+      }
+    } catch (ZkNoNodeException e) {
+      LOG.info("Task lock node may be got deleted. exists: {}", _zkclient.exists(lockPath));
```
DEEPTHIKORAT, line 1406: nit: Reword to `Task lock node might have been deleted`
If we got ZkNoNodeException doesn't that mean lockPath doesn't exist or is that a more generic exception?

DEEPTHIKORAT, line 1406: nit: Reword to Task lock node might have been deleted
If we got ZkNoNodeException doesn't that mean lockPath doesn't exist or is that a more generic exception?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1554,12 +1580,16 @@ public ZkBackedLiveInstanceListProvider() {
...
-        if (hostname == null) {
-          // hostname can be null if a node dies immediately after reading all live instances
-          LOG.error("Node {} is dead. Likely cause it dies after reading list of nodes.", n);
-        } else {
-          liveInstances.add(formatZkInstance(hostname, n));
+        try {
+          String hostname = _zkclient.ensureReadData(KeyBuilder.liveInstance(_cluster, n));
+          if (hostname == null) {
+            // hostname can be null if a node dies immediately after reading all live instances
+            LOG.error("Node {} is dead. It died after the list of nodes were read.", n);
```
DEEPTHIKORAT, line 1587: Can you please reword this? Node dying is a bit confusing to read. The actual event might be helpful as in.. instance's session expired or something like that. Same for the exception below.

DEEPTHIKORAT (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1554,12 +1580,16 @@ public ZkBackedLiveInstanceListProvider() {
...
+          } else {
+            liveInstances.add(formatZkInstance(hostname, n));
+          }
+        } catch (ZkNoNodeException e) {
+          LOG.info("Node {} is dead. It died after the list of nodes were read.", n);
```
somandal, line 1592: can we make this a warn log?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
somandal: +1
Can we also make this a warn log?

somandal (approved):
I recall we had a discussion earlier where we said that when a node expires, and then reconnects, it may need to update the locks it used to hold (if it still has the same tasks) to indicate that it's the new live instance, right? Should this fix be addressed here?

The PR where we discussed this: https://github.com/linkedin/brooklin/pull/747/
Look for this comment:
"Quick question, if a Session expiry happens, the _instanceName remains the same? Just wondering if we could have a case where we're trying to release the lock but an expiry + connect happened before we call this, creating a new liveinstance node for this host. Will the task still be releasable?"

DEEPTHIKORAT (approved):

vmaheshw:
> > I recall we had a discussion earlier where we said that when a node expires, and then reconnects, it may need to update the locks it used to hold (if it still has the same tasks) to indicate that it's the new live instance, right? Should this fix be addressed here?
> > The PR where we discussed this: #747
> > Look for this comment:
> > "Quick question, if a Session expiry happens, the _instanceName remains the same? Just wondering if we could have a case where we're trying to release the lock but an expiry + connect happened before we call this, creating a new liveinstance node for this host. Will the task still be releasable?"
> > Not sure if such a fix is required, but it'll be great if you can validate and explain why or why not.
> 
> @vmaheshw can you please look at this comment and leave a response about whether this is a concern or not. If it is, please address it. If it is not, please explain why not. I just want to ensure that there is no weird race conditions that we need to think about here even though from what I understand this shouldn't be a problem.

Sorry, I forgot to reply. Yes, if the expiry+connect happened, before trying to release the lock, it will get the error "Not the owner" (As the owner was previous instance), then the task will move to some other instance and that instance while trying to acquire lock, will find this as orphan lock and force acquire it.

somandal (approved):

PR #773, opened 2020-11-02
Add a configurable max tasks per instance threshold

somandal:
We've run into issues where too many tasks get assigned to a given instance resulting in OOMs. This PR adds a configurable option to limit the number of tasks per instance at the cluster level. This works the same way irrespective of the assignment strategy used, and is at a global level for all tasks (across multiple connector types). A value of 0 indicates that this is disabled.

the contribution guidelines.

### DIFF
--- a/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
+++ b/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
@@ -48,6 +48,7 @@ import com.linkedin.datastream.common.Datastream;
 import com.linkedin.datastream.common.DatastreamConstants;
 import com.linkedin.datastream.common.DatastreamException;
 import com.linkedin.datastream.common.DatastreamMetadataConstants;
+import com.linkedin.datastream.common.DatastreamRuntimeException;
 import com.linkedin.datastream.common.DatastreamStatus;
 import com.linkedin.datastream.common.DatastreamUtils;
 import com.linkedin.datastream.common.JsonUtils;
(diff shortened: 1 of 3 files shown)


datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -656,6 +657,86 @@ public void testCoordinationWithStickyMulticastStrategy() throws Exception {
     zkClient.close();
   }
 
+  /**
+   * testCoordinationWithStickyMulticastStrategyAndMaxTaskLimit is a test to verify that assignment cannot complete
+   * when the tasks per instance is exceeded by the configured threshold
```
DEEPTHIKORAT, line 662: nit:  `when the tasks per instance exceeds the configured threshold`

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -656,6 +657,86 @@ public void testCoordinationWithStickyMulticastStrategy() throws Exception {
...
+    String testCluster = "testCoordinationMaxTasksPerInstance";
+    String testConnectorType = "testConnectorType";
+    Properties properties = new Properties();
+    // Set max tasks per instance to 5, since some instances will get 6 tasks if 4 datastreams with 4 tasks each are
+    // created across 3 instances. With 4 instances, the tasks per instance will be less than 5.
+    properties.put(CoordinatorConfig.CONFIG_MAX_DATASTREAM_TASKS_PER_INSTANCE, "5");
+    Coordinator instance1 = createCoordinator(_zkConnectionString, testCluster, properties);
+
+    TestHookConnector connector1 = new TestHookConnector("connector1", testConnectorType);
+    //Question why the multicast strategy is within one coordinator rather than shared between list of coordinators
```
DEEPTHIKORAT, line 680: Is the comment a todo that you had to remove?

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -656,6 +657,86 @@ public void testCoordinationWithStickyMulticastStrategy() throws Exception {
...
+    List<TestHookConnector> connectors = new ArrayList<>();
+    connectors.add(connector1);
+    connectors.add(connector2);
+    connectors.add(connector3);
+    List<String> datastreamNames = ImmutableList.of("datastream1", "datastream2", "datastream3", "datastream4");
+
+    for (String name : datastreamNames) {
+      DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, testConnectorType, name);
+    }
+    boolean assignmentComplete = waitTillAssignmentIsComplete(16, WAIT_TIMEOUT_MS,
```
DEEPTHIKORAT, line 707: Is there are way to make this test more unit-test like? Just test `validateNewAssignment` I mean? My concern is that if assignment fails for any other reason, it would still look the same for the reader of waitTillAssignmentIsComplete's output even though it actually did not fail due to max tax limit if I am reading this correct

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1322,6 +1323,23 @@ private void performPartitionMovement(Long notifyTimestamp) {
     }
   }
 
+  private void validateNewAssignment(Map<String, List<DatastreamTask>> newAssignmentsByInstance) {
+    if (_config.getMaxDatastreamTasksPerInstance() > 0) {
+      // If the cluster was configured to limit the max tasks per instance, check if any instances have a higher
+      // number of tasks than expected, and fail the leader assignment on violation of this limit. This can be useful
+      // to prevent other issues such as OOMs due to high memory usage which may be seen if we exceed the supportable
+      // number of tasks per instance.
+      Map<String, Integer> instancesWithTaskCountAboveThreshold = newAssignmentsByInstance.entrySet().stream()
```
DEEPTHIKORAT, line 1332: Should we also consider doing assignments only if a "quorum" number of instances are participating in the membership?

DEEPTHIKORAT (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
somandal: Discussed this with @DEEPTHIKORAT offline. One problem with using a "quorum" is that it only helps catch cases when the number of tasks within the cluster are within the limit overall, but for some reason some of the instances are down.

There can be another case, where someone adds a new datastream or bumps the maxTasks in such a way that some of instances get a higher number of tasks than the instance can support without running into problems (e.g. OOMs). In such cases, the right thing to do is to add additional instances to the cluster such that the higher capacity can be supported.

So for now using a limit on tasks per instance is sufficient.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
somandal: done

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
somandal: Added a new unit test which is more like a unit test. Wanted to keep this one, since we are testing that increasing the instances allows the assignment to complete.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
somandal: done - removed it. was a copy paste error

vmaheshw (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1322,6 +1323,24 @@ private void performPartitionMovement(Long notifyTimestamp) {
     }
   }
 
+  @VisibleForTesting
+  void validateNewAssignment(Map<String, List<DatastreamTask>> newAssignmentsByInstance) {
+    if (_config.getMaxDatastreamTasksPerInstance() > 0) {
+      // If the cluster was configured to limit the max tasks per instance, check if any instances have a higher
```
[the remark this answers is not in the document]
somandal: done

vmaheshw (approved):

DEEPTHIKORAT (approved):

PR #774, opened 2020-11-03
Add preAssignmentCleanup in Coordinator and extend assignment strategy

vmaheshw:
In StickyMulticastPartitionAssignmentStrategy, during assignment, if the partition count changes for a task A, the new task A' is created and the existing task A is added to the dependency list. The existing task A is removed from the assignment list. The coordinator then updates the assignment to the zookeeper, by first adding the new node A' followed by deleting the old node A, to avoid assignment loss. If the coordinator either gets interrupted, or hits OOM, or session expiry, before completing the assignment update, there is a possibility that the deletion of the existing task A did not succeed. This results in both the new task A' and old task A exist in the assignment list. The next leader has no clue that the old task A needs to be cleaned up and ends up reusing the old task A (which should never happen). This results in the old task A continue to hold the task lock and the new task A' wait for the old task A to release the lock, to acquire and start processing. So, the new task A' will never run. Now, suppose another assignment is performed, where the new task A' is going to get added to the dependency list of task A'', the task  A'' object creation will fail, as it checks that the dependency should already have the lock in the constructor of DatastreamTaskImpl. This check is important to avoid A' and A'' fight for the task lock at the same time.

Partition-Managed Kafka connector assignment is in two steps: a, assign the task b, assign the partition(which modifies the task as well).  Because of these 2 step process, it is not possible for the new leader to identify the tasks that need to be cleaned up. Also, the clean up of the old task should complete before the new assignment for 2 reasons:
a. to prevent the old task to continue running and fail the partition assignment repeatedly. (The only way to recover the setup in this case is to restart the datastreams)
b. Because old task has no dependency, if the old task moved to another instance, there is a possibility that task A' is able to acquire the lock and task A moving to another instance will also be able to acquire the task lock because it has no dependency. This will result in both task A and task A' running simultaneously, which will result in overlapping partition record processing.

To solve this problem, This PR is introducing preassignment cleanup which will allow the strategy to identify the stale tasks and allow the coordinator to clean them. This will be done only once when the node becomes leader for the first time.


datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
```diff
@@ -89,4 +90,16 @@
       Map<String, Set<String>> targetAssignment, DatastreamGroupPartitionsMetadata partitionsMetadata) {
     throw new NotImplementedException("movePartitions are not implemented");
   }
+
+  /**
+   * Validate the current assignment and return the list of DatastreamTask per instance to be cleaned up.
+   *
+   * @param datastreams all the datastreams for which the associated tasks to be validated
```
jzakaryan, line 97: Maybe "need to be validated"?

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorConfig.java
```diff
@@ -59,6 +61,7 @@ public CoordinatorConfig(Properties config) {
     _defaultTransportProviderName = _properties.getString(CONFIG_DEFAULT_TRANSPORT_PROVIDER, "");
     _zkCleanUpOrphanConnectorTask = _properties.getBoolean(CONFIG_ZK_CLEANUP_ORPHAN_CONNECTOR_TASK, false);
     _zkCleanUpOrphanConnectorTaskLock = _properties.getBoolean(CONFIG_ZK_CLEANUP_ORPHAN_CONNECTOR_TASK_LOCK, false);
+    _performPreAssignmentCleanup = _properties.getBoolean(CONFIG_PERFORM_PRE_ASSIGNMENT_CLEANUP, true);
```
jzakaryan, line 64: Do you think we should turn on this feature by default without testing it in production?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -268,6 +271,46 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
     return newAssignment;
   }
 
+  /**
+   * This method checks the current assignment and returns the list of the task which are in the
```
jzakaryan, line 275: Nit: `list of tasks`

Also could be reworded to:

```
If a task is both in the dependency list as well as the current assignment list, it's possible that the previous leader went down before getting a chance to clean it up and update the zookeeper. This method identifies such tasks.
```

jzakaryan (changes requested):
Minor improvements to inline documentation.

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyPartitionAssignment.java
```diff
@@ -87,11 +86,24 @@ public void testAddPartitions() {
...
+
+    // Adding the dependency task as well in the assignment list to simulate the scenario where
+    // the dependency task nodes are not deleted and the leader gets interrupted, OOM or hit session expiry.
+    // The next leader should be able to identify and cleanup.
+    Map<String, Set<DatastreamTask>> finalAssignment = assignment;
+    newAssignment.forEach((instance, taskSet1) -> taskSet1.addAll(finalAssignment.get(instance)));
+
+    taskToCleanup = strategy.getTasksToCleanUp(datastreams, newAssignment);
+    Assert.assertEquals(taskToCleanup.size(), 1);
+    taskToCleanup.forEach((instance, taskList1) -> Assert.assertEquals(taskList1.size(), 3));
```
somandal, line 106: Should we also validate that the tasksToCleanup only contains the tasks we expect it to by matching task names?

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorConfig.java
vmaheshw: From testing, you mean we should just identify and print them?

somandal: No, at least I meant, let's disable this by default, enable it through config when testing in cert, and once it looks good, add the config to enable it everywhere.

somandal (approved):
lgtm

jzakaryan (approved):

somandal (approved):

jzakaryan (approved):

PR #775, opened 2020-11-04
Fix multiple issues found in the partition movement code

somandal:
This PR fixes multiple issues in the partition movement code:

- If a task has no change in partitions, do not create a new task. Earlier all tasks were getting added to the tasks to mutate list,  even if it did not contain any partitions going away or getting added
- If the target instance receives a partition which it already owns, this should be a no-op, and the partition should remain in it's original task.
- Only add task dependencies for partitions getting added to that task rather than all partition being moved. E.g. earlier if p1 and p2 were to be moved, and p1 got added to task1, then p1's task and p2's task were both added as dependencies. Only p1's task should be added as a dependency to task1.
- When adding extra dependencies, ensure that these are locked too (like we do with the predecessor task)

This PR does not change the overall logic of this piece.

the contribution guidelines.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
@@ -152,12 +152,12 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy {
   /**
    * Move a partition for a datastream group according to the targetAssignment. As we are only allowed to mutate the
    * task once. It follow the steps
-   * Step 1) Preprocess the targetAssignment to remove any partitions with no-op moves (partition currently assigned
+   * Step 1) Pre-process the targetAssignment to remove any partitions with no-op moves (partition currently assigned
    *         to the same instance where the partition is to be moved)
-   * Step 2) Get the partitions that are to be moved, and find their source task
+   * Step 2) Get the partitions that are to be moved, and find their source tasks
    * Step 3) If the instance is the one we want to move, we choose a task to which we should assign the partition
    *         from that instance
-   * Step 4) Scan the current assignment, compute the new task if the old task belongs to this source tasks or if it
+   * Step 4) Scan the current assignment, compute the new task if the old task belongs to this source task or if it
    *         is the target task we want to move the partition to
    *
    * @param currentAssignment the old assignment
@@ -176,23 +176,22 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy {
     // Pre-process the targetAssignment to filter out partitions that are supposed to move to the target instance that
     // they are already on to avoid unnecessary partition movements. That is, if partition1 is currently on instance1
     // and is supposed to move to instance1, remove it from the targetAssignment as this should be a no-op.
-    Map<String, Set<String>> preprocessedTargetAssignment = new HashMap<>();
+    Map<String, Set<String>> preProcessedTargetAssignment = new HashMap<>();
     targetAssignment.forEach((instance, partitions) -> {
-      Set<DatastreamTask> tasks = currentAssignment.get(instance);
-      Set<String> partitionsAcrossAllDatastreamTasks = tasks.stream().filter(dg::belongsTo)
+      Set<String> partitionsAcrossAllDatastreamTasks = currentAssignment.get(instance).stream().filter(dg::belongsTo)
           .map(DatastreamTask::getPartitionsV2).flatMap(Collection::stream).collect(Collectors.toSet());
 
       Set<String> updatedTargetPartitionList = new HashSet<>(partitions);
       updatedTargetPartitionList.removeAll(partitionsAcrossAllDatastreamTasks);
 
-      preprocessedTargetAssignment.computeIfAbsent(instance,
+      preProcessedTargetAssignment.computeIfAbsent(instance,
           val -> updatedTargetPartitionList.isEmpty() ? null : updatedTargetPartitionList);
     });
 
-    LOG.info("The preprocessed targetAssignment: {}", preprocessedTargetAssignment);
+    LOG.info("The pre-processed targetAssignment: {}", preProcessedTargetAssignment);
 
     Set<String> allToReassignPartitions = new HashSet<>();
-    preprocessedTargetAssignment.values().forEach(allToReassignPartitions::addAll);
+    preProcessedTargetAssignment.values().forEach(allToReassignPartitions::addAll);
     allToReassignPartitions.retainAll(partitionsMetadata.getPartitions());
 
     // construct a map to store the tasks and if it contain the partitions that can be released
@@ -232,9 +231,9 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy {
 
       // check if this instance has any partition to be added
       final Set<String> toAddedPartitions = new HashSet<>();
-      if (preprocessedTargetAssignment.containsKey(instance)) {
+      if (preProcessedTargetAssignment.containsKey(instance)) {
         // filter the target assignment by the partitions which have a confirmed source
-        Set<String> p = preprocessedTargetAssignment.get(instance).stream()
+        Set<String> p = preProcessedTargetAssignment.get(instance).stream()
             .filter(toReleasePartitions::contains).collect(Collectors.toSet());
         toAddedPartitions.addAll(p);
       }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
@@ -6,6 +6,7 @@
 package com.linkedin.datastream.server.assignment;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -151,11 +152,13 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy {
   /**
    * Move a partition for a datastream group according to the targetAssignment. As we are only allowed to mutate the
    * task once. It follow the steps
-   * Step 1) get the partitions that to be moved, and find out their source task
-   * Step 2) If the instance is the one we want to move, we find out the task which we should assign the partition
+   * Step 1) Preprocess the targetAssignment to remove any partitions with no-op moves (partition currently assigned
+   *         to the same instance where the partition is to be moved)
+   * Step 2) Get the partitions that are to be moved, and find their source task
+   * Step 3) If the instance is the one we want to move, we choose a task to which we should assign the partition
    *         from that instance
-   * Step 3) Scan the current assignment, compute new task if the old task belongs to these source tasks or if it
-   *         is the target task we want to move to
+   * Step 4) Scan the current assignment, compute the new task if the old task belongs to this source tasks or if it
+   *         is the target task we want to move the partition to
    *
    * @param currentAssignment the old assignment
    * @param targetAssignment the target assignment retrieved from Zookeeper
@@ -170,8 +173,26 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy {
 
     DatastreamGroup dg = partitionsMetadata.getDatastreamGroup();
 
+    // Pre-process the targetAssignment to filter out partitions that are supposed to move to the target instance that
+    // they are already on to avoid unnecessary partition movements. That is, if partition1 is currently on instance1
+    // and is supposed to move to instance1, remove it from the targetAssignment as this should be a no-op.
+    Map<String, Set<String>> preprocessedTargetAssignment = new HashMap<>();
+    targetAssignment.forEach((instance, partitions) -> {
+      Set<DatastreamTask> tasks = currentAssignment.get(instance);
+      Set<String> partitionsAcrossAllDatastreamTasks = tasks.stream().filter(dg::belongsTo)
+          .map(DatastreamTask::getPartitionsV2).flatMap(Collection::stream).collect(Collectors.toSet());
+
+      Set<String> updatedTargetPartitionList = new HashSet<>(partitions);
+      updatedTargetPartitionList.removeAll(partitionsAcrossAllDatastreamTasks);
+
+      preprocessedTargetAssignment.computeIfAbsent(instance,
+          val -> updatedTargetPartitionList.isEmpty() ? null : updatedTargetPartitionList);
+    });
+
+    LOG.info("The preprocessed targetAssignment: {}", preprocessedTargetAssignment);
+
     Set<String> allToReassignPartitions = new HashSet<>();
-    targetAssignment.values().forEach(allToReassignPartitions::addAll);
+    preprocessedTargetAssignment.values().forEach(allToReassignPartitions::addAll);
     allToReassignPartitions.retainAll(partitionsMetadata.getPartitions());
 
     // construct a map to store the tasks and if it contain the partitions that can be released
@@ -179,8 +200,8 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy {
     Map<String, Set<String>> confirmedPartitionsTaskMap = new HashMap<>();
 
     // construct a map to store the partition and its source task
-    // map: <partitions that need to be released, source taskName>
-    Map<String, String> partitionToSourceTaskMap = new HashMap<>();
+    // map: <partitions that need to be released, source task>
+    Map<String, DatastreamTaskImpl> partitionToSourceTaskMap = new HashMap<>();
 
     // We first confirm that the partitions in the target assignment which can be removed, and we find out its source task
     // If the partitions cannot be found from any task, we ignore these partitions
@@ -190,15 +211,18 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy {
         if (dg.belongsTo(task)) {
           Set<String> toMovePartitions = new HashSet<>(task.getPartitionsV2());
           toMovePartitions.retainAll(allToReassignPartitions);
-          confirmedPartitionsTaskMap.put(task.getDatastreamTaskName(), toMovePartitions);
-          toMovePartitions.forEach(p -> partitionToSourceTaskMap.put(p, task.getDatastreamTaskName()));
+          if (!toMovePartitions.isEmpty()) {
+            // Only update the confirmedPartitionsTaskMap if a partition is indeed being deleted from it
+            confirmedPartitionsTaskMap.put(task.getDatastreamTaskName(), toMovePartitions);
+            toMovePartitions.forEach(p -> partitionToSourceTaskMap.put(p, (DatastreamTaskImpl) task));
+          }
         }
       });
     });
 
     Set<String> tasksToMutate = confirmedPartitionsTaskMap.keySet();
     Set<String> toReleasePartitions = new HashSet<>();
-    confirmedPartitionsTaskMap.values().forEach(v -> toReleasePartitions.addAll(v));
+    confirmedPartitionsTaskMap.values().forEach(toReleasePartitions::addAll);
 
     // Compute new assignment from the current assignment
     Map<String, Set<DatastreamTask>> newAssignment = new HashMap<>();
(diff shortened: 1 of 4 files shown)


datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyPartitionAssignment.java
```diff
@@ -182,14 +177,13 @@ public void testMovePartitionToInstanceWithoutTask() {
...
-    assignment.put("empty", new HashSet<DatastreamTask>());
+    assignment.put("empty", new HashSet<>());
 
     Map<String, Set<String>> targetAssignment = new HashMap<>();
     targetAssignment.put("empty", ImmutableSet.of("t-3", "t-2", "t-1", "t-5"));
-    assignment = strategy.movePartitions(assignment, targetAssignment, partitionsMetadata);
+    strategy.movePartitions(assignment, targetAssignment, partitionsMetadata);
   }
 
-
```
jzakaryan, line 192: Nit: Avoid unnecessary changes to make reviews shorter.

jzakaryan (dismissed):
LGTM

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -151,11 +152,13 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
-   * Step 1) get the partitions that to be moved, and find out their source task
-   * Step 2) If the instance is the one we want to move, we find out the task which we should assign the partition
+   * Step 1) Preprocess the targetAssignment to remove any partitions with no-op moves (partition currently assigned
+   *         to the same instance where the partition is to be moved)
+   * Step 2) Get the partitions that are to be moved, and find their source task
+   * Step 3) If the instance is the one we want to move, we choose a task to which we should assign the partition
    *         from that instance
-   * Step 3) Scan the current assignment, compute new task if the old task belongs to these source tasks or if it
-   *         is the target task we want to move to
+   * Step 4) Scan the current assignment, compute the new task if the old task belongs to this source tasks or if it
```
vmaheshw, line 160: nits: tasks to task

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -151,11 +152,13 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
   /**
    * Move a partition for a datastream group according to the targetAssignment. As we are only allowed to mutate the
    * task once. It follow the steps
-   * Step 1) get the partitions that to be moved, and find out their source task
-   * Step 2) If the instance is the one we want to move, we find out the task which we should assign the partition
+   * Step 1) Preprocess the targetAssignment to remove any partitions with no-op moves (partition currently assigned
+   *         to the same instance where the partition is to be moved)
+   * Step 2) Get the partitions that are to be moved, and find their source task
```
vmaheshw, line 157: nit: task to tasks, preprocess to pre-process

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -170,17 +173,35 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
+
+      preprocessedTargetAssignment.computeIfAbsent(instance,
+          val -> updatedTargetPartitionList.isEmpty() ? null : updatedTargetPartitionList);
+    });
+
+    LOG.info("The preprocessed targetAssignment: {}", preprocessedTargetAssignment);
+
     Set<String> allToReassignPartitions = new HashSet<>();
-    targetAssignment.values().forEach(allToReassignPartitions::addAll);
+    preprocessedTargetAssignment.values().forEach(allToReassignPartitions::addAll);
```
vmaheshw, line 195: addAll with throw NullPointerException for the value set to null. The value is set to null for the instance where there is no movement.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -170,17 +173,35 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
 
     DatastreamGroup dg = partitionsMetadata.getDatastreamGroup();
 
+    // Pre-process the targetAssignment to filter out partitions that are supposed to move to the target instance that
+    // they are already on to avoid unnecessary partition movements. That is, if partition1 is currently on instance1
+    // and is supposed to move to instance1, remove it from the targetAssignment as this should be a no-op.
+    Map<String, Set<String>> preprocessedTargetAssignment = new HashMap<>();
+    targetAssignment.forEach((instance, partitions) -> {
+      Set<DatastreamTask> tasks = currentAssignment.get(instance);
+      Set<String> partitionsAcrossAllDatastreamTasks = tasks.stream().filter(dg::belongsTo)
```
vmaheshw, line 182: Can you merge these two lines?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -190,15 +211,18 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
         if (dg.belongsTo(task)) {
           Set<String> toMovePartitions = new HashSet<>(task.getPartitionsV2());
           toMovePartitions.retainAll(allToReassignPartitions);
-          confirmedPartitionsTaskMap.put(task.getDatastreamTaskName(), toMovePartitions);
-          toMovePartitions.forEach(p -> partitionToSourceTaskMap.put(p, task.getDatastreamTaskName()));
+          if (!toMovePartitions.isEmpty()) {
```
vmaheshw, line 214: I feel we can avoid splitting this in two steps. Please check and let me know your opinion.

``` // We first confirm that the partitions in the target assignment which can be removed, and we find out its source task
    // If the partitions cannot be found from any task, we ignore these partitions
    currentAssignment.keySet().forEach(instance -> {
      Set<DatastreamTask> tasks = currentAssignment.get(instance);
      Set<String> partitionsAcrossAllDatastreamTasks = currentAssignment.get(instance).stream().filter(dg::belongsTo)
          .map(DatastreamTask::getPartitionsV2).flatMap(Collection::stream).collect(Collectors.toSet());
      //Remove the partitions that do not need migration.
      allToReassignPartitions.removeAll(partitionsAcrossAllDatastreamTasks);
      tasks.forEach(task -> {
        if (dg.belongsTo(task)) {
          Set<String> toMovePartitions = new HashSet<>(task.getPartitionsV2());
          toMovePartitions.retainAll(allToReassignPartitions);
          if (!toMovePartitions.isEmpty()) {
            // Only update the confirmedPartitionsTaskMap if a partition is indeed being deleted from it
            confirmedPartitionsTaskMap.put(task.getDatastreamTaskName(), toMovePartitions);
            toMovePartitions.forEach(p -> partitionToSourceTaskMap.put(p, (DatastreamTaskImpl) task));
          }
        }
      });
    });`

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java [hunk 7]
```diff
@@ -228,37 +251,39 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
...
-        // compute new assignment for that instance
-        Set<DatastreamTask> newAssignedTask = tasks.stream().map(task -> {
-          if (!dg.belongsTo(task)) {
-            return task;
-          }
-          boolean partitionChanged = false;
-          List<String> newPartitions = new ArrayList<>(task.getPartitionsV2());
-          Set<String> extraDependencies = new HashSet<>();
+      // compute new assignment for that instance
+      Set<DatastreamTask> newAssignedTask = tasks.stream().map(task -> {
```
vmaheshw, line 255: Sorry for my ignorance. I'm not able to find anything changing in the next few code diff blocks, except empty lines. Just want to be sure that I'm not missing anything.

datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java
```diff
@@ -38,47 +38,84 @@
...
     Datastream stream = DatastreamTestUtils.createDatastream("dummy", "dummy", "dummy");
     stream.getMetadata().put(DatastreamMetadataConstants.TASK_PREFIX, DatastreamTaskImpl.getTaskPrefix(stream));
 
     DatastreamTaskImpl task = new DatastreamTaskImpl(Collections.singletonList(stream));
     ZkAdapter mockZkAdapter = mock(ZkAdapter.class);
     task.setZkAdapter(mockZkAdapter);
 
-    task.addDependency("task0");
+    DatastreamTaskImpl task0 = new DatastreamTaskImpl(Collections.singletonList(stream));
+    ZkAdapter mockZkAdapter2 = mock(ZkAdapter.class);
```
vmaheshw, line 50: nit: mockZkAdapter2 to mockZkAdapter0

datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java
```diff
@@ -38,47 +38,84 @@
...
+    when(mockZkAdapter2.checkIsTaskLocked(anyString(), anyString(), anyString())).thenReturn(true);
+
+    task.addDependency(task0);
     task.acquire(Duration.ofMillis(60));
     verify(mockZkAdapter, atLeastOnce()).waitForDependencies(any(DatastreamTaskImpl.class), any(Duration.class));
   }
 
   @Test(expectedExceptions = DatastreamTransientException.class)
-  public void testCreateNewTaskFromUnlockedTask() throws Exception {
+  public void testCreateNewTaskFromUnlockedTask() {
```
vmaheshw, line 60: Nit: Not your fault, Do you think this test can be changed to validate the exception using AssertThrows to ensure that all the blocks including the new code is getting executed.

datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java
```diff
@@ -38,47 +38,84 @@
...
     stream.getMetadata().put(DatastreamMetadataConstants.TASK_PREFIX, DatastreamTaskImpl.getTaskPrefix(stream));
 
     DatastreamTaskImpl task = new DatastreamTaskImpl(Collections.singletonList(stream));
-    task.addDependency("task0");
+
     ZkAdapter mockZkAdapter = mock(ZkAdapter.class);
     task.setZkAdapter(mockZkAdapter);
+
+    DatastreamTaskImpl task0 = new DatastreamTaskImpl(Collections.singletonList(stream));
+    ZkAdapter mockZkAdapter2 = mock(ZkAdapter.class);
```
vmaheshw, line 91: same nit: mockZkAdapter2 to 0, here and other places.

vmaheshw, line 91: Also, do you see a value of extracting the method for this new duplicate code block?

datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java
```diff
@@ -38,47 +38,84 @@
...
+
+    task.addDependency(task0);
+
     when(mockZkAdapter.checkIsTaskLocked(anyString(), anyString(), anyString())).thenReturn(true);
     DatastreamTaskImpl task2 = new DatastreamTaskImpl(task, new ArrayList<>());
     Assert.assertEquals(new HashSet<>(task2.getDependencies()), ImmutableSet.of(task.getDatastreamTaskName()));
   }
 
+  @Test(expectedExceptions = DatastreamTransientException.class)
+  public void testTaskAddUnlockedDependency() {
```
vmaheshw, line 103: Please transform this test to AssertThrows as well.

vmaheshw (commented):
Suggested an idea to avoid extra step. Please check and let me know your opinion.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
somandal: In the previous step, where we create `preprocessedTargetAssignment`, the `computeIfAbsent` does not add the key to the map if the computed value is null. I've modified this to filter out null values anyways just to be safe, but using `addAll` here without null checks should be safe.

Code that updates `preProcessedTargetAssignment`:
```
      preProcessedTargetAssignment.computeIfAbsent(instance,
          val -> updatedTargetPartitionList.isEmpty() ? null : updatedTargetPartitionList);
```

Comments on `computeIfAbsent`:
>      * If the specified key is not already associated with a value (or is mapped
>     * to {@code null}), attempts to compute its value using the given mapping
>     * function and enters it into this map unless {@code null}.

vmaheshw: Make sense. You can ignore this comment. 

Feel free to ignore this suggestion. Just wondering should we make this piece more simple.
```
if (!updatedTargetPartitionList.isEmpty()) {
preProcessedTargetAssignment.put(instance, updatedTargetPartitionList);
}

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java (hunk 6 above)
Comment (somandal) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:214]: I tried this as is, but it doesn't work (`preprocessedTargetAssignment` is used in a later step, so I also added code here to create this too) and it's also getting very tricky to get it to work. Just removing the `partitionsAcrossAllDatastreamTasks` from `allToReassignPartitions` results in `allToReassignPartitions` becoming  empty eventually. I tried creating a copy of `allToReassignPartitions` too, and removed from that, but  the logic wasn't correct. Tried to retain all of `partitionsAcrossAllDatastreamTasks`, then there are issues with creating the `confirmedPartitionsTaskMap`, and when I try to fix that then I see that some partitions are completely removed from the final assignment, etc.

At least for me, this code is quite tricky to reason with, so the two step assignment just feels cleaner and easier to understand. It may be okay to perform a two step process here, since it's just recalculating the assignments. What do you think?

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java (hunk 4 above)
Comment (somandal) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:195]: I sort of prefer `computeIfAbsent` here, so will leave this as is.

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java (hunk 7 above)
Comment (somandal) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:255]: yeah sorry, there was an indentation issue in this part of the code, so just fixed that. next time i'll do this as part of a separate change.

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java (hunk 3 above)
Comment (somandal) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:157]: done

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java (hunk 2 above)
Comment (somandal) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:160]: done

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java (hunk 5 above)
Comment (somandal) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:182]: done

File: datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java (hunk 8 above)
Comment (somandal) [reply datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java:50]: moved to a separate function

File: datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java (hunk 9 above)
Comment (somandal) [reply datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java:60]: done

File: datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java (hunk 10 above)
Comment (somandal) [reply datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java:91]: moved to a separate function

Comment (somandal) [reply datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java:91]: done

File: datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java (hunk 11 above)
Comment (somandal) [reply datastream-server/src/test/java/com/linkedin/datastream/server/TestDatastreamTask.java:103]: done

File: datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java (hunk 6 above)
Comment (somandal) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:220]: Found a way to make this work. Silly me should have tried moving the allToReassignPartitions to within the targetAssignment.containsKey() block.

Comment (vmaheshw) [reply datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java:220]: Can you please check this code block? This logic is correct (with one minor addition required) but not so easy to get it. 
```
 if (targetAssignment.containsKey(instance)) {
        // Process the targetAssignment to filter out partitions that are supposed to move to the target instance that
        // they are already on to avoid unnecessary partition movements. That is, if partition1 is currently on
        // instance1 and is supposed to move to instance1, remove it from the targetAssignment as this should be a
        // no-op. Accordingly the allToReassignPartitions should also be updated to remove such partitions.
        //Partitions not suppose to move
        Set<String> partitionsRemovedFromTargetAssignment = new HashSet<>(targetAssignment.get(instance));
        partitionsRemovedFromTargetAssignment.retainAll(partitionsAcrossAllDatastreamTasks);

        Set<String> updatedTargetPartitionList = new HashSet<>(targetAssignment.get(instance));
        updatedTargetPartitionList.removeAll(partitionsRemovedFromTargetAssignment);
        // filter invalid partitions
        updatedTargetPartitionList.retainAll(allToReassignPartitions);

        processedTargetAssignment.computeIfAbsent(instance,
            val -> updatedTargetPartitionList.isEmpty() ? null : updatedTargetPartitionList);

        allToReassignPartitions.removeAll(partitionsRemovedFromTargetAssignment);
      }`

In this, I have reordered your calls,
a. first find the partitions  from the target, which are not moving. 
b. Remove these partitions from the targetAssignment and retain  only the valid partitions in allToReassignPartitions (missing in your code) and save it in updatedTargetPartitionList
c. Remove these partitions from allToReassignPartitions

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -154,11 +154,13 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
   /**
    * Move a partition for a datastream group according to the targetAssignment. As we are only allowed to mutate the
    * task once. It follow the steps
-   * Step 1) get the partitions that to be moved, and find out their source task
-   * Step 2) If the instance is the one we want to move, we find out the task which we should assign the partition
+   * Step 1) Process the targetAssignment to remove any partitions with no-op moves (partition currently assigned
+   *         to the same instance where the partition is to be moved)
+   * Step 2) Get the partitions that are to be moved, and find their source tasks
+   * Step 3) If the instance is the one we want to move, we choose a task to which we should assign the partition
```
vmaheshw, line 160: Also, can you please add the steps in the comments in the code, so it is clear which code is which step?

somandal: done

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
somandal: done.

> b. Remove these partitions from the targetAssignment and retain  only the valid partitions in allToReassignPartitions (missing in your code) and save it in updatedTargetPartitionList

That step isn't really necessary (though I added it) since allToReassignPartitions has already filtered out invalid partitions, and all the other data-structures are created based on this.

jzakaryan (approved):

vmaheshw (approved):
LGTM.

PR #776, opened 2020-11-05
Add flushless config to EventProducer to allow disabling periodic flush on send

somandal:
BMM has a flushless mode, where the connector keeps track of all acknowledged sends for every TopicPartition and commits offsets based on these. The EventProducer tries to flush periodically on send, and this lands up calling flush even in flushless mode. This PR adds the capability to configure the EventProducer as flushless.

the contribution guidelines.

### DIFF
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamEventProducer.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamEventProducer.java
@@ -39,4 +39,12 @@ public interface DatastreamEventProducer {
    * As such, it is not efficient to be invoked very frequently.
    */
   void flush();
+
+  /**
+   * Enable periodic flush on send
+   *
+   * @param enableFlushOnSend Whether to enable flushing on send or not
+   */
+  default void flushOnSend(boolean enableFlushOnSend) {
+  }
 }
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamEventProducer.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamEventProducer.java
@@ -45,6 +45,6 @@ public interface DatastreamEventProducer {
    *
    * @param enableFlushOnSend Whether to enable flushing on send or not
    */
-  default void flushOnSend(boolean enableFlushOnSend) {
+  default void enablePeriodicFlushOnSend(boolean enableFlushOnSend) {
   }
 }
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/FlushlessEventProducerHandler.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/FlushlessEventProducerHandler.java
@@ -44,6 +44,7 @@ public class FlushlessEventProducerHandler<T extends Comparable<T>> {
    */
   public FlushlessEventProducerHandler(DatastreamEventProducer eventProducer) {
     _eventProducer = eventProducer;
+    _eventProducer.flushOnSend(false);
   }
 
   /**
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/FlushlessEventProducerHandler.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/FlushlessEventProducerHandler.java
@@ -44,7 +44,7 @@ public class FlushlessEventProducerHandler<T extends Comparable<T>> {
    */
   public FlushlessEventProducerHandler(DatastreamEventProducer eventProducer) {
     _eventProducer = eventProducer;
-    _eventProducer.flushOnSend(false);
+    _eventProducer.enablePeriodicFlushOnSend(false);
   }
 
   /**
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
@@ -49,6 +49,7 @@ public class EventProducer implements DatastreamEventProducer {
   public static final String CONFIG_FLUSH_INTERVAL_MS = "flushIntervalMs";
   public static final String CONFIG_ENABLE_PER_TOPIC_METRICS = "enablePerTopicMetrics";
   public static final String CONFIG_ENABLE_PER_TOPIC_EVENT_LATENCY_METRICS = "enablePerTopicEventLatencyMetrics";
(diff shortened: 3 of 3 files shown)


jzakaryan (approved):
LGTM

vmaheshw (commented):
Instead of this, can you define a method may be isFlushAllowed, which FlushlessEventProducer can override? Adding the FlushLessMode in the base class looks wrong from class Inheritence perspective.

somandal:
> Instead of this, can you define a method may be isFlushAllowed, which FlushlessEventProducer can override? Adding the FlushLessMode in the base class looks wrong from class Inheritence perspective.

Sure, wasn't too happy about this change, but unfortunately `FlushlessEventProducerHandler` does not extend `EventProducer` or the interface `DatastreamEventProducer`. The only way I can think of getting around this is to add a new send() API to `DatastreamEventProducer` which takes a flag to flush or not. E.g.:

`  /**
   * Send event onto the transport
   *
   * <p>
   * Note that the onComplete callbacks will generally execute in the I/O thread of the TransportProvider and
   * should be reasonably fast or they will delay sending messages for other threads.
   * If you want to execute an expensive callbacks it is recommended to use your own
   * {@link java.util.concurrent.Executor} in the callback body to parallelize processing.
   *
   * @param event event to send
   * @param callback call back that needs to called when the send completes.
   * @param possiblyFlush flag denoting whether to possibly flush or not
   */
  void send(DatastreamProducerRecord event, SendCallback callback, boolean possiblyFlush);`

Problem is that nowhere in the interface is there a contract that says that send() flushes or can flush. So the new method added to the abstract class is somewhat odd. I can make this change, and have `FlushlessEventProducerHandler` call the new send() method and pass `possiblyFlush` as false. Let me know if y

vmaheshw:
> > Instead of this, can you define a method may be isFlushAllowed, which FlushlessEventProducer can override? Adding the FlushLessMode in the base class looks wrong from class Inheritence perspective.
> 
> Sure, wasn't too happy about this change, but unfortunately `FlushlessEventProducerHandler` does not extend `EventProducer` or the interface `DatastreamEventProducer`. The only way I can think of getting around this is to add a new send() API to `DatastreamEventProducer` which takes a flag to flush or not. E.g.:
> 
> ```
>   /**
>    * Send event onto the transport
>    *
>    * <p>
>    * Note that the onComplete callbacks will generally execute in the I/O thread of the TransportProvider and
>    * should be reasonably fast or they will delay sending messages for other threads.
>    * If you want to execute an expensive callbacks it is recommended to use your own
>    * {@link java.util.concurrent.Executor} in the callback body to parallelize processing.
>    *
>    * @param event event to send
>    * @param callback call back that needs to called when the send completes.
>    * @param possiblyFlush flag denoting whether to possibly flush or not
>    */
>   void send(DatastreamProducerRecord event, SendCallback callback, boolean possiblyFlush);
> ```
> 
> Problem is that nowhere in the interface is there a contract that says that send() flushes or can flush. So the new method added to the abstract class is somewhat odd. I can make this change, and have `FlushlessEventProducerHandler` call the new send() method and pass `possiblyFlush` as false. Let me know if you think this approach is appropriate.

How about introducing a variable flushAllowed set default to true method in EventProducer which set flushAllowed to false and FlushlessEventProducer can call the method after instantiating EventProducer object. I would like to avoid changing the send interface itself.

somandal:
> > > Instead of this, can you define a method may be isFlushAllowed, which FlushlessEventProducer can override? Adding the FlushLessMode in the base class looks wrong from class Inheritence perspective.
> > 
> > 
> > Sure, wasn't too happy about this change, but unfortunately `FlushlessEventProducerHandler` does not extend `EventProducer` or the interface `DatastreamEventProducer`. The only way I can think of getting around this is to add a new send() API to `DatastreamEventProducer` which takes a flag to flush or not. E.g.:
> > ```
> >   /**
> >    * Send event onto the transport
> >    *
> >    * <p>
> >    * Note that the onComplete callbacks will generally execute in the I/O thread of the TransportProvider and
> >    * should be reasonably fast or they will delay sending messages for other threads.
> >    * If you want to execute an expensive callbacks it is recommended to use your own
> >    * {@link java.util.concurrent.Executor} in the callback body to parallelize processing.
> >    *
> >    * @param event event to send
> >    * @param callback call back that needs to called when the send completes.
> >    * @param possiblyFlush flag denoting whether to possibly flush or not
> >    */
> >   void send(DatastreamProducerRecord event, SendCallback callback, boolean possiblyFlush);
> > ```
> > 
> > 
> > Problem is that nowhere in the interface is there a contract that says that send() flushes or can flush. So the new method added to the abstract class is somewhat odd. I can make this change, and have `FlushlessEventProducerHandler` call the new send() method and pass `possiblyFlush` as false. Let me know if you think this approach is appropriate.
> 
> How about introducing a variable flushAllowed set default to true method in EventProducer which set flushAllowed to false and FlushlessEventProducer can call the method after instantiating EventProducer object. I would like to avoid changing the send interface itself.

I've updated the `DatastreamEventProducer` interface with a new method to set this variable. Let me know what you think. Since the AbstractKafkaBasedConnectorTask stores the producer as the generic type, i.e. `DatastreamEventProducer`, the function had to be added at this level, rather than just in the `EventProducer`. Wanted to avoid ugly typecasting, since it means that we cannot change the type.

jzakaryan (dismissed):

datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamEventProducer.java
```diff
@@ -39,4 +39,12 @@
    * As such, it is not efficient to be invoked very frequently.
    */
   void flush();
+
+  /**
+   * Enable periodic flush on send
+   *
+   * @param enableFlushOnSend Whether to enable flushing on send or not
+   */
+  default void flushOnSend(boolean enableFlushOnSend) {
```
vmaheshw, line 48: Can you change it to enablePeriodicFlushOnSend ?

vmaheshw, line 48: Can you change it to enablePeriodicFlushOnSend as method name?

vmaheshw (dismissed):

datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamEventProducer.java
somandal: done

jzakaryan (approved):

vmaheshw (approved):
LGTM

PR #778, opened 2020-11-06
Fix discrepancy in Gauge Metrics in PartitionMetrics and KafkaBasedConnectorTaskMetrics

vmaheshw:
Gauge metrics use a supplier function to get the latest value. For aggregate gauge metrics, we use a static object to have only one supplier function across all the tasks in the host. For task level gauge metrics, we maintain a local supplier function within the task object. This works fine as long as the key is different for each task. For KafkaMirrorMakerConnectorTask object, the metrics key is datastream name and not task name, which means if there are multiple tasks of the same datastream group on the same host, they will have conflict and gauge will return incorrect/unpredictable result. 

To solve this problem, making the gauge level supplier as a static object, so that there will be only one copy of the supplier function per key and we will not see the discrepancy in the result. 

I was able to reproduce the problem by extending the unit-tests. The unit-tests pass after the fix. This problem also exists for PartitionMetrics which has 2 gauge metrics "numPartitions" and "numStuckPartitions".


datastream-common/src/main/java/com/linkedin/datastream/connectors/CommonConnectorMetrics.java
```diff
@@ -186,23 +187,30 @@ protected void deregisterAggregates() {
...
+    final String _fullMetricsKey;
+
     // Map from connector class name to its stuck and total partition counter
     // This is needed for Gauge metrics which need long-typed suppliers.
+    static final Map<String, AtomicLong> NUM_STUCK_PARTITIONS_PER_METRIC_KEY = new ConcurrentHashMap<>();
+    static final Map<String, AtomicLong> NUM_PARTITIONS_PER_METRIC_KEY = new ConcurrentHashMap<>();
     static final Map<String, AtomicLong> AGGREGATED_NUM_STUCK_PARTITIONS = new ConcurrentHashMap<>();
     static final Map<String, AtomicLong> AGGREGATED_NUM_PARTITIONS = new ConcurrentHashMap<>();
+    private long _numStuckPartitions = 0;
+    private long _numPartitions = 0;
```
somandal, line 204: Once concern I have here is what if two threads update these metrics at the same time? Since these aren't atomic anymore, will that be something to worry about? Same for the numTopics variable for the numTopics metric.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
```diff
@@ -54,28 +55,48 @@
...
+  private static final Map<String, AtomicLong> NUM_AUTO_PAUSED_PARTITIONS_ON_INFLIGHT_MESSAGES_PER_METRIC_KEY =
+      new ConcurrentHashMap<>();
+  private static final Map<String, AtomicLong> NUM_AUTO_PAUSED_PARTITIONS_WAITING_FOR_DEST_TOPIC_PER_METRIC_KEY =
+      new ConcurrentHashMap<>();
+
+  private long _numConfigPausedPartitions = 0;
+  private long _numAutoPausedPartitionsOnError = 0;
+  private long _numAutoPausedPartitionsOnInFlightMessages = 0;
+  private long _numAutoPausedPartitionsAwaitingDestTopic = 0;
+  private long _numTopics = 0;
```
somandal, line 71: Same concern about these no longer being atomic, and what happens if multiple threads try to update any of these at the same time? Can't they race and cause incorrect updates?

datastream-common/src/main/java/com/linkedin/datastream/connectors/CommonConnectorMetrics.java
vmaheshw: changed it to AtomicLong.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
vmaheshw: changed it to AtomicLong.

somandal (approved):
lgtm, thanks for taking care of this

jzakaryan (approved):
lgtm

PR #781, opened 2020-11-13
Pull in some Kafka consumer metrics for the AbstractKafkaBasedConnectorTask

somandal:
This PR pulls in the following Kafka consumer metrics:

- consumer-offset-watermark-span
- consumer-liclosest-data-loss-estimation

These metrics are not present in JMX, so they need to be manually pulled in via the consumer.metrics() API.
This PR also bumps up the LiKafkaVersion to ensure that the above metrics are present.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
@@ -7,11 +7,18 @@ package com.linkedin.datastream.connectors.kafka;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
 
+import org.apache.commons.lang3.StringUtils;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
 import org.slf4j.Logger;
 
 import com.codahale.metrics.Histogram;
@@ -45,6 +52,10 @@ public class KafkaBasedConnectorTaskMetrics extends CommonConnectorMetrics {
   public static final String TIME_SPENT_BETWEEN_POLLS_MS = "timeSpentBetweenPollsMs";
   // keeps track of process + send time per event returned from poll() in nanoseconds
   public static final String PER_EVENT_PROCESSING_TIME_NANOS = "perEventProcessingTimeNs";
+  // keeps track of the sum of how far the safe offset is behind the watermark for all TopicPartitions
+  public static final String CONSUMER_OFFSET_WATERMARK_SPAN = "consumerOffsetWatermarkSpan";
+  // keeps track of how many times the consumer had OffsetOutOfRangeException or NoOffsetForPartitionException
+  public static final String CONSUMER_LICLOSEST_DATA_LOSS_ESTIMATION = "consumerLiclosestDataLossEstimation";
 
   private static final Map<String, AtomicLong> AGGREGATED_NUM_TOPICS = new ConcurrentHashMap<>();
   private static final Map<String, AtomicLong> AGGREGATED_NUM_CONFIG_PAUSED_PARTITIONS = new ConcurrentHashMap<>();
@@ -74,6 +85,8 @@ public class KafkaBasedConnectorTaskMetrics extends CommonConnectorMetrics {
   private final Histogram _timeSpentBetweenPollsMsMetric;
   private final Histogram _perEventProcessingTimeNanosMetric;
   private final String _fullMetricsKey;
+
+  private boolean _kafkaConsumerMetricsRegistered = false;
 
   KafkaBasedConnectorTaskMetrics(String className, String metricsKey, Logger errorLogger,
       boolean enableAdditionalMetrics) {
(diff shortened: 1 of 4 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
```diff
@@ -151,6 +201,11 @@ public void deregisterMetrics() {
       DYNAMIC_METRICS_MANAGER.unregisterMetric(_className, _key, TIME_SPENT_BETWEEN_POLLS_MS);
       DYNAMIC_METRICS_MANAGER.unregisterMetric(_className, _key, PER_EVENT_PROCESSING_TIME_NANOS);
     }
+
+    if (_kafkaConsumerMetricsRegistered) {
```
ahmedahamid, line 205: I don't have a strong opinion on the matter but `unregisterMetric` no-ops if the metrics hasn't been registered. This may spare you having to maintain and check `_kafkaConsumerMetricsRegistered`.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
```diff
@@ -130,6 +143,43 @@
...
+
+    Supplier<Double> offsetResetSupplier = () -> getConsumerLiclosestDataLossEstimationMetric(consumer, clientId);
+    DYNAMIC_METRICS_MANAGER.registerGauge(_className, _key, CONSUMER_LICLOSEST_DATA_LOSS_ESTIMATION, offsetResetSupplier);
+
+    _kafkaConsumerMetricsRegistered = true;
+  }
+
+  private double getConsumerOffsetWatermarkSpanMetric(Consumer<?, ?> consumer, String clientId) {
+    return getConsumerMetricValue(consumer, clientId, "consumer-offset-watermark-span");
+
```
ahmedahamid, line 169: nit: remove empty line

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
```diff
@@ -130,6 +143,43 @@
...
   }
 
+  /**
+   * Register some of the Kafka consumer metrics of interest. This cannot be done as part of the constructor
+   * as the Kafka consumer may be created at a later time than this object.
+   * @param consumer the Kafka consumer for which to register the metrics
+   * @param clientId the Kafka consumer's client.id
+   */
+  public void registerKafkaConsumerMetrics(Consumer<?, ?> consumer, String clientId) {
+    if (consumer == null || StringUtils.isBlank(clientId)) {
```
ahmedahamid, line 153: Why not throw on null consumer (and possibly `clientId` as well)?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
```diff
@@ -130,6 +143,43 @@
...
+
+  private double getConsumerLiclosestDataLossEstimationMetric(Consumer<?, ?> consumer, String clientId) {
+    return getConsumerMetricValue(consumer, clientId, "consumer-liclosest-data-loss-estimation");
+  }
+
+  private double getConsumerMetricValue(Consumer<?, ?> consumer, String clientId, String metricName) {
+    Map<String, String> tags = new HashMap<>(2);
+    tags.put("client-id", clientId);
+    MetricName name = new MetricName(metricName, "lnkd", "", tags);
+    return Optional.ofNullable(consumer).map(c -> c.metrics().get(name)).map(Metric::value).orElse(0.0);
```
ahmedahamid, line 180: I don't think `consumer` can be null since you already guard against that prior to calling `getConsumerMetricValue()`. The suppliers/lambdas are bound to the `consumer` param.

ahmedahamid, line 180: Are you deliberately using the deprecated `Metric::value` instead of `Metric.metricValue()` + casting?

ahmedahamid (approved):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
somandal: I've removed the consumer check, as if the consumer is null, we _should_ throw. I do believe that clientId isn't a mandatory consumer config, due to which I'd like to ensure that we don't throw an exception if this is blank, so keeping this one around.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
somandal: oops, done

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
somandal: Yeah good point. Have refactored this bit.

somandal: As discussed offline, my usage of the deprecated `Metric::value` was indeed deliberate. For using `Metric.metricValue()` + casting, you need to know the metric type, and unfortunately if Kafka changes the metric type underneath us, it can cause us to not emit the metric and there is no easy way to know why. On the other hand, if the deprecated method is removed or if the behavior is buggy, we have a problem too. Decided to go with the `Metric.metricValue()` + casting

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
somandal: Removed the extra check and removed the boolean

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
ahmedahamid: Thank you

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
ahmedahamid: Makes sense. Feel free to add a validate.notnull for `consumer`, too.

ahmedahamid (approved):

jzakaryan (approved):
LGTM

PR #782, opened 2020-11-18
Fix StickyMulticastStrategy to not assign more tasks than expected.

vmaheshw:
StickyMulticastStrategy copies the tasks from the existing active instances and does not check if the assignment is more than the expected. This results in more tasks assigned for the datastream than expected. This results in validation failure and Coordinator keeps on retrying LeaderDoAssignment. 

Fixing the problem by only reusing the number of tasks that are expected.


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyMulticastStrategy.java
```diff
@@ -128,6 +128,9 @@ public StickyMulticastStrategy(Optional<Integer> maxTasks, Optional<Integer> imb
         allAliveTasks.addAll(foundDatastreamTasks);
 
         if (!foundDatastreamTasks.isEmpty()) {
+          if (foundDatastreamTasks.size() > numTasks) {
```
jzakaryan, line 131: Nit: this check is not really necessary here. `limit()` will work on `foundDatastreamTasks` with `size <= numTasks`

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
```diff
@@ -475,6 +475,31 @@ public void testTriggerRebalanceWhenDeletingDatastream() {
         newAssignment.get(instancesBySize.get(instances.size() - 1)).size());
   }
 
+  @Test
+  public void testExtraTasksAreNotAssignedDuringReassignment() {
+    String[] instances = new String[]{"instance1"};
```
jzakaryan, line 480: Nit: spacing missing from line.

jzakaryan (approved):

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
vmaheshw: sorry, not able to figure out where is space missing?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyMulticastStrategy.java
vmaheshw: Keeping this check, so that we don't stream() and limit() every time and do it when necessary. Also, added a log on Sonam's recommendation, so that we can debug if something goes wrong in partition managed.

somandal (approved):
lgtm!

datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
jzakaryan: After the opening curly brace and before the closing curly brace.

jzakaryan (approved):

PR #783, opened 2020-12-09
Fix kafka connector shutdown path

vmaheshw:
If the task is in the shutdown path and an exception is encountered during the shutdown path, there is no need to rewind the checkpoint and pause, instead the entire processing can be aborted and the consumer can be closed. This way, if the task is getting killed because of session expiry, we will not be unnecessarily processing the entire batch. If the session expiry is because of network exception, the kafka calls do not go through resulting in the task blocking the cpu. 

Also cleaning some logs as they produce really large log messages in partition-managed datastream.


somandal (approved):
lgtm
Thanks for the logging improvements

jzakaryan (approved):
LGTM

PR #784, opened 2020-12-09
Handle ZkNoNodeException exception thrown by ensureReadData zkclient

vmaheshw:
Handle the ZkNoNodeException to ensure that we handle the nodes deleted around the time we query to get the data of the existing node. Ideally, we can skip these nodes as they got deleted.


somandal (approved):

jzakaryan (approved):
LGTM

PR #785, opened 2020-12-10
Fix the ConcurrentModificationException on the _trackEventsOutsideAltSlaMap in the EventProducer

somandal:
On enabling the additional warning log to track TopicPartitions seeing events outside alternate SLA, we saw an instance of STALLED partitions. Further investigation showed that the logging function in the EventProducer was throwing a ConcurrentModificationException under certain circumstances. When this occurs, the caller's onCompletion send callback isn't called. This can cause an issue where we never acknowledge the successful send, and the BMM flushless producer never commits a newer offset.

This PR fixes this in two ways:
- Make the _trackEventsOutsideAltSlaMap a ConcurrentHashMap
- Catch all exceptions that can be thrown by hashmap operations and log them rather than throwing.

the contribution guidelines.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
@@ -10,11 +10,12 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.ConcurrentModificationException;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.lang3.Validate;
@@ -108,7 +109,7 @@ public class EventProducer implements DatastreamEventProducer {
 
   private Instant _lastFlushTime = Instant.now();
   private long _lastEventsOutsideAltSlaLogTimeMs = System.currentTimeMillis();
-  private Map<TopicPartition, Integer> _trackEventsOutsideAltSlaMap = new HashMap<>();
+  private Map<TopicPartition, Integer> _trackEventsOutsideAltSlaMap = new ConcurrentHashMap<>();
   private boolean _enableFlushOnSend = true;
 
   /**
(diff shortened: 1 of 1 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -272,20 +273,28 @@ private void performSlaRelatedLogging(DatastreamRecordMetadata metadata, long ev
...
+        long timeSinceLastLog = System.currentTimeMillis() - _lastEventsOutsideAltSlaLogTimeMs;
+        if (timeSinceLastLog >= _numEventsOutsideAltSlaFrequencyMs) {
+          _trackEventsOutsideAltSlaMap.forEach((topicPartition, numEvents) -> _logger.warn(
+              "{} had {} event(s) with latency greater than alternate SLA of {} ms in the last {} ms for "
+                  + "datastream {}", topicPartition, numEvents, _availabilityThresholdAlternateSlaMs, timeSinceLastLog,
+              getDatastreamName()));
+          _trackEventsOutsideAltSlaMap.clear();
+          _lastEventsOutsideAltSlaLogTimeMs = System.currentTimeMillis();
+        }
+      } catch (ConcurrentModificationException | NullPointerException | IllegalArgumentException | ClassCastException
```
vmaheshw, line 292: ConcurrentHashMap will not throw ConcurrentModificationException. You can get rid of it.

vmaheshw (commented):
Do you think there is a value in moving sendCallback.onCompletion to finally block in `onSendCallback`, to avoid any future bug in `checkpoint()` or `reportMetrics` does not impact the checkpoint. The reason is this is a critical piece for FlushlessEventProducer otherwise we will be sending duplicate data, till the issue is identified and fixed.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
somandal: done

somandal:
> Do you think there is a value in moving sendCallback.onCompletion to finally block in `onSendCallback`, to avoid any future bug in `checkpoint()` or `reportMetrics` does not impact the checkpoint. The reason is this is a critical piece for FlushlessEventProducer otherwise we will be sending duplicate data, till the issue is identified and fixed.

As discussed offline, I've modified the code to catch all exceptions and wrap them in the sendFailureException and call the onCompletion in the finally block.

vmaheshw (approved):

jzakaryan (approved):

PR #786, opened 2020-12-14
Pre-commit and post-commit hooks in AbstractKafkaMirrorMakerConnectorTask

jzakaryan:
Sometimes it's necessary to perform pre-commit or post-commit operations in BMM connectors tasks. One could override/extend abstract method `maybeCommitOffsets()` for that purpose, but that method is called each time messages are polled and sent, and not always a call to `maybeCommitOffsets` results in offsets actually being committed.
The changes in this pull request introduce dedicated pre-commit and post-commit hooks that are called before and after a commit attempt. The default implementation of `preCommitHook()` is a no-op, while the default implementation of `postCommitHook()` logs an error if the commit failed and throws an exception. Please note that the latter behavior was previously part of the `commitWithRetries()` method.

Update: Post-commit hook has been removed after discussing with @vmaheshw and @somandal. There's no use case for it now and it can be added back when there is one.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -608,6 +608,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
 
   protected void commitWithRetries(Consumer<?, ?> consumer, Optional<Map<TopicPartition, OffsetAndMetadata>> offsets)
       throws DatastreamRuntimeException {
+    preCommitHook();
     boolean result = PollUtils.poll(() -> {
       try {
         if (offsets.isPresent()) {
@@ -629,12 +630,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
 
       return true;
     }, COMMIT_RETRY_INTERVAL_MILLIS, COMMIT_RETRY_TIMEOUT_MILLIS);
-
-    if (!result) {
-      String msg = "Commit failed after several retries, Giving up.";
-      _logger.error(msg);
-      throw new DatastreamRuntimeException(msg);
-    }
+    postCommitHook(result);
   }
 
   /**
@@ -771,6 +767,23 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
       } else {
         throw new IllegalStateException("Found null update type in task updates set.");
       }
+    }
+  }
+
+  /**
+   * Pre commit hook for all operations that need to be performed before committing offsets.
+   */
+  protected void preCommitHook() { }
+
+  /**
+   * Post commit hook for all operations that need to be performed after committing offsets.
+   * @param success Indicates whether the commit attempt was successful or not.
+   */
+  protected void postCommitHook(boolean success) {
+    if (!success) {
+      String msg = "Commit failed after several retries, Giving up.";
+      _logger.error(msg);
+      throw new DatastreamRuntimeException(msg);
     }
   }
 
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -634,8 +634,6 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
       return true;
     }, COMMIT_RETRY_INTERVAL_MILLIS, COMMIT_RETRY_TIMEOUT_MILLIS);
 
-    postCommitHook(result, lastKafkaExceptionRef.get());
-
     if (!result) {
       String msg = "Commit failed after several retries, Giving up.";
       _logger.error(msg);
@@ -782,19 +780,12 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
 
   /**
    * Pre commit hook for all operations that need to be performed before committing offsets.
+   * <p>
+   *   Note: An unhandled exception in pre-commit hook may result in an interrupted commit. Classes overriding the hook
+   *   are expected to handle exceptions.
+   * </p>
    */
   protected void preCommitHook() { }
-
-  /**
-   * Post commit hook for all operations that need to be performed after committing offsets.
-   * <p>
-   *     Note: A commit attempt can be declared successful even when there's an exception. This can happen when a commit
-   *     exception is thrown during task shutdown.
-   * </p>
-   * @param success Indicates whether the commit attempt was successful or not.
-   * @param exception Exception caught during a commit attempt.
-   */
-  protected void postCommitHook(boolean success, KafkaException exception) { }
 
   /**
    * The method, given a datastream task - checks if there is any update in the existing task.
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.Validate;
 import org.apache.commons.lang.exception.ExceptionUtils;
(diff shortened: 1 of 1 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -774,6 +770,23 @@ protected void preConsumerPollHook() {
...
+  /**
+   * Pre commit hook for all operations that need to be performed before committing offsets.
+   */
+  protected void preCommitHook() { }
+
+  /**
+   * Post commit hook for all operations that need to be performed after committing offsets.
+   * @param success Indicates whether the commit attempt was successful or not.
+   */
+  protected void postCommitHook(boolean success) {
```
abhishekmendhekar, line 782: It'll be good if the `postCommitHook()` can get hold of the exception causing the commit to fail.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -774,6 +770,23 @@ protected void preConsumerPollHook() {
...
+    if (!success) {
```
abhishekmendhekar, line 783: I think the `postCommitHook()` should not throw the exception because if the method is overridden the derived class and it fails to implement this logic then ab exception will never be thrown.
Instead the logic should remain the same as before and this operation can remain as a no-op.

abhishekmendhekar (commented):
Since I am not the an approver here just left my thoughts about the code. :)

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: +1

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: Since we have multiple retries for commit, caching  the last exception should be sufficient.

jzakaryan: I'm not sure if there's value in propagating the Kafka exception up except for logging purposes, and the code in `commitWithRetries` already logs the exception. Do you think that there's value in propagating the last exception, @somandal?

somandal: Right, at the moment I don't really fully see a need for the postCommitHook either, so i'm not yet clear about how the exception will help as such. I'll leave it up to you @jzakaryan to make a call on this.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -618,6 +621,7 @@ protected void commitWithRetries(Consumer<?, ?> consumer, Optional<Map<TopicPart
         }
         _logger.info("Commit succeeded.");
       } catch (KafkaException e) {
+        lastKafkaExceptionRef.set(e);
```
somandal, line 624: I think there are many other exceptions that can be thrown here, other than KafkaException. (check documentation for commitSync())

I'd suggest adding a catch block for those exceptions too, where you just cache the exception.

jzakaryan: According to API reference, `commitSync` indeed throws a variety of exceptions other than `KafkaException`. However, I think there's a reason these exceptions are not handled here and are left to propagate up the call stack and I don't think there's a need to trigger the post-commit hook with the cached exception in this case.

somandal (approved):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
abhishekmendhekar: Assuming there is a set of actions that follow post commit using this hook.
true/false will give a signal of commit succeed or failed but an exception can help understand why it failed. Are you saying that there are more than 1 exception thrown? Generally Kafka should send only one commit message for a given consumer group and if this method commits more than 1 CG then I suggest returning a `Map<String, Exception>`. This gives user the most flexibility in terms of using this API.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -618,6 +621,7 @@ protected void commitWithRetries(Consumer<?, ?> consumer, Optional<Map<TopicPart
...
         if (_shutdown) {
           _logger.info("Caught KafkaException in commitWithRetries while shutting down, so exiting.", e);
           return true;
```
vmaheshw, line 627: During shutdown, if it hits KafkaException, it returns true, which does not really mean success. This can impact the postCommit hook.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -774,6 +780,22 @@ protected void preConsumerPollHook() {
...
+  /**
+   * Post commit hook for all operations that need to be performed after committing offsets.
+   * <p>
+   *     Note: A commit attempt can be declared successful even when there's an exception. This can happen when a commit
+   *     exception is thrown during task shutdown.
+   * </p>
+   * @param success Indicates whether the commit attempt was successful or not.
+   * @param exception Exception caught during a commit attempt.
+   */
+  protected void postCommitHook(boolean success, KafkaException exception) { }
```
vmaheshw, line 797: Since, we don't have a concrete usage for this hook in open source and we currently have identified only the need for preCommit hook, will it be okay to postpone it, till either we identify the use for it or we get any pull/feature request. This will help us in avoiding an unnecessary tech debt to keep on maintaining.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -608,6 +609,8 @@ protected void maybeCommitOffsetsInternal(Consumer<?, ?> consumer, boolean force
 
   protected void commitWithRetries(Consumer<?, ?> consumer, Optional<Map<TopicPartition, OffsetAndMetadata>> offsets)
       throws DatastreamRuntimeException {
+    preCommitHook();
```
vmaheshw, line 612: If preCommitHook() threw exception, the commit will not go through. Your current use-case to have audit related drain in the preCommitHook() also adds a dependency that if there is any failure in audit, the commits will not go through. Is this really expected ? Is it possible to have the audit system throwing errors while the regular producers are working fine.

vmaheshw (commented):
Please evaluate

a)  if you really want to introduce the postCommitHook at this stage, where we don't have a clear usecase for it.
b) if you want to tie the audit failures to block the commits?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
jzakaryan: This is no longer relevant now that post-commit hook is removed.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
jzakaryan: I made it explicit in javadoc that the overriding class needs to take care of handling exceptions from pre-commit hook.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -608,6 +609,8 @@ protected void maybeCommitOffsetsInternal(Consumer<?, ?> consumer, boolean force
...
+    AtomicReference<KafkaException> lastKafkaExceptionRef = new AtomicReference<>(null);
```
somandal, line 613: Can we remove this now that we no longer need to pass it to the postCommitHook()?

jzakaryan: Done

vmaheshw (approved):

somandal (approved):
Lgtm!

Issue #788, opened 2020-12-17
Start Offsets not taking effect in KafkaConnectorTask


shenodaguirguis:
### Subject of the issue
KafkaConnectorTask extends AbstractKafkaBasedConnectorTask, which enables users to set start offsets, to set its Kafka consumer to start from these offsets. However,  these start offsets takes effect only after an initial poll failure throwing NoOffsetForPartitionException. In particular, the AbstractKafkaBasedConnectorTask.pollRecords() calls handleNoOffsetForPartitionException() which in turns seeks to the set start offsets, if present. The intuition is that a new consumer starts with no checkpoints and the very first throw will throw the NoOffsetForPartitionException. However, Kafka consumer has a config [auto.offset.reset](https://kafka.apache.org/documentation/#consumerconfigs) which when set to '**earliest**' or '**latest**', will automatically seek to beginning or end of the topic, without throwing exceptions. This config has a default value of earliest. Therefore, if the auto.offset.reset config is omitted to use its default, or if it is set to a value other than '**none**', no exception will be thrown and the start offsets won't be used.

### Expected behaviour
Brooklin's KafkaConnector datastream should start consumption for the set start offsets when set.

### Actual behaviour
Start offsets are ignored.

PR #789, opened 2020-12-17
[issue 788] make start offsets take effect when set in KafkaConnectorTask

shenodaguirguis:
Please see https://github.com/linkedin/brooklin/issues/788 for a description of the problem
this changes addresses this issue by overriding the auto.offset.reset kafka consumer config and sets it to none whenever start offsets are set in the AbstractKafkaBasedConnectorTask, so that the set start offsets takes effect while handling the NoCheckpointForPartitoinException upon first poll (given that auto.offset.reset=none)
Modified unit test to verify the override, also validated from the logs of the run that added overriding logs are emitted.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -74,6 +74,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
 
   public static final String CONSUMER_AUTO_OFFSET_RESET_CONFIG_LATEST = "latest";
   public static final String CONSUMER_AUTO_OFFSET_RESET_CONFIG_EARLIEST = "earliest";
+  public static final String CONSUMER_AUTO_OFFSET_RESET_CONFIG_NONE = "none";
 
   protected long _lastCommittedTime = System.currentTimeMillis();
   protected int _eventsProcessedCount = 0;
@@ -104,7 +105,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected final Duration _pauseErrorPartitionDuration;
   protected final long _processingDelayLogThresholdMillis;
   protected final boolean _enableAdditionalMetrics;
-  protected final Optional<Map<Integer, Long>> _startOffsets;
+  protected final Map<Integer, Long> _startOffsets;
 
   protected volatile String _taskName;
   protected final DatastreamEventProducer _producer;
@@ -147,12 +148,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
     _taskName = task.getDatastreamTaskName();
 
     _consumerProps = config.getConsumerProps();
-    if (_datastream.getMetadata().containsKey(KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY)) {
-      String strategy = _datastream.getMetadata().get(KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY);
-      _logger.info("Datastream contains consumer config override for {} with value {}",
-          ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, strategy);
-      _consumerProps.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, strategy);
-    }
+
     if (Boolean.TRUE.toString()
         .equals(_datastream.getMetadata().get(KafkaDatastreamMetadataConstants.USE_PASSTHROUGH_COMPRESSION))) {
       _consumerProps.put("enable.shallow.iterator", Boolean.TRUE.toString());
@@ -165,9 +161,38 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
     _pausePartitionOnError = config.getPausePartitionOnError();
     _pauseErrorPartitionDuration = config.getPauseErrorPartitionDuration();
     _enableAdditionalMetrics = config.getEnableAdditionalMetrics();
-    _startOffsets = Optional.ofNullable(_datastream.getMetadata().get(DatastreamMetadataConstants.START_POSITION))
-        .map(json -> JsonUtils.fromJson(json, new TypeReference<Map<Integer, Long>>() {
-        }));
+
+    _startOffsets = new HashMap<>();
+    String json = _datastream.getMetadata().get(DatastreamMetadataConstants.START_POSITION);
+    if (StringUtils.isNotBlank(json)) {
+      _startOffsets.putAll(JsonUtils.fromJson(json, new TypeReference<Map<Integer, Long>>() {
+      }));
+    }
+
+    // if this datastream wants to use specific start offsets, we need to have the auto.offset.reset config
+    // set to "none" in the kafka consumer configs, so that the Kafka consumer throws NoOffsetForPartitionException
+    // upon first poll, to be handled by seeking to the startOffsets
+    if (!_startOffsets.isEmpty()) {
+      if (_datastream.getMetadata().containsKey(KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY)) {
+        _logger.info("Datastream contains metadata for both {} and {}, which are mutually exclusive. Ignoring the "
+                + "{}=\"{}\" metadata, as start offsets require it to be set to \"none\"",
+            DatastreamMetadataConstants.START_POSITION,
+            KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY,
+            KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY,
+            _datastream.getMetadata().get(KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY));
+      }
+      _logger.info("Datastream contains startOffsets, setting {} = \"{}\" in consumer configs "
+              + "(overriding the configs value of \"{}\")",
+          ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_NONE,
+          getConsumerAutoOffsetResetConfig());
+      _consumerProps.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_NONE);
+    } else if (_datastream.getMetadata().containsKey(KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY)) {
+        String strategy = _datastream.getMetadata().get(KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY);
+        _logger.info("Datastream contains consumer config override for {} with value {}",
+            ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, strategy);
+        _consumerProps.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, strategy);
+    }
+
     _offsetCommitInterval = config.getCommitIntervalMillis();
     _pollTimeoutMillis = config.getPollTimeoutMillis();
     _retrySleepDuration = config.getRetrySleepDuration();
@@ -675,9 +700,9 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   private void seekToStartPosition(Consumer<?, ?> consumer, Set<TopicPartition> partitions, String strategy) {
     // We would like to consume from latest when there is no offset. However, if we rewind due to exception, we want
     // to rewind to earliest to make sure the messages which are added after we start consuming don't get skipped
-    if (_startOffsets.isPresent()) {
-      _logger.info("Datastream is configured with StartPosition. Trying to start from {}", _startOffsets.get());
-      seekToOffset(consumer, partitions, _startOffsets.get());
+    if (!_startOffsets.isEmpty()) {
+      _logger.info("Datastream is configured with StartPosition. Trying to start from {}", _startOffsets);
+      seekToOffset(consumer, partitions, _startOffsets);
     } else {
       if (CONSUMER_AUTO_OFFSET_RESET_CONFIG_LATEST.equals(strategy)) {
         _logger.info("Datastream was not configured with StartPosition. Seeking to end for partitions: {}", partitions);
@@ -1034,5 +1059,10 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
       consumerMetrics.updateErrorRate(1, "Can't find group ID", e);
       throw e;
     }
+  }
+
+  @VisibleForTesting
+  public String getConsumerAutoOffsetResetConfig() {
+    return _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "");
   }
 }
--- a/datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
+++ b/datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
@@ -47,6 +47,7 @@ import com.linkedin.datastream.common.DatastreamRuntimeException;
 import com.linkedin.datastream.common.DatastreamSource;
 import com.linkedin.datastream.common.JsonUtils;
 import com.linkedin.datastream.common.PollUtils;
+import com.linkedin.datastream.kafka.KafkaDatastreamMetadataConstants;
 import com.linkedin.datastream.kafka.factory.KafkaConsumerFactory;
 import com.linkedin.datastream.kafka.factory.KafkaConsumerFactoryImpl;
 import com.linkedin.datastream.server.DatastreamEventProducer;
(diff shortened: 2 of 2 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -168,6 +164,22 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
     _startOffsets = Optional.ofNullable(_datastream.getMetadata().get(DatastreamMetadataConstants.START_POSITION))
         .map(json -> JsonUtils.fromJson(json, new TypeReference<Map<Integer, Long>>() {
         }));
+
+    // if this datastream want to use specific start offsets, we need to have the auto.offset.reset config
```
DEEPTHIKORAT, line 168: nit: If this datastream *wants*

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -168,6 +164,22 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
...
+    // set to "none" in the kafka consumer configs, so that the Kafka consumer throws NoOffsetForPartiionException
+    // upon first poll, to be handled by seeking to teh startOffsets
```
DEEPTHIKORAT, line 170: nit: typo on the.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -168,6 +164,22 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
...
+    if (_startOffsets.isPresent()) {
```
DEEPTHIKORAT, line 171: unrelated to your change, but should we make _startOffsets not Optional? It is a map and we are wrapping it in a Nullable only to be using isPresent everywhere if present needing to access it with the get call which is not serving the purpose of Optional at all.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -168,6 +164,22 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
...
+      _logger.info("Datastream contains startOffsets, override {} with value {} (was {} in the provided configs)",
```
DEEPTHIKORAT, line 172: nit: I was originally conflicted about the change because both start position and reset strategy are overridable values and we are basically giving higher precedence to start position by adjusting the reset strategy. But on more thought I do feel like this is the right thing to do, given that reset strategy isn't something a client will provide when they have already specified the start position. Although, it would be confusing is if they provided both as it wouldn't be possible to figure out which one they really wanted to set, I do feel like start position somehow feels like a more granular thing to specify and so should probably take precedence. 
Can we however fetch offset reset strategy first so that in the logs it can be made clear that that was ignored because startPosition was also set?

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
```diff
@@ -172,6 +172,9 @@ public void testConsumeWithStartingOffset() throws Exception {
 
     KafkaConnectorTask connectorTask = createKafkaConnectorTask(task);
 
+    // validate auto.offset.reset config is overridden to none (given the start offsets)
+    Assert.assertEquals(connectorTask.getConsumerAutoOffsetResetConfig(), "none");
```
DEEPTHIKORAT, line 176: Can we add a specific test case to try the use-cases out? Many cases crop up since it is 2 configs, but may be we need at least the one you are fixing, with start position specified and no reset strategy.

DEEPTHIKORAT (changes requested):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -168,6 +164,22 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
     _startOffsets = Optional.ofNullable(_datastream.getMetadata().get(DatastreamMetadataConstants.START_POSITION))
         .map(json -> JsonUtils.fromJson(json, new TypeReference<Map<Integer, Long>>() {
         }));
+
+    // if this datastream want to use specific start offsets, we need to have the auto.offset.reset config
+    // set to "none" in the kafka consumer configs, so that the Kafka consumer throws NoOffsetForPartiionException
```
somandal, line 169: nit: NoOffsetForPartiionException -> NoOffsetForPartitionException

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -1035,4 +1047,9 @@ public static String getKafkaGroupId(DatastreamTask task, GroupIdConstructor gro
       throw e;
     }
   }
+
+  @VisibleForTesting
+  public String getConsumerAutoOffsetResetConfig() {
+    return _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "NOT SET");
```
somandal, line 1053: Can you return a null string instead of "NOT SET"?

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
somandal: +1, I'd go as far as to make sure that we see the expected behavior when start positions are present, and when they aren't.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
somandal: +1
We could perhaps add some kind of validation that start positions cannot be present if the auto.offset.reset strategy is set to anything other than "none" if we feel the need to be careful, but I'll leave the decision to add this or not to you.

DEEPTHIKORAT: I was thinking about validation, but then the stream task would just die, right? I wasn't terribly sure if this was a good idea since clients are not necessarily good at figuring out what's happening on the server, but may be in such cases they are, and may be SREs should in fact get alerted for bad configs.

somandal: @DEEPTHIKORAT what about checking this during a) initialize datastream and b) datastream update? This way both stream creation/update will fail. 
Again don't have a strong opinion on whether we should add this or not, but it's just a thought if we're worried about explaining the behavior when both metadata fields are present.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
nisargthakkar: Nit: Another typo one the exception name

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
shenodaguirguis: makes sense. I made _startOffsets not Optional

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
shenodaguirguis: I agree with Deepthi that failing might not be the right thing to do. And that was exactly my thinking @DEEPTHIKORAT  that when users set the start offsets, they mean to use it. I am also doing exactly as you suggest: I am logging the original value before the update. Here is a sample log from the test:

`2020-12-17 16:27:41 INFO  KafkaConnectorTask:176 - Datastream contains startOffsets, override auto.offset.reset with value none (was earliest in the provided configs)`

I will attempt to reword to make it clearer

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
shenodaguirguis: testConsumeBaseCase tests when no startOffsets are present, 
testConsumeWithTheStartingOffset tests the case with startOffsets are present. It was testing with no auto.offset.rest set, I changed it to set auto.offset.reset set to earliest, and it already verifies the expected behavior (in terms of number of records consumed)

The actual value of the `auto.offset.reset` is irrelevant, we only need to test if it is set that we do override. Whether it was `earliest`,  `latest` or `none`, is orthogonal.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -165,9 +161,30 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
...
+    _startOffsets = new HashMap<>();
+    String json = _datastream.getMetadata().get(DatastreamMetadataConstants.START_POSITION);
+    if (StringUtils.isNotBlank(json)) {
+      _startOffsets.putAll(JsonUtils.fromJson(json, new TypeReference<Map<Integer, Long>>() {
+      }));
+    }
+
+    // if this datastream wants to use specific start offsets, we need to have the auto.offset.reset config
+    // set to "none" in the kafka consumer configs, so that the Kafka consumer throws NoOffsetForPartitionException
+    // upon first poll, to be handled by seeking to teh startOffsets
```
somandal, line 174: nit:  teh -> the

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -165,9 +161,30 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
...
+    if (!_startOffsets.isEmpty()) {
+      _logger.info("Datastream contains startOffsets, setting {} = \"{}\" in consumer configs "
```
somandal, line 176: In case someone adds the datastream metadata too, should we print that here too? It'll be helpful in debugging if someone sets this to something else and is wondering why it isn't taking effect.

`String strategy = _datastream.getMetadata().get(KafkaDatastreamMetadataConstants.CONSUMER_OFFSET_RESET_STRATEGY);`

This log is only printing the config.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -165,9 +161,30 @@ protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, Data
...
+              + "(overriding the configs value of \"{}\")",
+          ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_NONE,
+          _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, ""));
```
somandal, line 179: can we modify `getConsumerAutoOffsetResetConfig` to set an empty string "" instead of null, and call that here? Then instead of null checks we can use isBlank checks.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
```diff
@@ -448,6 +452,13 @@ static Datastream getDatastream(String broker, String topic) {
     return datastream;
   }
 
+  private KafkaConnectorTask createKafkaConnectorTaskWithAutoOffsetResetConfig(DatastreamTaskImpl task,
+      String autoOffsetResetStrategy) throws InterruptedException {
+    Properties consumerProperties = new Properties();
+    consumerProperties.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, autoOffsetResetStrategy);
+    return createKafkaConnectorTask(task, new KafkaBasedConnectorConfigBuilder()
+        .setConsumerProps(consumerProperties).build());
+  }
```
somandal, line 461: nit: add an empty line after this

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
shenodaguirguis: I added a test case for when the auto offset reset strategy is set in metadata, to verify startOffsets takes precedence

somandal (approved):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
DEEPTHIKORAT: I think that is a nice idea @somandal. @shenodaguirguis feel free to check in this version since Sonam has approved it as well, but we can consider making the change Sonam suggested in the future perhaps. At least then it is a user-facing failure and they won't be caught unaware.

shenodaguirguis: we can have a separate change for this

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
```diff
@@ -170,7 +171,58 @@ public void testConsumeWithStartingOffset() throws Exception {
...
+  @Test
+  public void testConsumeWithStartingOffsetAndResetStrategy() throws Exception {
+    String topic = "pizza1";
+    createTopic(_zkUtils, topic);
+
+    LOG.info("Sending first set of events");
+
+    //produce 100 msgs to topic before start
+    produceEvents(_kafkaCluster, _zkUtils, topic, 0, 100);
+    Map<Integer, Long> startOffsets = Collections.singletonMap(0, 100L);
```
DEEPTHIKORAT, line 203: was the choice of 100L intentional for start timestamp?

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
```diff
@@ -170,7 +171,58 @@ public void testConsumeWithStartingOffset() throws Exception {
...
+    LOG.info("Sending second set of events");
+
+    //produce 100 msgs to topic before start
+    produceEvents(_kafkaCluster, _zkUtils, topic, 100, 100);
+
+    //start
+    MockDatastreamEventProducer datastreamProducer = new MockDatastreamEventProducer();
+    Datastream datastream = getDatastream(_broker, topic);
+
+    // Unable to set the start position, OffsetToTimestamp is returning null in the embedded Kafka cluster.
```
DEEPTHIKORAT, line 214: The name start position unfortunately is confusing since it is a timestamp. Since you have set it to 100 I am guessing it is returning null? And so what happens when it returns null? What is the code flow when this happens?

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
shenodaguirguis: yes intentional, and it is a start offset, not timestamp.
we produce 100 events, set startOffsets to 100, produce 200 more and make sure we consume 200 only to verify start offset metadata took effect

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestKafkaConnectorTask.java
shenodaguirguis: a) I copied and modfied `testConsumeWithStartOffset`, so not sure what this comment refers to, I removed it.
b) it is not a timestamp, it is an offset. The metadata key is: `DatastreamMetadataConstants.START_POSITION = "system.start.position"`

DEEPTHIKORAT: I see. So it does seem to be offset in the implementation and there are some other connectors storing time. I was guessing that is where the comment about offsetForTimestamp was coming from. Good to know it actually causes no issues.

DEEPTHIKORAT (approved):

somandal (approved):

shenodaguirguis:
merging changes

PR #790, opened 2021-01-07
Avoid sending the message using the kafka transport provider which is already unassigned.

vmaheshw:
Handling the callback failures in case of producer close() because of task shutdown to not pollute the logs. Moving logs to debug to be able to enable later for debugging. We can also summarize the log and print every 100 callback failure. Reducing the error logs printed multiple times in the path.


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -296,7 +296,6 @@ void send(DatastreamTask task, ProducerRecord<K, V> producerRecord, Callback onC
           Thread.sleep(_sendFailureRetryWaitTimeMs);
         }
       } catch (Exception e) {
-        _log.error(String.format("Send failed for partition %d with an exception", producerRecord.partition()), e);
```
somandal, line 299: Is there any other log for this code path that tells us the partition and exception? Just want to make sure we'll have sufficient logs to debug issues if this occurs.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -161,8 +173,13 @@ public void send(String destinationUri, DatastreamProducerRecord record, SendCal
         producer.send(_datastreamTask, outgoing, (metadata, exception) -> {
           int partition = metadata != null ? metadata.partition() : -1;
           if (exception != null) {
-            LOG.error("Sending a message with source checkpoint {} to topic {} partition {} for datastream task {} "
-                    + "threw an exception.", record.getCheckpoint(), topicName, partition, _datastreamTask, exception);
+            if (_isUnassigned) {
+              LOG.debug("Sending a message with source checkpoint {} to topic {} partition {} for datastream task {} "
```
somandal, line 177: The log string is the same, can you construct the string outside the if-else, and just use that variable to print?

somandal (approved):

jzakaryan (approved):
Agree with @somandal's comment on identical log message. Otherwise looks good.

PR #791, opened 2021-01-07
LiveInstanceProvider should be used only by the leader

vmaheshw:
Earlier, LiveInstaceProvider was used only by the leader. We added it to all the hosts to identify the orphan locks for partition managed datastream. This can cause lot of zk read call. The easier fix can be to just check the orphan node during release the task lock and do not depend on live instance provider, since it is a much cheaper and less frequent call.


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1434,6 +1434,13 @@ public static String parseHostnameFromZkInstance(String instance) {
     return instance.substring(0, instance.lastIndexOf('-'));
   }
 
+  /**
+   * Parse the Zk instance (ex. hostname-0000) into liveinstance
+   */
+  public static String parseLiveInstanceFromZkInstance(String instance) {
```
somandal, line 1440: nit: rename "String instance" to  "String owner"?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -1434,6 +1434,13 @@ public static String parseHostnameFromZkInstance(String instance) {
     return instance.substring(0, instance.lastIndexOf('-'));
   }
 
+  /**
+   * Parse the Zk instance (ex. hostname-0000) into liveinstance
```
somandal, line 1438: nit: reword
Parse the liveinstance ID from the owner string (format: hostname-<liveinstance id>, ex. hostname-0000)

somandal (approved):

jzakaryan (dismissed):
LGTM

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
vmaheshw: Keeping this method generic since it is zk instance, which is used in lock ownership as well.

somandal (approved):

jzakaryan (approved):

PR #793, opened 2021-01-15
Avoid sending the message using the kafka transport provider which is already unassigned.

vmaheshw:
Handling the callback failures in case of producer close() because of task shutdown to not pollute the logs. Moving logs to debug to be able to enable later for debugging. We can also summarize the log and print every 100 callback failure. Reducing the error logs printed multiple times in the path.


jzakaryan (dismissed):

jzakaryan (approved):

somandal (approved):
lgtm, thanks!

PR #796, opened 2021-01-26
[DISCARD] Elastic task assignment should get and store the numTasks in ZooKeeper

somandal:
Elastic task assignment elastically figures out the number of tasks needed for supporting a limited number of partitions per task. This PR stores this information in ZooKeeper for persistence across leader changes and across assignments.

There is no access to the ZkAdapter object from within the AssignmentStrategies, so this PR gets around this by adding APIs to get the cached numTasks from the AssignmentStrategy, fetch from ZK if this is not present, and update ZK if the expected numTasks changes. This also only triggers for datastreams where elastic task assignment is enabled.

The ZK node is created under: <cluster>/dms/<datastream name>/numTasks, where <datastream name> matches the datastream's TASK_PREFIX.

the contribution guidelines.

### DIFF
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/api/strategy/AssignmentStrategy.java
@@ -8,6 +8,7 @@ package com.linkedin.datastream.server.api.strategy;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import org.apache.commons.lang.NotImplementedException;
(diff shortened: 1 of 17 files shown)


vmaheshw (commented):
One way to achieve this can be to perform the necessary zookeeper operation inside the strategy code. To achieve that you need to:

a. Pass a zkClient object the constructor of the strategy i.e. StickyPartitionAssignmentStrategy.

b. To get the zkClient object in StickyPartitionAssignmentStrategyFactory: you need to set zkClient configs i.e. zkAddress, zksessionTimeout and zkConnectionTimeout at STRATEGY_DOMAIN level i.e. `brooklin.server.connector.PartitionManagedKafkaMirrorMaker.strategy.`

c. Once you have the zkClient object, you can write numTasks to the zookeeper directly. We can put strong comments around the usage of zkclient object.

This way the changes will be contained to StickyPartitionAssignmentStrategy and only if the elastic task feature is enabled, zkclient object will be passed and used.

Please let me know if you want to discuss this idea.

PR #802, opened 2021-02-09
Add the api to display the message acked count that are out-of-order and cannot be used in safe offsets

vmaheshw:
Add the api in `FlushlessEventProducer` to print the count of the messages that are acked out of order and cannot be used in `commitSafeOffsets`. This will show how many kafka messages are acked and cannot be committed as part of `commitSafeOffsets` since there are still some pending inFlight messages to be acked in order.


somandal (approved):

jzakaryan (approved):
There's some inconsistent/extra spacing in the test file, but other than that, LGTM.

PR #803, opened 2021-02-10
Avoid overproduce in kafka.

vmaheshw:
During Task unassign, we shutdown the producer with 2 second timeout to abort the pending send and shutdown the task fast. We commit the safe offsets to avoid duplicating. But, there is a possibility that the send was completed, but the producer closed before the ack could come back. 
a) So, we are increasing the producer close in task unassignment path to longer duration.
b) During send, if the task is unassigned, we will preemptively fail and call `mayBeCommitOffsets` that internally calls `flush` and then `commitSafeOffsets` in case of `FlushlessEventProducer`.
c) The timeout to wait to kill the task is also extended.
d) Zookeeper expiry transition is changed from 30 sec to 120 sec. Same applies to task lock to be marked as orphan.
e) Reducing sendcallback error logs.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -129,6 +129,7 @@
   private long _maxInFlightMessagesThreshold;
   private long _minInFlightMessagesThreshold;
   private int _flowControlTriggerCount = 0;
+  private int _errorOnSendCallbackDuringShutdownCnt = 0;
```
somandal, line 132: nit: rename: _errorOnSendCallbackDuringShutdownCnt -> _errorOnSendCallbackDuringShutdownCount

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -322,9 +334,10 @@ void shutdownProducer() {
       // quickly to avoid delaying/blocking the sender thread. Thus schedule the actual producer.close() on a separate
       // thread
       _producerCloseExecutorService.submit(() -> {
-        _log.info("KafkaProducerWrapper: Closing the Kafka Producer");
+        int timeout = fastClose ? FAST_CLOSE_TIMEOUT_MS : _producerCloseTimeoutMs;
+        _log.info("KafkaProducerWrapper: Closing the Kafka Producer timeout: {}", timeout);
```
somandal, line 338: nit: Closing the Kafka Producer with timeout

datastream-server-api/src/main/java/com/linkedin/datastream/server/FlushlessEventProducerHandler.java
```diff
@@ -70,7 +70,7 @@ public void send(DatastreamProducerRecord record, String source, int sourceParti
     status.register(sourceCheckpoint);
     _eventProducer.send(record, ((metadata, exception) -> {
       if (exception != null) {
-        LOG.error("Failed to send datastream record: " + metadata, exception);
+        LOG.debug("Failed to send datastream record: " + metadata, exception);
```
somandal, line 73: Is it possible to expose whether we're on shutdown path and only then print this as debug?
If not, please ensure that there are logs elsewhere to catch send failures from logs.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
```diff
@@ -75,7 +75,8 @@
   private static final AtomicInteger NUM_PRODUCERS = new AtomicInteger();
   private static final Supplier<Integer> PRODUCER_GAUGE = NUM_PRODUCERS::get;
 
-  private static final int CLOSE_TIMEOUT_MS = 2000;
+  private static final int DEFAULT_PRODUCER_CLOSE_TIMEOUT_MS = 10000;
```
somandal, line 78: Is 10 seconds sufficient? Should this be longer to give more time to drain records?

datastream-server-api/src/main/java/com/linkedin/datastream/server/FlushlessEventProducerHandler.java
vmaheshw: The caller of this method has the shutdown path check and prints it otherwise.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: I will be setting 60 sec via config for MM cases. The default is for change capture case. Along with this, in places where flushlessEventProducer is enabled, I will be setting the producerFlush timeout as well to 60 sec.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
vmaheshw: done

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java
vmaheshw: done

somandal (approved):
One minor comment, only if you feel it’ll be useful. Can you add a short comment explaining when to use fast close vs. slow close? It may be useful for future people working with the code

somandal (approved):

somandal (approved):

jzakaryan (approved):
LGTM

PR #812, opened 2021-03-06
Check if the task node exists on zookeeper before saving the status

vmaheshw:
The existing code creates the task node if it does not exists before saving the status. This can result in orphan node, if the task node was deleted as part of datastream stop(). This results in generation of orphan nodes. To avoid this situation, it is better to only save the state if the task node exists. If it does not exists in /connectors/<connectorType>/ it implies the leader has already cleaned it up and is not really required.


DEEPTHIKORAT (approved):
LGTM. What was the code path that called saveStatus on a task that was already unassigned? this looks like a good catchall place, but wondering if we should still cleanup such code paths in the connectors.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -907,8 +907,12 @@ public String getDatastreamTaskStateForKey(DatastreamTask datastreamTask, String
   public void setDatastreamTaskStateForKey(DatastreamTask datastreamTask, String key, String value) {
     String path = KeyBuilder.datastreamTaskStateKey(_cluster, datastreamTask.getConnectorType(),
         datastreamTask.getDatastreamTaskName(), key);
-    _zkclient.ensurePath(path);
-    _zkclient.writeData(path, value);
+    String taskPath = KeyBuilder.connectorTask(_cluster, datastreamTask.getConnectorType(),
+        datastreamTask.getDatastreamTaskName());
+    if (_zkclient.exists(taskPath)) {
+      _zkclient.ensurePath(path);
```
somandal, line 913: I'm guessing this ensurePath() is necessary just to save us from race conditions, where the node gets deleted after we check for existence?

somandal (approved):

vmaheshw:
> LGTM. What was the code path that called saveStatus on a task that was already unassigned? this looks like a good catchall place, but wondering if we should still cleanup such code paths in the connectors.

The AbstractKafkaConnectorTask thread run() loop saves state on exception before getting killed. If the task has moved, then it makes sense to keep the state. If the task is already deleted, it is okay to not save the state and status.

PR #815, opened 2021-03-11
Refactor KafkaProducerWrapper flush() call

vmaheshw:
Linkedin-Kafka-Clients has fixed the flush() method in their latest version to have the same contract as flush() from kafka producer open source. So, we don't need to separately handle the `IllegalStateException` and block the flush() call on close to complete. The new code will block the flush() for the timeout till either the timeout is completed or the close is completed.


somandal (approved):
lgtm!

jzakaryan (dismissed):
LGTM

somandal (approved):

jzakaryan (approved):

PR #816, opened 2021-03-22
Skip dependency check during lock acquire if there are no partitionV2 in dependency task

vmaheshw:
Currently, in `StickyPartitionAssignmentStrategy` while creation of new task during partition assignment from existing task checks for the dependency task to be in running state. It validates the running state by checking if the lock is acquired by this task. In case of slow/bad host, this lock will not be acquired, resulting in partition assignment failure. To avoid the scenario where the dependency is the one with no partition, ideally the new task does not need to set dependency on it. Dependency is set to ensure that no two tasks process the same partition at the same time. If the dependency does not have partition, then there is no need to set dependency.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
@@ -162,9 +162,14 @@ public class DatastreamTaskImpl implements DatastreamTask {
    * @param partitionsV2 new partitions for this task
    */
   public DatastreamTaskImpl(DatastreamTaskImpl predecessor, Collection<String> partitionsV2) {
+    _dependencies = new ArrayList<>();
     if (!predecessor.isLocked()) {
-      throw new DatastreamTransientException("task " + predecessor.getDatastreamTaskName() + " is not locked, "
-          + "the previous assignment has not been picked up");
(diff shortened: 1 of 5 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/DatastreamTaskImpl.java
```diff
@@ -162,9 +162,14 @@ public DatastreamTaskImpl(List<Datastream> datastreams, String id, List<Integer>
    * @param partitionsV2 new partitions for this task
    */
   public DatastreamTaskImpl(DatastreamTaskImpl predecessor, Collection<String> partitionsV2) {
+    _dependencies = new ArrayList<>();
     if (!predecessor.isLocked()) {
-      throw new DatastreamTransientException("task " + predecessor.getDatastreamTaskName() + " is not locked, "
-          + "the previous assignment has not been picked up");
+      if (!predecessor._partitionsV2.isEmpty()) {
```
jzakaryan, line 167: Combine these two conditions in a single if statement? That would make code a bit more readable.

jzakaryan (dismissed):
LGTM, other than the minor improvement comment. Ignore or make the change at your discretion.

somandal (approved):
lgtm!

jzakaryan (approved):

PR #822, opened 2021-05-04
Add action endpoint to get host instance for a task

shrinandthakkar:

### DIFF
--- a/datastream-common/src/main/idl/com.linkedin.datastream.server.dms.datastream.restspec.json
+++ b/datastream-common/src/main/idl/com.linkedin.datastream.server.dms.datastream.restspec.json
@@ -31,6 +31,16 @@
         "type" : "string"
       } ]
     } ],
+    "actions" : [ {
+      "name" : "getTaskAssignment",
+      "doc" : "Finds the task's assignment node which hosts the given task",
+      "parameters" : [ {
+        "name" : "task",
+        "type" : "string",
+        "doc" : "Datastream task name"
+      } ],
+      "returns" : "string"
+    } ],
     "entity" : {
       "path" : "/datastream/{datastreamId}",
       "actions" : [ {
--- a/datastream-common/src/main/idl/com.linkedin.datastream.server.dms.datastream.restspec.json
+++ b/datastream-common/src/main/idl/com.linkedin.datastream.server.dms.datastream.restspec.json
@@ -31,19 +31,18 @@
         "type" : "string"
       } ]
     } ],
-    "actions" : [ {
-      "name" : "getTaskAssignment",
-      "doc" : "Finds the task's assignment node which hosts the given task",
-      "parameters" : [ {
-        "name" : "task",
-        "type" : "string",
-        "doc" : "Datastream task name"
-      } ],
-      "returns" : "string"
-    } ],
     "entity" : {
       "path" : "/datastream/{datastreamId}",
       "actions" : [ {
+        "name" : "getTaskAssignment",
+        "doc" : "Finds the task's assignment node which hosts the given task",
+        "parameters" : [ {
+          "name" : "task",
+          "type" : "string",
+          "doc" : "Datastream task name"
+        } ],
+        "returns" : "string"
+      }, {
         "name" : "movePartitions",
         "doc" : "Move partitions to a particular host\nService Returns: Result HTTP status",
         "parameters" : [ {
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -695,11 +695,11 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
   }
 
   /**
-   * Finds the task's assignment node which hosts the given task 
+   * Finds the task's assignment node which hosts the given task
    * @param pathKeys Datastream resource key
    * @param task Datastream task name
    */
-  @Action(name = "getTaskAssignment")
+  @Action(name = "getTaskAssignment", resourceLevel = ResourceLevel.ENTITY)
   public String getTaskAssignment(@PathKeysParam PathKeys pathKeys, @ActionParam("task") String task) {
     // Get datastream
     String datastreamName = pathKeys.getAsString(KEY_NAME);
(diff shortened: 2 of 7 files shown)


datastream-common/src/main/idl/com.linkedin.datastream.server.dms.datastream.restspec.json
```diff
@@ -31,6 +31,16 @@
         "type" : "string"
       } ]
     } ],
+    "actions" : [ {
```
vmaheshw, line 34: Can you help me understand the organization of this json? I noticed that actions is inside entities. Sorry, I'm not aware of how this piece of code works.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -686,6 +686,37 @@ public Datastream get(String name) {
...
+  public String getTaskAssignment(@PathKeysParam PathKeys pathKeys, @ActionParam("task") String task) {
+    // Get datastream
+    String datastreamName = pathKeys.getAsString(KEY_NAME);
+
+    Datastream stream;
+    String assignedInstance = null;
+    try {
+      LOG.info("Get assigned instance for task: {}", task);
+      _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, GET_CALL, 1);
+      stream = _store.getDatastream(datastreamName);
```
vmaheshw, line 704: what if the stream does not exists or is in paused state?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -686,6 +686,37 @@ public Datastream get(String name) {
...
+      assignedInstance = _store.getRawData(_store.getConnectorTaskPath(stream.getConnectorName(), task));
+    } catch (Exception e) {
+      _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, CALL_ERROR, 1);
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_500_INTERNAL_SERVER_ERROR,
+          "Failed to get assigned instance hosting task: " + task, e);
+    }
+
+    if (assignedInstance == null) {
```
vmaheshw, line 712: what if the assignedInstance is empty for some reason.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
```diff
@@ -21,6 +21,16 @@
    */
   Datastream getDatastream(String key);
 
+  /**
+   * Retrieves raw data in json from the given path.
```
vmaheshw, line 25: raw data may not be in json.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
```diff
@@ -21,6 +21,16 @@
...
+   */
+  String getRawData(String path);
+
+  /**
+   * Retrieves connector task path specific to the store implementation.
+   */
+  String getConnectorTaskPath(String connector, String task);
```
vmaheshw, line 32: Why do you need this? Can you not use KeyBuilders directly?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
vmaheshw, line 704: Consider if you can encapsulate this logic inside store itself, rather than exposing the rawData outside.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/dms/TestDatastreamResources.java
```diff
@@ -153,6 +156,39 @@ public void testReadDatastream() throws Exception {
     Assert.assertEquals(ds, datastreamToCreate);
   }
 
+  @Test
+  public void testGettingInstanceHostingTask() {
+    DatastreamResources resource1 = new DatastreamResources(_datastreamKafkaCluster.getPrimaryDatastreamServer());
```
vmaheshw, line 161: If you decide to encapsulate the logic in the store, please add tests in TestZookeeperBackedDatastreamStore and you can use `zkclient` object to create the zookeeper state (good and bad cases) and test without mocking.

vmaheshw (changes requested):

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
shrinandthakkar: In paused state, it will return "PAUSED_INSTANCE", and yea I'll put a null check.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
shrinandthakkar: I handled this case in the tool logic which calls this endpoint, but thinking about it again, it makes more sense to handle that here.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
shrinandthakkar: Umm no, since DatastreamResources should only depend upon the DatastreamStore Iface and not on something which implements it, like zk. that's why I made helper functions in the store Iface which helps retrieve zk- related data.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
shrinandthakkar: Instead of making multiple customized functions to retrieve specific data from the store, I wanted to make a generic function to the store which the brooklin-server can customize and handle.

shrinandthakkar: updated, it throws DatastreamException which then further throws RestLiServiceException from catch block.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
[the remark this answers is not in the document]
shrinandthakkar: If we are anyways passing connector and task to retrieve instance in any custom method from datastream store, that implies that the path consists of connector and task. This anyways has the same assumption that any other store implementation would need the same arguments to implement that method. 

I just wrote a simple function instead to retrieve the paths, which could be reused (in getters/setters) and so a generic getter (of raw data) could handle retrieving data using these functions.

And since the brooklin-server accesses the datastream store, we can prevent data leaks, as it sits on a layer before the client library accesses it.

Just wanted to share my thought process, let me know @vmaheshw @jzakaryan if you still feel otherwise, I'll update the code :)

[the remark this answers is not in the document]
vmaheshw: My recommendation is to just expose one method here `getDatastreamTaskInstance(String connector, String task)`. The input parameters are justified. If you want you can pass Datastream instead of connector.

Since this is in open-source, brooklin-server layer is not sufficient to ensure that there is no data leaks.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
```diff
@@ -21,6 +21,11 @@
    */
   Datastream getDatastream(String key);
 
+  /**
+   * Retrieves the assignee node on which the provided task is handled
+   */
+  String getAssigneeNode(String datastream, String task);
```
vmaheshw, line 27: Probably `getAssignedTaskInstance` instead of `getAssigneeNode`.

vmaheshw (changes requested):

vmaheshw (approved):
lgtm

jzakaryan (approved):

PR #823, opened 2021-05-04
LoadBasedPartitionAssignmentStrategy - base framework

jzakaryan:
This PR introduces a set of abstractions for performing partition assignment based on per-partition throughput.

- `PartitionThroughputProvider` is an interface that fetches partition throughput information. A dummy no-op implementation, as well as an implementation that reads from a resource json file are included, along with auxiliary structures for the data (`ClusterThroughputInfo` and `PartitionThroughputInfo`) 
- `LoadBasedPartitionAssignmentStrategy`, which is an extension of `StickyPartitionAssignmentStrategy` and makes use of the the elastic task count validation and increase logic in the latter. `StickyPartitionAssignmentStrategy` is also used as a fallback strategy in case partition throughput data is not available.
- A factory that builds `LoadBasedPartitionAssignmentStrategy` instances.
- `LoadBasedTaskCountEstimator` and `LoadBasedPartitionAssigner` are designed to encapsulate the actual logic of throughput based task count estimation and partition assignment. The implementations of these classes are to be published in the PR to come.

the contribution guidelines.


gradle/license.gradle
```diff
@@ -7,6 +7,7 @@ subprojects {
         // Skip Twitter bootstrap JS and CSS.
         skipExistingHeaders = true
         ext.year = Calendar.getInstance().get(Calendar.YEAR)
+        exclude "*.json"
```
jzakaryan, line 10: This was added here to skip json files from license header tests during build.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,127 @@
...
+      updateOrRegisterElasticTaskAssignmentMetrics(datastreamPartitions, taskCount);
+    }
+
+    // Calculating unassigned partitions
+    List<String> unassignedPartitions = new ArrayList<>(datastreamPartitions.getPartitions());
+    unassignedPartitions.removeAll(assignedPartitions);
+
+    // Getting a task count estimate based on throughput
+    String clusterName = _sourceClusterResolver.getSourceCluster(datastreamPartitions.getDatastreamGroup());
+    LoadBasedTaskCountEstimator taskCountEstimator = new LoadBasedTaskCountEstimator();
```
jzakaryan, line 100: Should this be an interface and injected in the `LoadBasedPartitionAssignmentStrategy` constructor? Do we plan to have multiple implementations for this, and swap one with another?

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaTopicPartitionInfo.java
```diff
@@ -34,7 +34,7 @@
  * Tests for TopicPartitionStats diag command
  */
 @Test
-public class TestKafkaTopicPartitionStats extends BaseKafkaZkTest {
+public class TestKafkaTopicPartitionInfo extends BaseKafkaZkTest {
```
vmaheshw, line 37: what is the reason behind changing the class name?

jzakaryan: It was done by the refactoring tool. Reverting the change.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import com.linkedin.datastream.server.providers.PartitionThroughputProvider;
+
+import static com.linkedin.datastream.server.assignment.BroadcastStrategyFactory.CFG_MAX_TASKS;
+import static com.linkedin.datastream.server.assignment.StickyMulticastStrategyFactory.CFG_IMBALANCE_THRESHOLD;
+
+
+/**
+ * A factory for creating {@link LoadBasedPartitionAssignmentStrategy} instances
+ */
+public class LoadBasedPartitionAssignmentStrategyFactory implements AssignmentStrategyFactory {
```
vmaheshw, line 31: Can you extend `StickyPartitionAssignmentStrategyFactory`? Most of this code is duplicate and you should be able to get rid of it.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -0,0 +1,57 @@
...
+import com.linkedin.datastream.server.PartitionThroughputInfo;
+
+
+/**
+ * Estimates the minimum number of tasks for a datastream based on per-partition throughput information
+ */
+public class LoadBasedTaskCountEstimator {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedTaskCountEstimator.class.getName());
+  private final static int TASK_CAPACITY_MBPS = 4;
```
vmaheshw, line 27: // Get this via config. You can set this as default

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Gets the estimated number of tasks based on per-partition throughput information.
+   * NOTE: This does not take into account numPartitionsPerTask configuration
+   * @param throughputInfo Per-partition throughput information
+   * @param assignedPartitions The list of assigned partitions
+   * @param unassignedPartitions The list of unassigned partitions
+   * @return The estimated number of tasks
+   */
+  public int getTaskCount(ClusterThroughputInfo throughputInfo, List<String> assignedPartitions,
```
vmaheshw, line 37: Lets discuss the utility of this method offline.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -187,20 +209,15 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
     Validate.isTrue(currentAssignment.size() > 0,
         "Zero tasks assigned. Retry leader partition assignment.");
 
+    DatastreamGroup datastreamGroup = datastreamPartitions.getDatastreamGroup();
     String dgName = datastreamPartitions.getDatastreamGroup().getName();
```
vmaheshw, line 213: Since you are adding additional variable `datastreamGroup` use that to get `dgName`

datastream-server/src/main/java/com/linkedin/datastream/server/providers/FileBasedPartitionThroughputProvider.java
```diff
@@ -0,0 +1,120 @@
...
+import com.linkedin.datastream.server.ClusterThroughputInfo;
+import com.linkedin.datastream.server.PartitionThroughputInfo;
+
+
+/**
+ * An implementation of {@link PartitionThroughputProvider} which reads throughput data from a resource file.
+ * Used for testing purposes.
+ */
+public class FileBasedPartitionThroughputProvider implements PartitionThroughputProvider {
+  private static final String RESOURCE_FILE_NAME = "partitionThroughput.json";
```
vmaheshw, line 29: Will it make more sense to get this via Constructor, so that there is a knob to test additional json files with partition input without tweaking this file

datastream-server/src/main/java/com/linkedin/datastream/server/providers/FileBasedPartitionThroughputProvider.java
```diff
@@ -0,0 +1,120 @@
...
+  private static final String ROOT_NODE_NAME = "stats";
+
+  @Override
```
vmaheshw, line 32: //{@inheritdoc}

datastream-server/src/main/java/com/linkedin/datastream/server/providers/FileBasedPartitionThroughputProvider.java
```diff
@@ -0,0 +1,120 @@
...
+      JsonNode clusterStats) {
+    TypeReference<HashMap<String, String>> mapTypeRef = new TypeReference<HashMap<String, String>>() { };
+    HashMap<String, PartitionThroughputInfo> partitionInfoMap = new HashMap<>();
+
+    try {
+      HashMap<String, String> partitionStats = mapper.readValue(clusterStats, mapTypeRef);
+      for (String partition : partitionStats.keySet()) {
+        String value = partitionStats.get(partition);
+        String[] tokens = StringUtils.split(value, ",");
+        int bytesInRate = Integer.parseInt(StringUtils.substring(tokens[0], 11));
```
vmaheshw, line 110: can you add the logic instead of hardcoding 11 and 7

datastream-server/src/test/java/com/linkedin/datastream/server/TestFileBasedPartitionThroughputProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import org.testng.annotations.Test;
+
+import com.linkedin.datastream.server.providers.FileBasedPartitionThroughputProvider;
+
+
+/**
+ * Tests for {@link FileBasedPartitionThroughputProvider}
+ */
+public class TestFileBasedPartitionThroughputProvider {
+
```
vmaheshw, line 20: For the open source `File` based solution, you may want to add more negative tests. like cluster not present, file not present.

vmaheshw (commented):
Please split future PRs into smaller chunks. For eg. 

a) Just the interfaces.
b) New Strategy
c) Task count prediction logic.

Currently, in this PR you have added new strategy without tests. Please confirm your plan and next steps.

datastream-server-api/src/main/java/com/linkedin/datastream/server/ClusterThroughputInfo.java
```diff
@@ -0,0 +1,69 @@
...
+  private String _clusterName;
+
+  private HashMap<String, PartitionThroughputInfo> _partitionInfoMap;
+
+  /**
+   * Creates an instance of {@link ClusterThroughputInfo}
+   * @param clusterName Cluster name
+   * @param partitionInfoMap A map, where the key is the partition name, and the value is the
+   *                         {@link PartitionThroughputInfo} for the partition.
+   */
```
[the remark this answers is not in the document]
jzakaryan: Assignment strategy only reads from this map. As long as `ThroughputProvider` handles the initialization carefully, I don't see how two threads would race for this structure.

datastream-server/src/main/java/com/linkedin/datastream/server/providers/FileBasedPartitionThroughputProvider.java
jzakaryan: Done

jzakaryan:
> Please split future PRs into smaller chunks. For eg.
> 
> a) Just the interfaces.
> b) New Strategy
> c) Task count prediction logic.
> 
> Currently, in this PR you have added new strategy without tests. Please confirm your plan and next steps.

The PRs to come would be smaller. I left testing for later, because I wanted to have the assignment piece done to implement tests for this. For now it even if the new assignment strategy is enabled, it will default to `StickyMulticastPartitionAssignmentStrategy`, because no actual throughput provider is supplied in the factory.

datastream-server-api/src/main/java/com/linkedin/datastream/server/ClusterThroughputInfo.java
```diff
@@ -0,0 +1,69 @@
...
+  public void setClusterName(String clusterName) {
+    _clusterName = clusterName;
+  }
+
+  /**
+   * Sets the partition information map.
+   * @param partitionInfoMap A map, where the key is the partition name, and the value is a {@link PartitionThroughputInfo}
+   *                         for the partition
+   */
+  public void setPartitionInfoMap(HashMap<String, PartitionThroughputInfo> partitionInfoMap) {
```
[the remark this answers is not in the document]
jzakaryan: I changed this class so now the setters are gone.

datastream-server-api/src/main/java/com/linkedin/datastream/server/ClusterThroughputInfo.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  public HashMap<String, PartitionThroughputInfo> getPartitionInfoMap() {
+    return _partitionInfoMap;
+  }
+
+  /**
+   * Sets the cluster name
+   * @param clusterName Name of the cluster
+   */
+  public void setClusterName(String clusterName) {
```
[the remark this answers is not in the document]
jzakaryan: Got rid of the setters and made it immutable.

datastream-server-api/src/main/java/com/linkedin/datastream/server/ClusterThroughputInfo.java
```diff
@@ -0,0 +1,69 @@
...
+  public String getClusterName() {
+    return _clusterName;
+  }
+
+  /**
+   * Gets the partition information map for partitions in the cluster
+   * @return A map, where the key is the partition name, and the value is a {@link PartitionThroughputInfo} for the
+   * partition
+   */
+  public HashMap<String, PartitionThroughputInfo> getPartitionInfoMap() {
```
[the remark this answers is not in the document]
jzakaryan: I was thinking of doing that, but that would incur overhead of copying object refs to the new map. And in production clusters we expect this to be a pretty big map. Do you think it's worth it?

datastream-server-api/src/main/java/com/linkedin/datastream/server/providers/PartitionThroughputProvider.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.HashMap;
+
+import com.linkedin.datastream.server.ClusterThroughputInfo;
+
+
+/**
+ * Abstraction that provides topic partition throughput information. Used by load-based assignment strategies to do
+ * partition assignment based on throughput.
+ */
+public interface PartitionThroughputProvider {
```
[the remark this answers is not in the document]
jzakaryan: I think the work Partition already implies Topic Partition, because it's not overloaded semantically with different meanings(s) in brooklin. By that same logic we would have to rename `StickyPartitionAssignmentStrategy` to `StickyTopicPartitionAssignmentStrategy`. Not necessary, in my opinion.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Partition assignment strategy that does assignment based on throughput information supplied by a
+ * {@link PartitionThroughputProvider}
+ */
+public class LoadBasedPartitionAssignmentStrategy extends StickyPartitionAssignmentStrategy {
+  private final PartitionThroughputProvider _throughputProvider;
+  private final DatastreamSourceClusterResolver _sourceClusterResolver;
+
+  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedPartitionAssignmentStrategy.class.getName());
+  private static final long THROUGHPUT_INFO_FETCH_TIMEOUT_MS_DEFAULT = Duration.ofSeconds(10).toMillis();
```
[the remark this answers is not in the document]
jzakaryan: I'm planning to make this configurable

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+  private static final long THROUGHPUT_INFO_FETCH_RETRY_PERIOD_MS_DEFAULT = Duration.ofSeconds(1).toMillis();
+
+  /**
+   * Creates an instance of {@link LoadBasedPartitionAssignmentStrategy}
+   */
+  public LoadBasedPartitionAssignmentStrategy(PartitionThroughputProvider throughputProvider,
+      DatastreamSourceClusterResolver sourceClusterResolver, Optional<Integer> maxTasks,
```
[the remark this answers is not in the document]
jzakaryan: Why? I think interfaces should come first.

datastream-server-api/src/main/java/com/linkedin/datastream/server/ClusterThroughputInfo.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  public void setClusterName(String clusterName) {
+    _clusterName = clusterName;
+  }
+
+  /**
+   * Sets the partition information map.
+   * @param partitionInfoMap A map, where the key is the partition name, and the value is a {@link PartitionThroughputInfo}
+   *                         for the partition
+   */
```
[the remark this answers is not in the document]
jzakaryan: I think this question is no longer relevant, after I made removed the setters

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyFactory.java
jzakaryan: I addressed this by moving configs to a separate class and extending `LoadBasedPartitionAssignmentStrategy` from `StickyPartitionAssignmentStrategy`.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -0,0 +1,33 @@
...
+public class LoadBasedPartitionAssigner {
+  /**
+   * Get partition assignment
+   * @param throughputInfo Per-partition throughput information
+   * @param assignedPartitions List of assigned partitions
+   * @param unassignedPartitions List of unassigned partitions
+   * @return Partition assignment
+   */
+  public Map<String, Set<DatastreamTask>> assignPartitions(ClusterThroughputInfo throughputInfo,
+      List<String> assignedPartitions, List<String> unassignedPartitions, int taskCount) {
```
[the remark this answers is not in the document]
jzakaryan: I will address this when putting out the actual PR for assignment. This was mostly meant as a stub here.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public Map<String, Set<DatastreamTask>> assignPartitions(Map<String, Set<DatastreamTask>> currentAssignment,
+      DatastreamGroupPartitionsMetadata datastreamPartitions) {
+    DatastreamGroup datastreamGroup = datastreamPartitions.getDatastreamGroup();
+    String datastreamGroupName = datastreamGroup.getName();
+    HashMap<String, ClusterThroughputInfo> partitionThroughputInfo;
+
+    // Attempting to retrieve partition throughput info with a fallback mechanism to StickyPartitionAssignmentStrategy
+    try {
+      partitionThroughputInfo = fetchPartitionThroughputInfo();
```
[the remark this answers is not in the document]
jzakaryan: That needs to be addressed in the base strategy as well. Let's discuss this offline

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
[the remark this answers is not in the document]
jzakaryan: Added a todo comment

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+    int maxTaskCount = Math.max(taskCount, taskCount2);
+
+    // Doing assignment
+    LoadBasedPartitionAssigner partitionAssigner = new LoadBasedPartitionAssigner();
+    return partitionAssigner.assignPartitions(clusterThroughputInfo, assignedPartitions, unassignedPartitions,
+        maxTaskCount);
+  }
+
+  private HashMap<String, ClusterThroughputInfo> fetchPartitionThroughputInfo() {
+    PollUtils.poll(() -> {
```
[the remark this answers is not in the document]
jzakaryan: The poll method is blocking, but it will return something before the specified timeout. I configured the call to throw an exception if the retries are exhausted.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: I removed the code for this to put in a separate PR with tests.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.commons.lang.NotImplementedException;
+
+import com.linkedin.datastream.server.ClusterThroughputInfo;
+import com.linkedin.datastream.server.DatastreamTask;
+
+
+/**
+ * Performs partition assignment based on partition throughput information
+ */
+public class LoadBasedPartitionAssigner {
```
atoomula, line 21: Shouldn't this be a pluggable interface ?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+
+  private Map<String, ClusterThroughputInfo> fetchPartitionThroughputInfo() {
+    return PollUtils.poll(() -> {
+      try {
+        return _throughputProvider.getThroughputInfo();
+      } catch (Exception ex) {
+        LOG.warn("Failed to fetch partition throughput info.");
+        return null;
+      }
+    }, Objects::nonNull, THROUGHPUT_INFO_FETCH_RETRY_PERIOD_MS_DEFAULT, THROUGHPUT_INFO_FETCH_TIMEOUT_MS_DEFAULT)
```
atoomula, line 119: If there are capacity issues to serve this request, would it help to have exponential retries ?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public Map<String, Set<DatastreamTask>> assignPartitions(Map<String, Set<DatastreamTask>> currentAssignment,
+      DatastreamGroupPartitionsMetadata datastreamPartitions) {
+    DatastreamGroup datastreamGroup = datastreamPartitions.getDatastreamGroup();
+    String datastreamGroupName = datastreamGroup.getName();
+    Map<String, ClusterThroughputInfo> partitionThroughputInfo;
+
+    // Attempting to retrieve partition throughput info with a fallback mechanism to StickyPartitionAssignmentStrategy
+    try {
+      partitionThroughputInfo = fetchPartitionThroughputInfo();
```
atoomula, line 69: We need to be super careful on how often fetchPartitionThroughputInfo() is called. We might want to add code here to avoid calling this too often, skip this step, throw an alert and move on.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
jzakaryan: We had this discussion with Vaibhav as well. For now we'll keep it as a class. When we have multiple implementations, we'll make it an interface.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
jzakaryan: We discussed this with @vmaheshw as well. We'll only be doing throughput based partition assignment initially (when no partitions are assigned), and not incrementally. I updated the PR with that change.

datastream-server-api/src/main/java/com/linkedin/datastream/server/ClusterThroughputInfo.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map;
+
+
+/**
+ * A structure that holds per-partition throughput information for a Kafka cluster
+ */
+public class ClusterThroughputInfo {
+
+  private final String _clusterName;
+
```
vmaheshw, line 17: nit: additional space.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Get partition assignment
+   * @param throughputInfo Per-partition throughput information
+   * @param assignedPartitions List of assigned partitions
+   * @param unassignedPartitions List of unassigned partitions
```
vmaheshw, line 26: `taskCount` missing in javadocs.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,114 @@
...
+    return partitionAssigner.assignPartitions(clusterThroughputInfo, assignedPartitions, unassignedPartitions,
+        maxTaskCount);
+  }
+
+  private Map<String, ClusterThroughputInfo> fetchPartitionThroughputInfo() {
+    return PollUtils.poll(() -> {
+      try {
+        return _throughputProvider.getThroughputInfo();
+      } catch (Exception ex) {
+        LOG.warn("Failed to fetch partition throughput info.");
```
vmaheshw, line 108: nit: may be interested in exception. Also retry count in log will be helpful. 

You can pick this in next PR.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyFactory.java
```diff
@@ -0,0 +1,58 @@
...
+    PartitionAssignmentStrategyConfig config = new PartitionAssignmentStrategyConfig(assignmentStrategyProperties);
+
+    boolean enableElasticTaskAssignment = config.isElasticTaskAssignmentEnabled();
+    // Create the zookeeper client
+    Optional<ZkClient> zkClient = Optional.empty();
+    try {
+      zkClient = constructZooKeeperClient(enableElasticTaskAssignment, config.getZkAddress(),
+          config.getZkSessionTimeout(), config.getZkSessionTimeout());
+    } catch (IllegalStateException ex) {
+      LOG.warn("Disabling elastic task assignment as zkClient initialization failed");
```
vmaheshw, line 39: printing the exception may help in debugging why zkclient construction failed.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyFactory.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public AssignmentStrategy createStrategy(Properties assignmentStrategyProperties) {
+    PartitionAssignmentStrategyConfig config = new PartitionAssignmentStrategyConfig(assignmentStrategyProperties);
+
+    boolean enableElasticTaskAssignment = config.isElasticTaskAssignmentEnabled();
+    // Create the zookeeper client
+    Optional<ZkClient> zkClient = Optional.empty();
+    try {
+      zkClient = constructZooKeeperClient(enableElasticTaskAssignment, config.getZkAddress(),
```
vmaheshw, line 36: this creation should happen only if `enableElastTaskAssignment` is set to true. Also, I think this can be simplified as all the parameters are present in config in the base class itself.
Base class can have something like this
`Optional<ZkClient> createZkClient() {
         if (config.isElasticTaskAssignmentEnabled()) {
             zkClient = construct ....
         }
     return zkClient
}`

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategyFactory.java
```diff
@@ -12,69 +12,49 @@
...
-    // Create the ZooKeeper Client
+    boolean enableElasticTaskAssignment = config.isElasticTaskAssignmentEnabled();
+    // Create the zookeeper client
     Optional<ZkClient> zkClient = Optional.empty();
-    String zkAddress = props.getString(CFG_ZK_ADDRESS, null);
-    if (enableElasticTaskAssignment && StringUtils.isBlank(zkAddress)) {
-      LOG.warn("Disabling elastic task assignment as zkAddress is not present or empty");
+    try {
+      zkClient = constructZooKeeperClient(enableElasticTaskAssignment, config.getZkAddress(),
+          config.getZkSessionTimeout(), config.getZkSessionTimeout());
```
vmaheshw, line 35: You are using `config.getZkSessionTimeout` instead of `config.getZkConnectionTimeout` for the 4th field.

Can you check if these kind of errors can be caught using tests. This reflects that our existing tests are missing some checks.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyFactory.java
```diff
@@ -0,0 +1,58 @@
...
+          config.getZkSessionTimeout(), config.getZkSessionTimeout());
```
vmaheshw, line 37: `config.getZkSessionTimeout()` used instead of `config.getZkConnectionTimeout` for the last field. We should get rid of duplicate code.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -171,6 +171,28 @@ public StickyPartitionAssignmentStrategy(Optional<Integer> maxTasks, Optional<In
     return super.assign(datastreams, instances, currentAssignment);
   }
 
+  protected List<String> getAssignedPartitionsForDatastreamGroup(Map<String, Set<DatastreamTask>> currentAssignment,
```
vmaheshw, line 174: Would it have worked, if you would have returned `Pair`. This change is directly adding an additional iteration to the existing code as well.

Also, for the existing strategy, it will walk the tasks list twice now and for new strategy, it will walk 3 times. There is a scope to extract out another method that gets assignedPartitions and taskcount as field and the new strategy can call it to avoid extra iteration.

See if you can address the `Pair` one, so it is exactly same as current code and you can mark TODO to reduce additional iteration for new strategy.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
```diff
@@ -536,7 +553,7 @@ protected int constructExpectedNumberOfTasks(DatastreamGroup dg, List<String> in
     return expectedNumberOfTasks;
   }
 
-  private void updateOrRegisterElasticTaskAssignmentMetrics(DatastreamGroupPartitionsMetadata datastreamPartitions,
+  protected void updateOrRegisterElasticTaskAssignmentMetrics(DatastreamGroupPartitionsMetadata datastreamPartitions,
```
vmaheshw, line 556: This change implies that we will be exposing the following configuration options for rollback and forward deployment.
a) Regular StickyPartitionAssignmentStrategy.
b) StickyPartitionAssignmentStrategy with elastic task count based on partition count.
c) LoadBalancedAssigmentStrategy with elastic task count based on partition count and elastic task count based on throughput info.
Basically if elastic Task count config is enabled with new strategy, it will do based on both partition count and throughput. We don't have knob to disable one calculation. If we want to rollback to just task count based on partition count, we will have to change the strategy itself.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategyFactory.java
```diff
@@ -12,69 +12,49 @@
...
 
+    return new StickyPartitionAssignmentStrategy(config.getMaxTasks(), config.getImbalanceThreshold(),
+        config.getMaxPartitions(), enableElasticTaskAssignment, config.getPartitionsPerTask(),
+        config.getPartitionFullnessThresholdPct(), zkClient, config.getCluster());
+  }
+
+  protected Optional<ZkClient> constructZooKeeperClient(boolean enableElasticTaskAssignment, String zkAddress,
+      int zkSessionTimeout, int zkConnectionTimeout) {
+    Optional<ZkClient> zkClient = Optional.empty();
+    if (enableElasticTaskAssignment && StringUtils.isBlank(zkAddress)) {
```
vmaheshw, line 49: Can you simplify this to 

if (!enableElasticTaskAssignment)
    return Optional.empty() 
if (StringUtils.isBlank(zkAddress)
    throw exception
 return createZkClient(...)

datastream-server/src/test/java/com/linkedin/datastream/server/TestFileBasedPartitionThroughputProvider.java
```diff
@@ -0,0 +1,46 @@
...
+    FileBasedPartitionThroughputProvider provider = new FileBasedPartitionThroughputProvider(THROUGHPUT_FILE_NAME);
+    Map<String, ClusterThroughputInfo> clusterInfoMap = provider.getThroughputInfo();
+    Assert.assertNotNull(clusterInfoMap);
+  }
+
+  @Test(expectedExceptions = IllegalArgumentException.class)
+  public void getPartitionThroughputThrowsWhenFileNotFound() {
+    String dummyFileName = "dummy.json";
+    PartitionThroughputProvider provider = new FileBasedPartitionThroughputProvider(dummyFileName);
+    Map<String, ClusterThroughputInfo> clusterInfoMap = provider.getThroughputInfo();
```
vmaheshw, line 44: can you change this to 
`Assert.assertThrows(IllegalArgumentException.class, function);`

This ensures that the exception is indeed thrown at the expected line and not in the earlier lines and the test is still passing.

vmaheshw (changes requested):

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyFactory.java
jzakaryan: That's what happens @vmaheshw. Here's the method from the base class:

```
protected Optional<ZkClient> constructZooKeeperClient(boolean enableElasticTaskAssignment, String zkAddress,
      int zkSessionTimeout, int zkConnectionTimeout) {
    Optional<ZkClient> zkClient = Optional.empty();
    if (enableElasticTaskAssignment && StringUtils.isBlank(zkAddress)) {
      LOG.warn("ZkAddress is not present or empty");
      throw new IllegalStateException("ZkAddress is empty or not provided");
    }

    if (enableElasticTaskAssignment) {
      zkClient = Optional.of(new ZkClient(zkAddress, zkSessionTimeout, zkConnectionTimeout));
    }

    return zkClient;
  }
```

vmaheshw: Right, Then there is no point of retrieving the values in both the places. Your constructZkClient method should work without passing any field and using the existing config object itself.

You should not need to tweak enableElasticTaskAssignment in both base and derived class.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
jzakaryan: It's a tradeoff between keeping the code readable and avoiding extra iteration. I can make the method calculate both assigned partitions and current task count and return a pair if you think it's better that way.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
jzakaryan: Added a todo comment

datastream-server-api/src/main/java/com/linkedin/datastream/server/ClusterThroughputInfo.java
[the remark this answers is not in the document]
jzakaryan: The implementation will use `ConcurrentHashMap` for thread safety.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
jzakaryan: I'll document the rollout and rollback instructions separately.

vmaheshw (approved):
lgtm.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
jzakaryan: Our poll utilities don't support exponential timeout at the moment. But I think it will be useful in general to have that, I'll create a separate PR for that.

atoomula (approved):

PR #825, opened 2021-05-12
Add kafka source connection string validation in datastream update path

vmaheshw:
Add validation in datastream update path to reject request for incorrect source regex connection string.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -332,6 +334,23 @@ public void validateUpdateDatastreams(List<Datastream> datastreams, List<Datastr
       throws DatastreamValidationException {
     // validate for paused partitions
     validatePausedPartitions(datastreams, allDatastreams);
+    //validate connection string
+    for (Datastream datastream : datastreams) {
+      validateSourceConnectionString(datastream);
+    }
+  }
+
+  protected void validateSourceConnectionString(Datastream stream) throws DatastreamValidationException {
```
somandal, line 343: Should this only be added for KafkaMirrorMakerConnector, since other connectors that use this abstract base class don't necessarily use regexes (KafkaConnector,  MySqlKafkaConnector)? Perhaps override the `validateUpdateDatastreams` function in the KafkaMirrorMakerConnector, call the super's `validateUpdateDatastreams` and add the logic for source regex validation there?

somandal (approved):
lgtm

atoomula (approved):

PR #829, opened 2021-05-25
Implemented load based task count estimator

jzakaryan:
Implementation of LoadBasedTaskCountEstimator and its tests.

the contribution guidelines.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
@@ -7,6 +7,7 @@ package com.linkedin.datastream.server.assignment;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -40,8 +41,13 @@ public class LoadBasedPartitionAssignmentStrategy extends StickyPartitionAssignm
   private static final long THROUGHPUT_INFO_FETCH_TIMEOUT_MS_DEFAULT = Duration.ofSeconds(10).toMillis();
   private static final long THROUGHPUT_INFO_FETCH_RETRY_PERIOD_MS_DEFAULT = Duration.ofSeconds(1).toMillis();
 
+  private static final int TASK_CAPACITY_MBPS_DEFAULT = 4;
+  private static final int TASK_CAPACITY_UTILIZATION_PCT_DEFAULT = 90;
+
   private final PartitionThroughputProvider _throughputProvider;
   private final DatastreamSourceClusterResolver _sourceClusterResolver;
+  private final int _taskCapacityMBps;
+  private final int _taskCapacityUtilizationPct;
 
 
   /**
@@ -50,12 +56,15 @@ public class LoadBasedPartitionAssignmentStrategy extends StickyPartitionAssignm
   public LoadBasedPartitionAssignmentStrategy(PartitionThroughputProvider throughputProvider,
       DatastreamSourceClusterResolver sourceClusterResolver, Optional<Integer> maxTasks,
       Optional<Integer> imbalanceThreshold, Optional<Integer> maxPartitionPerTask, boolean enableElasticTaskAssignment,
-      Optional<Integer> partitionsPerTask, Optional<Integer> partitionFullnessFactorPct, Optional<ZkClient> zkClient,
+      Optional<Integer> partitionsPerTask, Optional<Integer> partitionFullnessFactorPct,
+      Optional<Integer> taskCapacityMBps, Optional<Integer> taskCapacityUtilizationPct, Optional<ZkClient> zkClient,
       String clusterName) {
     super(maxTasks, imbalanceThreshold, maxPartitionPerTask, enableElasticTaskAssignment, partitionsPerTask,
         partitionFullnessFactorPct, zkClient, clusterName);
     _throughputProvider = throughputProvider;
     _sourceClusterResolver = sourceClusterResolver;
+    _taskCapacityMBps = taskCapacityMBps.orElse(TASK_CAPACITY_MBPS_DEFAULT);
+    _taskCapacityUtilizationPct = taskCapacityUtilizationPct.orElse(TASK_CAPACITY_UTILIZATION_PCT_DEFAULT);
   }
 
   @Override
@@ -85,17 +94,18 @@ public class LoadBasedPartitionAssignmentStrategy extends StickyPartitionAssignm
     Validate.isTrue(currentAssignment.size() > 0,
         "Zero tasks assigned. Retry leader partition assignment.");
 
+    // Resolving cluster name from datastream group
+    String clusterName = _sourceClusterResolver.getSourceCluster(datastreamPartitions.getDatastreamGroup());
+    ClusterThroughputInfo clusterThroughputInfo = partitionThroughputInfo.get(clusterName);
+
     // TODO Get task count estimate and perform elastic task count validation
     // TODO Get task count estimate based on throughput and pick a winner
-    int maxTaskCount = 0;
+    LoadBasedTaskCountEstimator estimator = new LoadBasedTaskCountEstimator(_taskCapacityMBps, _taskCapacityUtilizationPct);
+    int maxTaskCount = estimator.getTaskCount(clusterThroughputInfo, Collections.emptyList(), Collections.emptyList());
 
     // TODO Get unassigned partitions
     // Calculating unassigned partitions
     List<String> unassignedPartitions = new ArrayList<>();
-
-    // Resolving cluster name from datastream group
-    String clusterName = _sourceClusterResolver.getSourceCluster(datastreamPartitions.getDatastreamGroup());
-    ClusterThroughputInfo clusterThroughputInfo = partitionThroughputInfo.get(clusterName);
 
     // Doing assignment
     LoadBasedPartitionAssigner partitionAssigner = new LoadBasedPartitionAssigner();
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
@@ -22,8 +22,9 @@ import com.linkedin.datastream.server.PartitionThroughputInfo;
  * Estimates the minimum number of tasks for a datastream based on per-partition throughput information
  */
 public class LoadBasedTaskCountEstimator {
-  private final static int BYTES_IN_KB_RATE_DEFAULT = 5;
-  private final static int MESSAGES_IN_RATE_DEFAULT = 5;
+  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedTaskCountEstimator.class.getName());
+  private static final int BYTES_IN_KB_RATE_DEFAULT = 5;
+  private static final int MESSAGES_IN_RATE_DEFAULT = 5;
 
   private final int _taskCapacityMBps;
   private final int _taskCapacityUtilizationPct;
@@ -37,8 +38,6 @@ public class LoadBasedTaskCountEstimator {
     _taskCapacityMBps = taskCapacityMBps;
     _taskCapacityUtilizationPct = taskCapacityUtilizationPct;
   }
-
-  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedTaskCountEstimator.class.getName());
 
   /**
    * Gets the estimated number of tasks based on per-partition throughput information.
@@ -59,10 +58,12 @@ public class LoadBasedTaskCountEstimator {
     Set<String> allPartitions = new HashSet<>(assignedPartitions);
     allPartitions.addAll(unassignedPartitions);
 
+    PartitionThroughputInfo defaultThroughputInfo = new PartitionThroughputInfo(BYTES_IN_KB_RATE_DEFAULT,
+        MESSAGES_IN_RATE_DEFAULT, "");
     // total throughput in KB/sec
-    int totalThroughput = allPartitions.stream().map(p ->
-        throughputMap.getOrDefault(p, new PartitionThroughputInfo(BYTES_IN_KB_RATE_DEFAULT, MESSAGES_IN_RATE_DEFAULT, p))).
-        mapToInt(PartitionThroughputInfo::getBytesInKBRate).sum();
+    int totalThroughput = allPartitions.stream()
+        .map(p -> throughputMap.getOrDefault(p, defaultThroughputInfo))
+        .mapToInt(PartitionThroughputInfo::getBytesInKBRate).sum();
     LOG.info("Total throughput in all {} partitions: {}KB/sec", allPartitions.size(), totalThroughput);
 
     double taskCapacityUtilizationCoefficient = _taskCapacityUtilizationPct / 100.0;
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
@@ -14,8 +14,6 @@ import org.apache.commons.lang3.Validate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.annotations.VisibleForTesting;
-
 import com.linkedin.datastream.server.ClusterThroughputInfo;
 import com.linkedin.datastream.server.PartitionThroughputInfo;
 
@@ -24,12 +22,23 @@ import com.linkedin.datastream.server.PartitionThroughputInfo;
  * Estimates the minimum number of tasks for a datastream based on per-partition throughput information
  */
 public class LoadBasedTaskCountEstimator {
+  private final static int BYTES_IN_KB_RATE_DEFAULT = 5;
+  private final static int MESSAGES_IN_RATE_DEFAULT = 5;
+
+  private final int _taskCapacityMBps;
+  private final int _taskCapacityUtilizationPct;
+
+  /**
+   * Creates an instance of {@link LoadBasedTaskCountEstimator}
+   * @param taskCapacityMBps Task capacity in MB/sec
+   * @param taskCapacityUtilizationPct Task capacity utilization percentage
+   */
+  public LoadBasedTaskCountEstimator(int taskCapacityMBps, int taskCapacityUtilizationPct) {
+    _taskCapacityMBps = taskCapacityMBps;
+    _taskCapacityUtilizationPct = taskCapacityUtilizationPct;
+  }
 
   private static final Logger LOG = LoggerFactory.getLogger(LoadBasedTaskCountEstimator.class.getName());
-
-  // TODO Make this constant configurable
-  @VisibleForTesting
-  public final static int TASK_CAPACITY_MBPS_DEFAULT = 4;
 
   /**
    * Gets the estimated number of tasks based on per-partition throughput information.
@@ -51,13 +60,16 @@ public class LoadBasedTaskCountEstimator {
     allPartitions.addAll(unassignedPartitions);
 
     // total throughput in KB/sec
-    int totalThroughput = throughputMap.entrySet().stream().
-        filter(e -> allPartitions.contains(e.getKey())).mapToInt(e -> e.getValue().getBytesInKBRate()).sum();
-    LOG.info("Total throughput in all partitions: {}KB/sec", totalThroughput);
+    int totalThroughput = allPartitions.stream().map(p ->
+        throughputMap.getOrDefault(p, new PartitionThroughputInfo(BYTES_IN_KB_RATE_DEFAULT, MESSAGES_IN_RATE_DEFAULT, p))).
+        mapToInt(PartitionThroughputInfo::getBytesInKBRate).sum();
+    LOG.info("Total throughput in all {} partitions: {}KB/sec", allPartitions.size(), totalThroughput);
 
-    int taskCountEstimate = (int) Math.ceil((double) totalThroughput / (TASK_CAPACITY_MBPS_DEFAULT * 1024));
+    double taskCapacityUtilizationCoefficient = _taskCapacityUtilizationPct / 100.0;
+    int taskCountEstimate = (int) Math.ceil((double) totalThroughput /
+        (_taskCapacityMBps * 1024 * taskCapacityUtilizationCoefficient));
     taskCountEstimate = Math.min(allPartitions.size(), taskCountEstimate);
-    LOG.debug("Estimated number of tasks required to handle the throughput: {}", taskCountEstimate);
+    LOG.info("Estimated number of tasks required to handle the throughput: {}", taskCountEstimate);
     return taskCountEstimate;
   }
 }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
@@ -20,9 +20,14 @@ import com.linkedin.datastream.server.PartitionThroughputInfo;
 
 /**
  * Estimates the minimum number of tasks for a datastream based on per-partition throughput information
+ * <p>
+ *   The reason for having a separate class for task count estimation is the intent to refactor other task count
+ *   estimation mechanisms (e.g. elastic task count estimation) and make them all implement the same interface.
+ * </p>
  */
 public class LoadBasedTaskCountEstimator {
   private static final Logger LOG = LoggerFactory.getLogger(LoadBasedTaskCountEstimator.class.getName());
+  // TODO Move these to config class
   private static final int BYTES_IN_KB_RATE_DEFAULT = 5;
   private static final int MESSAGES_IN_RATE_DEFAULT = 5;
 
@@ -52,9 +57,10 @@ public class LoadBasedTaskCountEstimator {
     Validate.notNull(throughputInfo, "null throughputInfo");
     Validate.notNull(assignedPartitions, "null assignedPartitions");
     Validate.notNull(unassignedPartitions, "null unassignedPartitions");
+    LOG.info("Assigned partitions: {}", assignedPartitions);
+    LOG.info("Unassigned partitions: {}", unassignedPartitions);
 
     Map<String, PartitionThroughputInfo> throughputMap = throughputInfo.getPartitionInfoMap();
-
     Set<String> allPartitions = new HashSet<>(assignedPartitions);
     allPartitions.addAll(unassignedPartitions);
 
@@ -63,7 +69,8 @@ public class LoadBasedTaskCountEstimator {
     // total throughput in KB/sec
     int totalThroughput = allPartitions.stream()
         .map(p -> throughputMap.getOrDefault(p, defaultThroughputInfo))
-        .mapToInt(PartitionThroughputInfo::getBytesInKBRate).sum();
+        .mapToInt(PartitionThroughputInfo::getBytesInKBRate)
+        .sum();
     LOG.info("Total throughput in all {} partitions: {}KB/sec", allPartitions.size(), totalThroughput);
 
     double taskCapacityUtilizationCoefficient = _taskCapacityUtilizationPct / 100.0;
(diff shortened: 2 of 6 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,22 @@
...
+
+    Set<String> allPartitions = new HashSet<>(assignedPartitions);
+    allPartitions.addAll(unassignedPartitions);
+
+    // total throughput in KB/sec
+    int totalThroughput = throughputMap.entrySet().stream().
+        filter(e -> allPartitions.contains(e.getKey())).mapToInt(e -> e.getValue().getBytesInKBRate()).sum();
+    LOG.debug("Total throughput in all partitions: {}KB/sec", totalThroughput);
+
+    int taskCountEstimate = (int) Math.ceil((double) totalThroughput / (TASK_CAPACITY_MBPS_DEFAULT * 1024));
```
shrinandthakkar, line 58: do we know if the KBps rate we get from the provider is KB or KiB as there are a couple of conventions going around with that and based on it, we might have to change 1024 to 1000?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,22 @@
...
+    Validate.notNull(assignedPartitions, "null assignedPartitions");
+    Validate.notNull(unassignedPartitions, "null unassignedPartitions");
+
+    Map<String, PartitionThroughputInfo> throughputMap = throughputInfo.getPartitionInfoMap();
+
+    Set<String> allPartitions = new HashSet<>(assignedPartitions);
+    allPartitions.addAll(unassignedPartitions);
+
+    // total throughput in KB/sec
+    int totalThroughput = throughputMap.entrySet().stream().
```
vmaheshw, line 54: 1. Should you account some default weight for the partitions for which you did not get entry in throughput map? For eg. We may setup query to get all the partitions with more than 25KBps, in which case, we can assume the partitions with <25KBps to have some default weight rather than 0.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
vmaheshw, line 58: Along with `ceil`, I feel there is a need for some additional buffer. This task estimator should keep some scope for growth rather than tightly packing all the partitions in the task.

Since the kafka limit is 4MB, I think one way to get additional tasks is by either considering the default to be 3MB, which in other words means 75% throughput utilization and additional 25% buffer to grow, without increasing the task itself. 

The distribution will have some gaps/holes and it demands for some approximation.

The task estimator will need some smartness in these aspects

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,22 @@
    */
   public int getTaskCount(ClusterThroughputInfo throughputInfo, List<String> assignedPartitions,
       List<String> unassignedPartitions) {
-    throw new NotImplementedException();
+    Validate.notNull(throughputInfo, "null throughputInfo");
+    Validate.notNull(assignedPartitions, "null assignedPartitions");
+    Validate.notNull(unassignedPartitions, "null unassignedPartitions");
+
+    Map<String, PartitionThroughputInfo> throughputMap = throughputInfo.getPartitionInfoMap();
```
vmaheshw, line 48: What happens in the throughput info is not available ? Will it be the callers responsibility in that case? It will be better to add those information.

datastream-server/src/test/resources/partitionThroughput.json
```diff
@@ -213,6 +213,16 @@
       "PeanutButter-7" : "bytesInKB: 8, msgIn:24",
       "PeanutButter-8" : "bytesInKB: 8, msgIn:26",
       "PeanutButter-9" : "bytesInKB: 8, msgIn:26"
+    },
+    "ice-cream" : {
+      "Vanilla-1" : "bytesInKB: 1024, msgIn:100",
+      "Vanilla-2" : "bytesInKB: 1024, msgIn:100",
+      "Vanilla-3" : "bytesInKB: 1024, msgIn:100",
+      "Vanilla-4" : "bytesInKB: 1024, msgIn:100",
```
vmaheshw, line 221: pick a case, where the byteInKB is more than the max supported in one task.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -5,18 +5,32 @@
...
 /**
  * Estimates the minimum number of tasks for a datastream based on per-partition throughput information
  */
 public class LoadBasedTaskCountEstimator {
 
+  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedTaskCountEstimator.class.getName());
+
+  // TODO Make this constant configurable
+  @VisibleForTesting
+  public final static int TASK_CAPACITY_MBPS_DEFAULT = 4;
```
vmaheshw, line 32: since it is visible for testing should you remove the `final` ?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,22 @@
...
+
+    Map<String, PartitionThroughputInfo> throughputMap = throughputInfo.getPartitionInfoMap();
+
+    Set<String> allPartitions = new HashSet<>(assignedPartitions);
+    allPartitions.addAll(unassignedPartitions);
+
+    // total throughput in KB/sec
+    int totalThroughput = throughputMap.entrySet().stream().
+        filter(e -> allPartitions.contains(e.getKey())).mapToInt(e -> e.getValue().getBytesInKBRate()).sum();
+    LOG.debug("Total throughput in all partitions: {}KB/sec", totalThroughput);
```
vmaheshw, line 56: Set this level to INFO, since we don't expect a lot of calls for this.

vmaheshw (changes requested):

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: The field name suggests that it is rate in KB/sec, hence 1024.

jzakaryan: The way I see it, the buffer for growth will be reserved by elastic task mechanism. We already have that there, right? Do we need to bring it here as well?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: I'd prefer to leave it as it is.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: I was thinking about this too. What if throughput info is not available for a subset of partitions? I think it makes sense to just print a warning message in that case. Worst case scenario, the old mechanism for task count estimation will win, but a warning message would be helpful to have.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: That's an interesting point. Do you think that weight should be configurable?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: There's a slight problem here though. If we want to filter the provider results and get only info for partitions with throughput > X, there won't be a way to tell if the throughput info was filtered out or it wasn't available.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
vmaheshw: Yes, weight should be configurable. Please add that as well in this PR.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
vmaheshw: Empty map can be treated as throughput not available. In case of Pinot, you will get success or failure. In case of Brooklin monitor, you can assume and retry.

vmaheshw: Also, Throughput > X is very critical to get the results faster in case of Pinot.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,23 @@
...
+    allPartitions.addAll(unassignedPartitions);
+
+    // total throughput in KB/sec
+    int totalThroughput = throughputMap.entrySet().stream().
+        filter(e -> allPartitions.contains(e.getKey())).mapToInt(e -> e.getValue().getBytesInKBRate()).sum();
+    LOG.info("Total throughput in all partitions: {}KB/sec", totalThroughput);
+
+    int taskCountEstimate = (int) Math.ceil((double) totalThroughput / (TASK_CAPACITY_MBPS_DEFAULT * 1024));
+    taskCountEstimate = Math.min(allPartitions.size(), taskCountEstimate);
+    LOG.debug("Estimated number of tasks required to handle the throughput: {}", taskCountEstimate);
```
vmaheshw, line 60: Make it info, as this is important for us and will not be very frequent, also add assigned and unassigned partition count

datastream-server/src/test/java/com/linkedin/datastream/server/TestLoadBasedTaskCountEstimator.java
```diff
@@ -0,0 +1,85 @@
...
+import com.linkedin.datastream.server.assignment.LoadBasedTaskCountEstimator;
+import com.linkedin.datastream.server.providers.FileBasedPartitionThroughputProvider;
+
+
+/**
+ * Tests for {@link LoadBasedTaskCountEstimator}
+ */
+public class TestLoadBasedTaskCountEstimator {
+  private static final String THROUGHPUT_FILE_NAME = "partitionThroughput.json";
+  FileBasedPartitionThroughputProvider _provider;
```
vmaheshw, line 25: add negative tests for throughput information not present, assigned partitions, unassigned partitions

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,23 @@
    */
   public int getTaskCount(ClusterThroughputInfo throughputInfo, List<String> assignedPartitions,
```
vmaheshw, line 42: One of the major issue I have with task counting using both assigned partition and unassigned partition is, sticky strategy reuses the task object and it will be tricky in this case.

This will not show up as an issue, since we will call the method with assignedPartitions as empty. Do you think, it will make more sense to expose another method `getTaskCount(throughputInfo, unassignedPartitions)` and internally call `getTaskCount(throughputInfo, Collections.emptyList(), unassignedPartitions)`. This way when we later read the code, it will not create unnecessary confusion.

vmaheshw (commented):
Mostly looks good. Please add default weight change and threshold to fill (default can be 90%).

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,23 @@
...
       List<String> unassignedPartitions) {
-    throw new NotImplementedException();
+    Validate.notNull(throughputInfo, "null throughputInfo");
+    Validate.notNull(assignedPartitions, "null assignedPartitions");
+    Validate.notNull(unassignedPartitions, "null unassignedPartitions");
+
```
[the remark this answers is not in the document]
jzakaryan: This will go in `LoadBasedPartitionAssignmentStrategy`

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: I think javadoc and variable naming should take care of that.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -85,18 +94,19 @@ public LoadBasedPartitionAssignmentStrategy(PartitionThroughputProvider throughp
...
         "Zero tasks assigned. Retry leader partition assignment.");
 
+    // Resolving cluster name from datastream group
+    String clusterName = _sourceClusterResolver.getSourceCluster(datastreamPartitions.getDatastreamGroup());
+    ClusterThroughputInfo clusterThroughputInfo = partitionThroughputInfo.get(clusterName);
+
     // TODO Get task count estimate and perform elastic task count validation
     // TODO Get task count estimate based on throughput and pick a winner
-    int maxTaskCount = 0;
+    LoadBasedTaskCountEstimator estimator = new LoadBasedTaskCountEstimator(_taskCapacityMBps, _taskCapacityUtilizationPct);
```
jzakaryan, line 103: This were added here just to make the code compile. Final implementation of `LoadBasedPartitionAssignmentStrategy` will be published in a separate PR.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +41,23 @@
...
+    Map<String, PartitionThroughputInfo> throughputMap = throughputInfo.getPartitionInfoMap();
+
+    Set<String> allPartitions = new HashSet<>(assignedPartitions);
+    allPartitions.addAll(unassignedPartitions);
+
+    // total throughput in KB/sec
+    int totalThroughput = throughputMap.entrySet().stream().
+        filter(e -> allPartitions.contains(e.getKey())).mapToInt(e -> e.getValue().getBytesInKBRate()).sum();
+    LOG.info("Total throughput in all partitions: {}KB/sec", totalThroughput);
```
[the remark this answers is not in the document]
jzakaryan: It's added in the previous log message.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -40,8 +41,13 @@
...
+  private static final int TASK_CAPACITY_MBPS_DEFAULT = 4;
+  private static final int TASK_CAPACITY_UTILIZATION_PCT_DEFAULT = 90;
+
   private final PartitionThroughputProvider _throughputProvider;
   private final DatastreamSourceClusterResolver _sourceClusterResolver;
+  private final int _taskCapacityMBps;
+  private final int _taskCapacityUtilizationPct;
 
 
   /**
```
vmaheshw, line 53: javadocs missing.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -85,18 +94,19 @@ public LoadBasedPartitionAssignmentStrategy(PartitionThroughputProvider throughp
     Validate.isTrue(currentAssignment.size() > 0,
         "Zero tasks assigned. Retry leader partition assignment.");
 
+    // Resolving cluster name from datastream group
+    String clusterName = _sourceClusterResolver.getSourceCluster(datastreamPartitions.getDatastreamGroup());
+    ClusterThroughputInfo clusterThroughputInfo = partitionThroughputInfo.get(clusterName);
```
vmaheshw, line 99: Just wondering, if fetchPartitionThroughputInfo object should have cluster information. Since you have the retry logic at the earlier step (`fetchPartitionThroughputInfo`, the implementation for Brooklin-monitor/Pinot will be tricky, as you will need fabric/cluster information.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -5,17 +5,40 @@
...
 
 import com.linkedin.datastream.server.ClusterThroughputInfo;
+import com.linkedin.datastream.server.PartitionThroughputInfo;
 
 
 /**
  * Estimates the minimum number of tasks for a datastream based on per-partition throughput information
  */
 public class LoadBasedTaskCountEstimator {
+  private final static int BYTES_IN_KB_RATE_DEFAULT = 5;
```
vmaheshw, line 25: Make these fields configurable.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -5,17 +5,40 @@
...
+   * Creates an instance of {@link LoadBasedTaskCountEstimator}
+   * @param taskCapacityMBps Task capacity in MB/sec
+   * @param taskCapacityUtilizationPct Task capacity utilization percentage
+   */
+  public LoadBasedTaskCountEstimator(int taskCapacityMBps, int taskCapacityUtilizationPct) {
+    _taskCapacityMBps = taskCapacityMBps;
+    _taskCapacityUtilizationPct = taskCapacityUtilizationPct;
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedTaskCountEstimator.class.getName());
```
vmaheshw, line 41: move this before the constructor.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +50,26 @@
    */
   public int getTaskCount(ClusterThroughputInfo throughputInfo, List<String> assignedPartitions,
       List<String> unassignedPartitions) {
-    throw new NotImplementedException();
+    Validate.notNull(throughputInfo, "null throughputInfo");
+    Validate.notNull(assignedPartitions, "null assignedPartitions");
+    Validate.notNull(unassignedPartitions, "null unassignedPartitions");
```
vmaheshw, line 55: One good validation will be to check if `assignedPartitions` and `unassignedPartitions` have any intersection. This is critical as we don't want two tasks to have the same partition.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +50,26 @@
...
+
+    Map<String, PartitionThroughputInfo> throughputMap = throughputInfo.getPartitionInfoMap();
+
+    Set<String> allPartitions = new HashSet<>(assignedPartitions);
+    allPartitions.addAll(unassignedPartitions);
+
+    // total throughput in KB/sec
+    int totalThroughput = allPartitions.stream().map(p ->
+        throughputMap.getOrDefault(p, new PartitionThroughputInfo(BYTES_IN_KB_RATE_DEFAULT, MESSAGES_IN_RATE_DEFAULT, p))).
```
vmaheshw, line 64: Will it make more sense to create one default PartitionThroughputInfo object, rather than N objects, since we don't it for long.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +50,26 @@
...
+    Validate.notNull(assignedPartitions, "null assignedPartitions");
+    Validate.notNull(unassignedPartitions, "null unassignedPartitions");
+
+    Map<String, PartitionThroughputInfo> throughputMap = throughputInfo.getPartitionInfoMap();
+
+    Set<String> allPartitions = new HashSet<>(assignedPartitions);
+    allPartitions.addAll(unassignedPartitions);
+
+    // total throughput in KB/sec
+    int totalThroughput = allPartitions.stream().map(p ->
```
vmaheshw, line 63: Change this for better readability. (One operator per line)

> int totalThroughput = allPartitions.stream()
        .map(p -> throughputMap.getOrDefault(p, new PartitionThroughputInfo(BYTES_IN_KB_RATE_DEFAULT, MESSAGES_IN_RATE_DEFAULT, p)))
        .mapToInt(PartitionThroughputInfo::getBytesInKBRate).sum();

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +50,26 @@
...
+        throughputMap.getOrDefault(p, new PartitionThroughputInfo(BYTES_IN_KB_RATE_DEFAULT, MESSAGES_IN_RATE_DEFAULT, p))).
+        mapToInt(PartitionThroughputInfo::getBytesInKBRate).sum();
+    LOG.info("Total throughput in all {} partitions: {}KB/sec", allPartitions.size(), totalThroughput);
+
+    double taskCapacityUtilizationCoefficient = _taskCapacityUtilizationPct / 100.0;
+    int taskCountEstimate = (int) Math.ceil((double) totalThroughput /
+        (_taskCapacityMBps * 1024 * taskCapacityUtilizationCoefficient));
```
vmaheshw, line 70: nit: if you save `_taskCapacityMBps * 1024 * taskCapacityUtilizationCoefficient` calculation, you can get rid of `_taskCapacityMBps`, `taskCapacityUtilizationCoefficient` and extra calculation.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +50,26 @@
    */
   public int getTaskCount(ClusterThroughputInfo throughputInfo, List<String> assignedPartitions,
       List<String> unassignedPartitions) {
-    throw new NotImplementedException();
+    Validate.notNull(throughputInfo, "null throughputInfo");
```
vmaheshw, line 53: Debug log to dump all the three fields for testing purpose will be very useful.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/PartitionAssignmentStrategyConfig.java
```diff
@@ -54,6 +58,8 @@ public PartitionAssignmentStrategyConfig(Properties config) {
     int cfgMaxParitionsPerTask = props.getInt(CFG_MAX_PARTITION_PER_TASK, 0);
     int cfgPartitionsPerTask = props.getInt(CFG_PARTITIONS_PER_TASK, 0);
     int cfgPartitionFullnessThresholdPct = props.getIntInRange(CFG_PARTITION_FULLNESS_THRESHOLD_PCT, 0, 0, 100);
+    int cfgTaskCapacityMBps = props.getInt(CFG_TASK_CAPACITY_MBPS, 0);
```
vmaheshw, line 61: Just wondering, if we should get the default values to this class and get rid of optional.

The counter reasoning to this change request can be: in the future, we may have some class that does not need the default value. But, we can move to Optional at that time. Right now, even though we pass optional, the constructor of the class, sets the default value if Optional is empty.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: I don't think there's need, @vmaheshw. I don't think it's something we will want to tune.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: This will be done in the strategy class itself.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: This will be done in the strategy class.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/PartitionAssignmentStrategyConfig.java
jzakaryan: I tried to stick to the way other configs were implemented. If we want to change this, we might want to address the older ones too to keep things consistent.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: You're right. We have the name for each partition if we do it in the loop, but then again, we're not using that name.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
jzakaryan: Will be added in separate PR.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
jzakaryan: `ClusterThroughputInfo` already has the `clusterName`. Let's take this offline, maybe I'm missing something.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: Yeah but then I'll need to name that variable 😄

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
jzakaryan: As per our discussion, this will go out in a separate PR.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -27,6 +54,30 @@
...
+    Validate.notNull(assignedPartitions, "null assignedPartitions");
+    Validate.notNull(unassignedPartitions, "null unassignedPartitions");
+    LOG.info("Assigned partitions: {}", assignedPartitions);
```
vmaheshw, line 60: We need throughput info as well. There is no easy way for us to get this info otherwise.

vmaheshw (approved):
Other than 1 last comment, everything else LGTM.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
jzakaryan: @vmaheshw let me think about this. I don't want to dump this in the log file. It can be huge.

shrinandthakkar (approved):

PR #832, opened 2021-06-04
Fix encode Avro Indexed Record for SpecificRecordBase

vmaheshw:
SpecificRecordBase extends both GenericRecord and SpecificRecord. Any record extending SpecificRecord needs to go via SpecificDatumWriter for encoding. According to the current logic, it assumes that the record will be either GenericRecord or SpecificRecord. 

To solve this problem, swapping the check and in the future will integrate with AvroCompatibilityHelper.


radai-rosenblatt (approved):
+1 for great justice

DEEPTHIKORAT (approved):

PR #840, opened 2021-06-25
Implemented metrics for LoadBasedPartitionAssignmentStrategy

jzakaryan:
Implemented the following metrics for `LoadBasedPartitionAssignmentStrategy`:

- Minimum partitions across tasks (per datastream)
- Maximum partitions across tasks (per datastream)
- Throughput information fetch rate

the contribution guidelines.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
@@ -6,7 +6,6 @@
 package com.linkedin.datastream.server.assignment;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -41,8 +40,8 @@ import com.linkedin.datastream.server.PartitionThroughputInfo;
  * Performs partition assignment based on partition throughput information
  */
 public class LoadBasedPartitionAssigner implements MetricsAware {
-  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedPartitionAssignmentStrategy.class.getName());
-  private static final String CLASS_NAME = LoadBasedPartitionAssignmentStrategy.class.getSimpleName();
+  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedPartitionAssigner.class.getName());
+  private static final String CLASS_NAME = LoadBasedPartitionAssigner.class.getSimpleName();
   private static final DynamicMetricsManager DYNAMIC_METRICS_MANAGER = DynamicMetricsManager.getInstance();
   private static final String MIN_PARTITIONS_ACROSS_TASKS = "minPartitionsAcrossTasks";
   private static final String MAX_PARTITIONS_ACROSS_TASKS = "maxPartitionsAcrossTasks";
@@ -228,9 +227,13 @@ public class LoadBasedPartitionAssigner implements MetricsAware {
    */
   @Override
   public List<BrooklinMetricInfo> getMetricInfos() {
+    List<BrooklinMetricInfo> metricInfos = new ArrayList<>();
     String prefix = CLASS_NAME + MetricsAware.KEY_REGEX;
-    return Arrays.asList(new BrooklinGaugeInfo(prefix + MIN_PARTITIONS_ACROSS_TASKS),
-        new BrooklinGaugeInfo(prefix + MAX_PARTITIONS_ACROSS_TASKS));
+
+    metricInfos.add(new BrooklinGaugeInfo(prefix + MIN_PARTITIONS_ACROSS_TASKS));
+    metricInfos.add(new BrooklinGaugeInfo(prefix + MAX_PARTITIONS_ACROSS_TASKS));
+
+    return Collections.unmodifiableList(metricInfos);
   }
 
   void cleanupMetrics() {
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
@@ -14,6 +14,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.PriorityQueue;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -23,6 +24,7 @@ import com.linkedin.datastream.server.ClusterThroughputInfo;
 import com.linkedin.datastream.server.DatastreamGroupPartitionsMetadata;
 import com.linkedin.datastream.server.DatastreamTask;
 import com.linkedin.datastream.server.DatastreamTaskImpl;
+import com.linkedin.datastream.server.Pair;
 import com.linkedin.datastream.server.PartitionThroughputInfo;
 
 
@@ -40,11 +42,11 @@ public class LoadBasedPartitionAssigner {
    * @param currentAssignment Current assignment
    * @param unassignedPartitions Unassigned partitions
    * @param partitionMetadata Partition metadata
-   * @return New assignment
+   * @return A pair where the first value is the assignment, and the second value is the assignment stats
    */
-  public Map<String, Set<DatastreamTask>> assignPartitions(ClusterThroughputInfo throughputInfo,
-      Map<String, Set<DatastreamTask>> currentAssignment, List<String> unassignedPartitions,
-      DatastreamGroupPartitionsMetadata partitionMetadata, int maxPartitionsPerTask) {
+  public Pair<Map<String, Set<DatastreamTask>>, PartitionAssignmentStats> assignPartitions(
+      ClusterThroughputInfo throughputInfo, Map<String, Set<DatastreamTask>> currentAssignment,
+      List<String> unassignedPartitions, DatastreamGroupPartitionsMetadata partitionMetadata, int maxPartitionsPerTask) {
     String datastreamGroupName = partitionMetadata.getDatastreamGroup().getName();
     Map<String, PartitionThroughputInfo> partitionInfoMap = throughputInfo.getPartitionInfoMap();
     Set<String> tasksWithChangedPartition = new HashSet<>();
@@ -130,6 +132,8 @@ public class LoadBasedPartitionAssigner {
       index = (index + 1) % tasks.size();
     }
 
+    AtomicInteger minPartitionsAcrossTasks = new AtomicInteger(Integer.MAX_VALUE);
+    AtomicInteger maxPartitionsAcrossTasks = new AtomicInteger(0);
     // build the new assignment using the new partitions for the affected datastream's tasks
     Map<String, Set<DatastreamTask>> newAssignments = new HashMap<>();
     currentAssignment.keySet().forEach(instance -> {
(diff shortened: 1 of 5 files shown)


vmaheshw (changes requested):
My recommendation is to not pass the metrics from the assigner class. Instead emit the metric from the assigner class itself and expose additional methods like `getMetricsInfos()` that you can use. This way all the metric logic will remain in one place and the code will look cleaner.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -29,7 +40,15 @@
 /**
  * Performs partition assignment based on partition throughput information
  */
-public class LoadBasedPartitionAssigner {
+public class LoadBasedPartitionAssigner implements MetricsAware {
+  private static final Logger LOG = LoggerFactory.getLogger(LoadBasedPartitionAssignmentStrategy.class.getName());
```
jzakaryan, line 44: The class name for logging and metrics is set to `LoadBasedPartitionAssignmentStrategy` on purpose. It will be easier to filter for these along with logs and metrics from the main strategy class if they share the same namespace.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java
```diff
@@ -166,4 +176,29 @@ private ClusterThroughputInfo fetchPartitionThroughputInfo(DatastreamGroup datas
...
+   */
+  @Override
+  public void cleanupStrategy() {
+    _assigner.cleanupMetrics();
+    super.cleanupStrategy();
+  }
+
+  @Override
+  protected void unregisterMetrics(String datastream) {
+    _assigner.unregisterMetricsForDatastream(datastream);
```
jzakaryan, line 202: This method still has to be here, because there's a path for cleaning metrics in `StickyPartitionAssignmentStrategy` which calls this method.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -29,7 +40,15 @@
...
+  private static final String CLASS_NAME = LoadBasedPartitionAssignmentStrategy.class.getSimpleName();
```
vmaheshw, line 45: I'm not sure if you should use a different strategy class name here. This removes the flexibiilty of reuse of this class in the new strategy. This same thing applies to the LOGGER as well. Since the changes are in open-source, there is always a possibility of new strategy.

jzakaryan: Let's worry about that when we have another strategy that uses `LoadBasedPartitionAssigner`. At the moment it's only used by `LoadBasedPartitionAssignmentStrategy`.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -170,4 +203,79 @@ int findTaskWithRoomForAPartition(List<String> tasks, Map<String, Set<String>> p
...
+        maxPartitionsAcrossTasksSupplier);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public List<BrooklinMetricInfo> getMetricInfos() {
+    String prefix = CLASS_NAME + MetricsAware.KEY_REGEX;
+    return Arrays.asList(new BrooklinGaugeInfo(prefix + MIN_PARTITIONS_ACROSS_TASKS),
```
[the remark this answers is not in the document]
jzakaryan: Why is it better than what I'm doing? The strategy class exposes it as an unmodifiable list anyway (if that's the concern).

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -144,6 +169,14 @@
       newAssignments.put(instance, newTasks);
     });
 
+    // update metrics
+    PartitionAssignmentStats stats = new PartitionAssignmentStats(minPartitionsAcrossTasks.get(),
```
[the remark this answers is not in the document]
jzakaryan: I think it makes the code cleaner. If I remove it, I'll have to maintain two separate maps for min/max partitions across tasks per datastream. This is more extensible as we can add more stats and add more per-datastream metrics in the future (e.g. the maxThroughputAcrossTasks we discussed earlier).

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
shrinandthakkar: Both of these cases are easily achieved by tags in metrics while graphing them, but I don't think ingraphs have tags as of now. But for this one, I would also prefer to keep the class name as it is.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -170,4 +203,79 @@ int findTaskWithRoomForAPartition(List<String> tasks, Map<String, Set<String>> p
...
+    if (!_partitionAssignmentStatsMap.containsKey(datastream)) {
+      registerLoadBasedPartitionAssignmentMetrics(datastream);
+    }
+    _partitionAssignmentStatsMap.put(datastream, stats);
+  }
+
+  private void registerLoadBasedPartitionAssignmentMetrics(String datastream) {
+    Supplier<Integer> minPartitionsAcrossTasksSupplier = () -> _partitionAssignmentStatsMap
+        .getOrDefault(datastream, PartitionAssignmentStats.DEFAULT).getMinPartitionsAcrossTasks();
+    DYNAMIC_METRICS_MANAGER.registerGauge(CLASS_NAME, datastream, MIN_PARTITIONS_ACROSS_TASKS,
```
shrinandthakkar, line 217: are these configurable? like the interval at which they emit metrics?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -170,4 +203,79 @@ int findTaskWithRoomForAPartition(List<String> tasks, Map<String, Set<String>> p
...
+   * Encapsulates assignment metrics for a single datastream group
+   */
+  private static class PartitionAssignmentStats {
+    private final int _minPartitionsAcrossTasks;
+    private final int _maxPartitionsAcrossTasks;
+
+    public static final PartitionAssignmentStats DEFAULT = new PartitionAssignmentStats(0, 0);
+
+    /**
+     * Creates an instance of {@link PartitionAssignmentStats}
```
shrinandthakkar, line 256: can we add a comment here to say that this class is for 'per-datastream metrics'

shrinandthakkar (commented):
Also, in the case when a datastream is deleted, how would this work? Since we are never deleting the record for a datastream from the _partitionAssignmentStatsMap.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
vmaheshw: @jzakaryan I thought about this use-case again and I could think of bunch of alternatives:

a) Pass Logger and Metric Prefix in the class constructor and use it. There are other instances in this repo where Logger is passed instead of created. I can think of this to be the main reason.

b) If you want the stronger stitching, then you can move this class as inner class of the strategy. Based on our offline discussion, you have less preference for this.

So, option (a) will unify both of our thoughts on this.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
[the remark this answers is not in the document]
vmaheshw: The major problem I can see with this future extension is, if there is another method created and that method calculates some stats, then either you will have to define all the getter and setters and make the fields non-final. Suppose we want to create any alert based metric, then all the other fields also need to be set. To overcome this, you may have to optional to figure out which fields are set and it will be an overkill.

You can have separate methods for each metric for create/Update the metric and just call those method, just like it is done in other classes in this repo.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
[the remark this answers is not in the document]
vmaheshw: Yes, Collections.unmodifiableList was the concern. I agree that the strategy class can take care of this. But, since this class is public and can be reused in the future, it is a better practice to ensure that we do most of the damage control in the class Library itself rather than relying on the callers to take care of it. We have seen in the past where the list was not suppose to be modified and the caller modifies it and the issue was super difficult to catch.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
vmaheshw: It is a guage. So, it gets emitted every minute. I will share the metrics doc with you for reading.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
jzakaryan: I think the javadoc on the class already conveys this:
`Encapsulates assignment metrics for a single datastream group`

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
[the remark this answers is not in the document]
jzakaryan: I'm not sure what you mean by an alert-based metric, but from the description it seems like it doesn't fit with what I see as partition assignment stats. Semantically they might be incompatible, and we don't have to put it in the class. I still see value in keeping `PartitionAssignmentStats` around, for the reasons I mentioned above.

shrinandthakkar (approved):

shrinandthakkar:
> Also, in the case when a datastream is deleted, how would this work? Since we are never deleting the record for a datastream from the _partitionAssignmentStatsMap.

talked with jhora offline and he said, "on datastream delete/stop, the metric gets unregistered i.e. the values for deleted/stopped metrics don’t get accessed and emitted"

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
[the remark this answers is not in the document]
vmaheshw: I mean if you have to emit a new metric which is in the exception path (and you don't calculate the other stats), then reusing this stat will be an issue and will require refactoring, V/s addition of new metric not impacting any other code change.

vmaheshw (approved):
Approving with the note that there might be refactoring required for stats in the future based on the requirement.

PR #845, opened 2021-07-29
Handle leaking TP exception in handleAssignmentChange

surajkn:
During handleAssignmentChange, while initializing task we were leaking a possible
exception while creating TransportProvider object. This patch fixes it by issuing
a retry on failure.

the contribution guidelines.

### DIFF
--- a/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
+++ b/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
@@ -511,6 +511,58 @@ public class TestCoordinator {
   }
 
   @Test
+  public void testHandleAssignmentChangeTransientFailure() throws Exception {
+    String testCluster = "testCoordinationSmoke";
+    String testConnectorType = "testConnectorType";
+    String datastreamName1 = "datastream1";
+
+    DummyTransportProviderAdminFactory dummyTransportProviderAdminFactory = new DummyTransportProviderAdminFactory(
+        false, true, false);
+    Coordinator instance1 = createCoordinator(_zkConnectionString, testCluster, new Properties(),
+        dummyTransportProviderAdminFactory);
+    TestHookConnector connector1 = new TestHookConnector("connector1", testConnectorType);
+    instance1.addConnector(testConnectorType, connector1, new BroadcastStrategy(Optional.empty()), false,
+        new SourceBasedDeduper(), null);
+    instance1.start();
+
+    ZkClient zkClient = new ZkClient(_zkConnectionString);
+    DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, testConnectorType, datastreamName1);
+    //verify the assignment
+    assertConnectorAssignment(connector1, WAIT_TIMEOUT_MS, datastreamName1);
+    String instance1Path = KeyBuilder.instanceAssignments(testCluster, instance1.getInstanceName());
+    Assert.assertNotEquals(zkClient.getChildren(instance1Path).size(), 0);
+
+    instance1.stop();
+    instance1.getDatastreamCache().getZkclient().close();
+    zkClient.close();
+  }
+
+  @Test
+  public void testHandleAssignmentChangeFailure() throws Exception {
+    String testCluster = "testCoordinationSmoke";
+    String testConnectorType = "testConnectorType";
+    String datastreamName1 = "datastream1";
+
+    DummyTransportProviderAdminFactory dummyTransportProviderAdminFactory = new DummyTransportProviderAdminFactory(
+        false, false, true);
+    Coordinator instance1 = createCoordinator(_zkConnectionString, testCluster, new Properties(),
+        dummyTransportProviderAdminFactory);
+    TestHookConnector connector1 = new TestHookConnector("connector1", testConnectorType);
+    instance1.addConnector(testConnectorType, connector1, new BroadcastStrategy(Optional.empty()), false,
+        new SourceBasedDeduper(), null);
+    instance1.start();
+
+    ZkClient zkClient = new ZkClient(_zkConnectionString);
+    DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, testConnectorType, datastreamName1);
+    String instance1Path = KeyBuilder.instanceAssignments(testCluster, instance1.getInstanceName());
+    Assert.assertEquals(zkClient.getChildren(instance1Path).size(), 0);
+
+    instance1.stop();
+    instance1.getDatastreamCache().getZkclient().close();
+    zkClient.close();
+  }
+
+  @Test
   public void testStopAndResumeDatastream() throws Exception {
     String testCluster = "testCoordinationSmoke";
     String testConnectorType = "testConnectorType";
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -12,6 +12,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -27,6 +28,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BooleanSupplier;
 import java.util.function.Function;
@@ -81,6 +83,8 @@ import static com.linkedin.datastream.common.DatastreamMetadataConstants.TTL_MS;
 import static com.linkedin.datastream.common.DatastreamUtils.hasValidDestination;
 import static com.linkedin.datastream.common.DatastreamUtils.isReuseAllowed;
 import static com.linkedin.datastream.server.CoordinatorEvent.EventType;
+import static com.linkedin.datastream.server.CoordinatorEvent.EventType.HANDLE_ASSIGNMENT_CHANGE;
+import static com.linkedin.datastream.server.CoordinatorEvent.EventType.HANDLE_DATASTREAM_CHANGE_WITH_UPDATE;
 
 
 /**
@@ -181,6 +185,10 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
 
   // Currently assigned datastream tasks by taskName
   private final Map<String, DatastreamTask> _assignedDatastreamTasks = new ConcurrentHashMap<>();
+  // Maintain a count of onAssignmentChange retries in case of task initialization/submission failure.
+  // If count reaches maximum threshold do not queue the event further. Reset the count if event raised
+  // from any other path.
+  private AtomicInteger _assignmentRetryCount = new AtomicInteger(0);
 
   // One coordinator heartbeat per minute, heartbeat helps detect dead/live-lock
   // where no events can be handled if coordinator locks up. This can happen because
@@ -273,7 +281,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
 
     // now that instance is started, make sure it doesn't miss any assignment created during
     // the slow startup
-    _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
+    queueHandleAssignmentOrDatastreamChangeEvent(CoordinatorEvent.createHandleAssignmentChangeEvent(), true);
 
     // Queue up one heartbeat per period with a initial delay of 3 periods
     _executor.scheduleAtFixedRate(() -> _eventQueue.put(CoordinatorEvent.HEARTBEAT_EVENT),
@@ -435,7 +443,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
         }
       });
     }
-    _eventQueue.put(CoordinatorEvent.createHandleDatastreamChangeEvent());
+    queueHandleAssignmentOrDatastreamChangeEvent(CoordinatorEvent.createHandleDatastreamChangeEvent(), true);
     _log.info("Coordinator::onDatastreamUpdate completed successfully");
   }
 
@@ -572,8 +580,42 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   @Override
   public void onAssignmentChange() {
     _log.info("Coordinator::onAssignmentChange is called");
-    _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
+    queueHandleAssignmentOrDatastreamChangeEvent(CoordinatorEvent.createHandleAssignmentChangeEvent(), true);
     _log.info("Coordinator::onAssignmentChange completed successfully");
+  }
+
+  private int getAssignmentTaskCount(Map<String, List<DatastreamTask>> assignment) {
+    return assignment.values().stream().mapToInt(List::size).sum();
+  }
+
+  private void queueHandleAssignmentOrDatastreamChangeEvent(CoordinatorEvent event, boolean isReset) {
+    _eventQueue.put(event);
+    if (isReset) {
+      _assignmentRetryCount.set(0);
+    } else {
+      _assignmentRetryCount.incrementAndGet();
+    }
+  }
+
+  private void retryHandleAssignmentChange(boolean isDatastreamUpdate) {
+    boolean queueEvent = true;
+    if (_assignmentRetryCount.compareAndSet(_config.getMaxAssignmentRetryCount(), 0)) {
+      _log.error(
+          "onAssignmentChange retries reached threshold of {}. Not queuing further retry on onAssignmentChange event.",
+          _config.getMaxAssignmentRetryCount());
+      queueEvent = false;
+    }
+
+    EventType meter = isDatastreamUpdate ? HANDLE_DATASTREAM_CHANGE_WITH_UPDATE : HANDLE_ASSIGNMENT_CHANGE;
+    _log.warn("Updating metric for event " + meter);
+    _metrics.updateKeyedMeter(CoordinatorMetrics.getKeyedMeter(meter), 1);
+
+    if (queueEvent) {
+      _log.warn("Queuing onAssignmentChange event");
+      CoordinatorEvent event = isDatastreamUpdate ? CoordinatorEvent.createHandleDatastreamChangeEvent() :
+          CoordinatorEvent.createHandleAssignmentChangeEvent();
+      queueHandleAssignmentOrDatastreamChangeEvent(event, false);
+    }
   }
 
   private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutException {
@@ -598,6 +640,8 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
         currentAssignment.get(connectorType).add(task);
       }
     });
+
+    int totalTasks = getAssignmentTaskCount(currentAssignment);
 
     _log.info(printAssignmentByType(currentAssignment));
 
@@ -634,6 +678,8 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
         .filter(Objects::nonNull)
         .collect(Collectors.toList()));
 
+    int submittedTasks = getAssignmentTaskCount(currentAssignment);
+
     // Wait till all the futures are complete or timeout.
     Instant start = Instant.now();
     try {
@@ -641,12 +687,8 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
     } catch (TimeoutException e) {
       // if it's timeout then we will retry
       _log.warn("Timeout when doing the assignment", e);
-      if (isDatastreamUpdate) {
-        _eventQueue.put(CoordinatorEvent.createHandleDatastreamChangeEvent());
-      } else {
-        _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
-      }
-      throw e;
+      retryHandleAssignmentChange(isDatastreamUpdate);
+      return;
     } catch (InterruptedException e) {
       _log.warn("onAssignmentChange call got interrupted", e);
     } finally {
@@ -659,6 +701,13 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
         .stream()
         .flatMap(Collection::stream)
         .collect(Collectors.toMap(DatastreamTask::getDatastreamTaskName, Function.identity())));
+
+    if ((totalTasks - submittedTasks) > 0) {
+      _log.warn("Failed to submit {} tasks from currentAssignment. Queueing onAssignmentChange event again",
+          totalTasks - submittedTasks);
+      // Update the metric and queue the event only once for all the tasks that failed to be initialized
+      retryHandleAssignmentChange(isDatastreamUpdate);
+    }
 
     _log.info("END: Coordinator::handleAssignmentChange, Duration: {} milliseconds",
         System.currentTimeMillis() - startAt);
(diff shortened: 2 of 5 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -712,8 +728,12 @@ private DatastreamTask getDatastreamTask(String taskName) {
           err += " Queuing up a new onAssignmentChange event for retry.";
           _eventQueue.put(CoordinatorEvent.createHandleInstanceErrorEvent(ExceptionUtils.getRootCauseMessage(ex)));
           if (isDatastreamUpdate) {
+            _metrics.updateKeyedMeter(CoordinatorMetrics.getKeyedMeter(
+                CoordinatorEvent.EventType.HANDLE_DATASTREAM_CHANGE_WITH_UPDATE), 1);
```
vmaheshw, line 732: This specific metric is taken care in the generic exception catch of `handleEvents`. If you see value in the number of retries, then you will have to justify the reason and add another metric.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -690,7 +705,8 @@ private DatastreamTask getDatastreamTask(String taskName) {
     if (isDatastreamUpdate || !addedTasks.isEmpty() || !removedTasks.isEmpty()) {
       // Populate the event producers before calling the connector with the list of tasks.
       addedTasks.stream().filter(t -> t.getEventProducer() == null).forEach(this::initializeTask);
-      return submitAssignment(connectorType, assignment, isDatastreamUpdate, connector, removedTasks, retryAndSaveError);
+      return submitAssignment(connectorType, assignment, isDatastreamUpdate, connector, removedTasks,
```
vmaheshw, line 708: I don't see any change here.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -605,22 +605,23 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
...
-    deactivated.removeAll(newConnectorList);
-    List<Future<Boolean>> assignmentChangeFutures = deactivated.stream()
-        .map(connectorType -> dispatchAssignmentChangeIfNeeded(connectorType, new ArrayList<>(), isDatastreamUpdate, true))
-        .filter(Objects::nonNull)
-        .collect(Collectors.toList());
+    List<Future<Boolean>> assignmentChangeFutures = new ArrayList<>();
+    try {
+      deactivated.removeAll(newConnectorList);
+      assignmentChangeFutures = deactivated.stream()
+          .map(connectorType -> dispatchAssignmentChangeIfNeeded(connectorType, new ArrayList<>(), isDatastreamUpdate, true))
```
vmaheshw, line 612: One thing to understand is, for the exception in `initializeTask`, it is possible that there is one old datastream or one datastream having issue. Based on this logic, even if one task is bad, or the host cannot process one task, the other tasks will also be not attempted. You may want to check the initializeTask loop as well.

vmaheshw: Basically, error in the addition of one task, should not impact the creation/removal of other tasks. So, the next retry will only attempt the task it was not able to start in the previous attempt.

vmaheshw (changes requested):

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: Got it, that makes sense. So basically its ok to process the new assignments partially and it doesn't have to be all or nothing. But even in that case, if even a single task processing fails we still need to queue in another handle assignment event correct ?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: Looks like I split this line into two lines by mistake, will fix it.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: Based on your earlier comments. I think the value is in knowing the number of tasks that failed to be processed. But probably it will help to have some additional context as well like which datastream or connector the task belonged to?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -606,7 +609,7 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
 
     List<String> deactivated = new ArrayList<>(oldConnectorList);
     deactivated.removeAll(newConnectorList);
-    List<Future<Boolean>> assignmentChangeFutures = deactivated.stream()
+    List<Future<Boolean>> assignmentChangeFutures = assignmentChangeFutures = deactivated.stream()
```
surajkn, line 612: I will fix this in a following update to this patch along with additional review comments.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -737,19 +743,33 @@ private void uninitializeTasks(List<DatastreamTask> tasks) {
...
+      TransportProvider transportProvider = tpAdmin.assignTransportProvider(task);
+      EventProducer producer = new EventProducer(task, transportProvider, _cpProvider, _eventProducerConfig, customCheckpointing);
 
-    taskImpl.setEventProducer(producer);
-    Map<Integer, String> checkpoints = producer.loadCheckpoints(task);
-    taskImpl.setCheckpoints(checkpoints);
+      taskImpl.setEventProducer(producer);
+      Map<Integer, String> checkpoints = producer.loadCheckpoints(task);
+      taskImpl.setCheckpoints(checkpoints);
+    } catch (Exception e) {
```
surajkn, line 759: In the cases reported so far the specific exception type is "KafkaException" but I have a generic catch all exception here because all the reported cases were for LiKafkaTransportProvider and I wasn't sure if all transport providers would throw the same error. So to make it generic I have a catch all exception here, pls suggest the appropriate exception type is this is not correct.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -689,7 +694,8 @@ private DatastreamTask getDatastreamTask(String taskName) {
 
     if (isDatastreamUpdate || !addedTasks.isEmpty() || !removedTasks.isEmpty()) {
       // Populate the event producers before calling the connector with the list of tasks.
-      addedTasks.stream().filter(t -> t.getEventProducer() == null).forEach(this::initializeTask);
+      addedTasks.stream().filter(t -> t.getEventProducer() == null).forEach(
```
vmaheshw, line 697: nit: You can move each operator to a separate line for better readability

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -689,7 +694,8 @@ private DatastreamTask getDatastreamTask(String taskName) {
...
+          task -> initializeTask(task, isDatastreamUpdate, retryAndSaveError));
       return submitAssignment(connectorType, assignment, isDatastreamUpdate, connector, removedTasks, retryAndSaveError);
```
vmaheshw, line 699: `submitAssignment` will get called with the assignment list containing the failedTasks. You can trim the list inside the `submitAssignment` method.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -181,6 +181,9 @@
   // Currently assigned datastream tasks by taskName
   private final Map<String, DatastreamTask> _assignedDatastreamTasks = new ConcurrentHashMap<>();
 
+  // Failed datastream task assignments by taskName
+  private final Map<String, DatastreamTask> _failedDatastreamTasks = new ConcurrentHashMap<>();
```
vmaheshw, line 185: You can change this to Set, since you are not using the value portion anyways.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -643,6 +646,8 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
         .stream()
         .flatMap(Collection::stream)
         .collect(Collectors.toMap(DatastreamTask::getDatastreamTaskName, Function.identity())));
+    _assignedDatastreamTasks.keySet().removeAll(_failedDatastreamTasks.keySet());
```
vmaheshw, line 649: what if this code does not get executed because of some exception. and in the next pass the task initialization passes. You will remove the entry from the `_assignedDatastreamTasks` list.

How long do you want to persist the list of `_failedDatastreamTasks`? Plz add it in the object description, so that we can evaluate that this object state is cleared after that.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: This is a good start, but evaluate the `InterruptedException` path as well. I think if the thread gets interrupted then we want to stop all the processing asap. 

Plz check CoordinatorEventProcessor code and the `handleEvent` call.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -737,19 +743,33 @@ private void uninitializeTasks(List<DatastreamTask> tasks) {
...
+      EventProducer producer = new EventProducer(task, transportProvider, _cpProvider, _eventProducerConfig, customCheckpointing);
 
-    taskImpl.setEventProducer(producer);
-    Map<Integer, String> checkpoints = producer.loadCheckpoints(task);
-    taskImpl.setCheckpoints(checkpoints);
+      taskImpl.setEventProducer(producer);
+      Map<Integer, String> checkpoints = producer.loadCheckpoints(task);
+      taskImpl.setCheckpoints(checkpoints);
+    } catch (Exception e) {
+      String err = String.format("Initializing task failed.");
```
vmaheshw, line 760: nit: String.format is not really required.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -737,19 +743,33 @@ private void uninitializeTasks(List<DatastreamTask> tasks) {
...
+      _failedDatastreamTasks.put(task.getDatastreamTaskName(), task);
+      if (retryAndSaveError) {
+        err += " Queuing up a new onAssignmentChange event for retry.";
+        _eventQueue.put(CoordinatorEvent.createHandleInstanceErrorEvent(ExceptionUtils.getRootCauseMessage(e)));
+        if (isDatastreamUpdate) {
+          _eventQueue.put(CoordinatorEvent.createHandleDatastreamChangeEvent());
+        } else {
+          _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
+        }
+      }
```
vmaheshw, line 770: This looks like duplicate code that can be refactored.

datastream-server/src/main/java/com/linkedin/datastream/server/DummyTransportProviderAdminFactory.java
```diff
@@ -28,6 +28,7 @@
   public static final String PROVIDER_NAME = "default";
 
   private final boolean _throwOnSend;
+  private boolean _failTransportProvider;
```
vmaheshw, line 31: Looks like you want to fail the transport provider only once. You can change the name to `_failTransportProviderOnce`.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -510,6 +510,29 @@ public void testCoordinationWithBroadcastStrategy() throws Exception {
...
+    instance1.addConnector(testConnectorType, connector1, new BroadcastStrategy(Optional.empty()), false,
+        new SourceBasedDeduper(), null);
+    instance1.start();
+
+    ZkClient zkClient = new ZkClient(_zkConnectionString);
+    DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, testConnectorType, datastreamName1);
+    //verify the assignment
+    assertConnectorAssignment(connector1, WAIT_TIMEOUT_MS, datastreamName1);
+    String instance1Path = KeyBuilder.instanceAssignments(testCluster, instance1.getInstanceName());
+    Assert.assertNotEquals(zkClient.getChildren(instance1Path).size(), 0);
```
vmaheshw, line 533: Cleanup the instantiated objects like zkclient, instance1. Some of these tests run in parallel and these open object handlers create issues.

vmaheshw (changes requested):

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
surajkn: Done

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: Done

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: I modified this to throw back the error up the call stack in case of InterruptedException so that it can fall back to the existing behavior in case of InterruptedException.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -644,6 +649,20 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
...
+        _metrics.updateKeyedMeter(CoordinatorMetrics.getKeyedMeter(
+            CoordinatorEvent.EventType.HANDLE_DATASTREAM_CHANGE_WITH_UPDATE), 1);
+        _eventQueue.put(CoordinatorEvent.createHandleDatastreamChangeEvent());
+      } else {
+        _metrics.updateKeyedMeter(CoordinatorMetrics.getKeyedMeter(
+            CoordinatorEvent.EventType.HANDLE_ASSIGNMENT_CHANGE), 1);
+        _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
+      }
+    }
+
```
surajkn, line 665: I refactored the code a bit from my previous patch.
As we had discussed to update the metric count only once irrespective of the number of task failures I thought I could do that and also queue the corresponding event together.
To check if there were any task failures I am checking for a diff in task count instead of previous _failedDatastreamTasks set/map that I was using earlier. I thought that it would avoid having to maintain a consistent state for this additional variable, however this way I end up (potentially) modifying "currentAssignment" in the dispatch method. However I am not sure if this is against any coding guidelines that we follow. Pls let me know what is the correct way to handle this. Or maybe I can simply rename "dispatchAssignmentChangeIfNeeded" to "dispatchAndTrimAssignmentChangeIfNeeded" ?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: Looks like the circleci failure for this patch is because of the InterruptedException at line 778.
The circleci warning is as follows 
"""
Bug type BC_IMPOSSIBLE_INSTANCEOF (click for details)In class com.linkedin.datastream.server.CoordinatorIn method com.linkedin.datastream.server.Coordinator.initializeTask(DatastreamTask, Set, boolean)Actual type RuntimeExceptionExpected InterruptedExceptionValue loaded from eAt Coordinator.java:[line 778]
"""
In the code if I change the code to add a explicit catch block for InterruptedException Intellij tells me that its not possible to receive InterruptedException in this block of code. I read up on it and looks like InterruptedException is not thrown immediately and its upto the thread to interrupt itself whenever convenient next and that is typically whenever a blocking call (like Thread.sleep) is made. So basically blocking calls can throw InterruptedException and in this block there is no blocking call so we dont need to worry about InterruptedException here. Plz correct me if I misunderstood it, if not I will update the patch and remove InterruptedException handling from here.

vmaheshw: I checked the `CoordinatorEventProcessor` code and looks like you don't need to worry about the `InterruptedException` since it checks whether the thread is interrupted before picking the message from the queue. This means even if a new message is added, it will be skipped while processing if the exception is InterruptedException.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -737,19 +761,30 @@ private void uninitializeTasks(List<DatastreamTask> tasks) {
...
+      taskImpl.setCheckpoints(checkpoints);
+    } catch (Exception e) {
+      if (e instanceof InterruptedException) {
+        _log.warn("onAssignmentChange call got interrupted", e);
+        throw e;
+      }
+      _log.warn("Failed to initialize {} task", task.getDatastreamTaskName());
+      if (retryAndSaveError) {
+        _eventQueue.put(CoordinatorEvent.createHandleInstanceErrorEvent(ExceptionUtils.getRootCauseMessage(e)));
+        failedDatastreamTasks.add(task);
```
shrinandthakkar, line 785: could we try and mock this scenario in a test?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: Ack. Will remove the InterruptedException handling from here.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: I have added a unit test for this in TestCoordinator. What I am trying to do there in to fail TransportProvider creation once and which should result in the event to be queued again and the next time TransportProvider creation goes through and the task is successfully initialized.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -618,6 +621,8 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
         .filter(Objects::nonNull)
         .collect(Collectors.toList()));
 
+    int submittedTasks = currentAssignment.values().stream().mapToInt(List::size).sum();
```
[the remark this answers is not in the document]
surajkn: In "dispatchAssignmentChangeIfNeeded" I trim the "assignment" (DatastreamTask list) based on if any task initialization failed or not which will reflect here as well. Line 716 below.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -583,6 +584,8 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
       }
     });
 
+    int totalTasks = currentAssignment.values().stream().mapToInt(List::size).sum();
```
vmaheshw, line 587: nit: You can extract method out of this logic.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -644,6 +649,20 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
         .flatMap(Collection::stream)
         .collect(Collectors.toMap(DatastreamTask::getDatastreamTaskName, Function.identity())));
 
+    if ((totalTasks - submittedTasks) > 0) {
+      _log.warn("Failed to submit {} tasks from currentAssignment. Queueing onAssignmentChange event again", totalTasks - submittedTasks);
+      // Update the metric and queue the event only once for all the tasks that failed to be initialized
+      if (isDatastreamUpdate) {
```
vmaheshw, line 655: Qn: Why do you need a different code logic for this v/s TimeoutException, when the expectation is the same in both the cases.

``` if (isDatastreamUpdate) {
        _eventQueue.put(CoordinatorEvent.createHandleDatastreamChangeEvent());
      } else {
        _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
      }
      throw e;
``` There is no direct exception known in this path, but you can create DatastreamRuntimeException(). If you extract the highlighted code into a method and call it, it may work for you.

Logically it is the same code, but different way of achieving the same thing.

File: datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java [hunk 19]
```diff
@@ -583,6 +584,8 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
       }
     });
 
+    int totalTasks = currentAssignment.values().stream().mapToInt(List::size).sum();
+
```
vmaheshw, line 588: call `List<String> assignment = _adapter.getInstanceAssignment(_adapter.getInstanceName());` may become an expensive operation putting a direct load on the zookeeper server, in case of unlimited retries. I have a feeling, you will need the max retry logic as well, in case this issue happens in the production and makes the zookeeper server really busy with the `getChildren` call.

vmaheshw (commented):
Some code refactoring comments. One thought related to load on zookeeper server after this change.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
surajkn: Done

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -273,6 +280,7 @@ public void start() {
     // now that instance is started, make sure it doesn't miss any assignment created during
     // the slow startup
     _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
+    _assignmentRetryCount.set(0);
```
vmaheshw, line 283: Extracting this into a method with variable, whether to reset counter, will ensure any future changes to not miss the reset.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -180,6 +182,11 @@
 
   // Currently assigned datastream tasks by taskName
   private final Map<String, DatastreamTask> _assignedDatastreamTasks = new ConcurrentHashMap<>();
+  // Maintain a count of onAssignmentChange retries in case of task initialization/submission failure.
+  // If count reaches maximum threshold do not queue the event further. Reset the count if event raised
+  // from any other path.
+  private static final int ASSIGNMENT_MAX_RETRY_COUNT = 10;
```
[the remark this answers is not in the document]
vmaheshw: Also, retry count of -1 should mean unlimited retries.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -557,9 +566,44 @@ private void getAssignmentsFuture(List<Future<Boolean>> assignmentChangeFutures,
...
+    } else {
+      _log.warn("Updating metric for event " + CoordinatorEvent.EventType.HANDLE_ASSIGNMENT_CHANGE);
+      _metrics.updateKeyedMeter(CoordinatorMetrics.getKeyedMeter(
+          CoordinatorEvent.EventType.HANDLE_ASSIGNMENT_CHANGE), 1);
+      if (queueEvent) {
+        _log.warn("Queuing onAssignmentChange event");
+        _eventQueue.put(CoordinatorEvent.createHandleAssignmentChangeEvent());
+        _assignmentRetryCount.incrementAndGet();
+      }
+    }
```
vmaheshw, line 604: Based on `isDatastreamUpdate` you can create 2 variables eventType and Event and directly use them and avoid the duplicate code.

vmaheshw (dismissed):
Some cosmetic changes and make the retry field configurable.

shrinandthakkar (approved):

jzakaryan (approved):

PR #848, opened 2021-08-10
Migrating from AdminUtils with AdminClient

surajkn:
Replace kafka.admin.AdminUtils usage with kafka.client.AdminClient because AdminUtils is deprecated.
Also in a following effort 101tec ZkClient will be replaced by helix ZkClient and this is required for that.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
+++ b/datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
@@ -30,7 +30,6 @@ import org.testng.annotations.Test;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
-import kafka.admin.AdminUtils;
 
 import com.linkedin.data.template.StringMap;
 import com.linkedin.datastream.common.Datastream;
@@ -221,9 +220,7 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
     coordinator.initializeDatastream(datastream);
 
     // create topic
-    if (!AdminUtils.topicExists(_zkUtils, topic)) {
-      AdminUtils.createTopic(_zkUtils, topic, 2, 1, new Properties(), null);
-    }
+    createTopic(_adminClient, topic, 2);
 
     // Make sure "*" is converted to a list of partitions
     pausedPartitions.put(topic, new HashSet<>(Collections.singletonList("*")));
@@ -274,9 +271,9 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
     String saltyTopic = "SaltyPizza";
     String saladTopic = "HealthySalad";
 
-    createTopic(_zkUtils, saladTopic);
-    createTopic(_zkUtils, yummyTopic);
-    createTopic(_zkUtils, saltyTopic);
+    createTopic(_adminClient, saladTopic);
+    createTopic(_adminClient, yummyTopic);
+    createTopic(_adminClient, saltyTopic);
 
     // create a datastream to consume from topics ending in "Pizza"
     Datastream datastream = KafkaMirrorMakerConnectorTestUtils.createDatastream("pizzaStream", _broker, "\\w+Pizza");
@@ -378,7 +375,7 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
     String yummyTopic = "YummyPizza";
     String saltyTopic = "SaltyPizza";
 
-    createTopic(_zkUtils, yummyTopic);
+    createTopic(_adminClient, yummyTopic);
 
     // create a datastream to consume from topics ending in "Pizza"
     Datastream datastream = KafkaMirrorMakerConnectorTestUtils.createDatastream("pizzaStream", _broker, "\\w+Pizza");
@@ -409,7 +406,7 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
         "Expected record from YummyPizza topic but got record from " + record.getDestination().get());
 
     // create a new topic ending in "Pizza" and produce an event to it
-    createTopic(_zkUtils, saltyTopic);
+    createTopic(_adminClient, saltyTopic);
     KafkaMirrorMakerConnectorTestUtils.produceEvents(saltyTopic, 1, _kafkaCluster);
 
     if (!PollUtils.poll(() -> datastreamProducer.getEvents().size() == 2, POLL_PERIOD_MS, POLL_TIMEOUT_MS)) {
@@ -429,8 +426,8 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
     String yummyTopic = "YummyPizza";
     String saltyTopic = "SaltyPizza";
 
-    createTopic(_zkUtils, yummyTopic);
-    createTopic(_zkUtils, saltyTopic);
+    createTopic(_adminClient, yummyTopic);
+    createTopic(_adminClient, saltyTopic);
 
     // create a datastream to consume from topics ending in "Pizza"
     Datastream datastream = KafkaMirrorMakerConnectorTestUtils.createDatastream("pizzaStream", _broker, "\\w+Pizza");
@@ -460,7 +457,7 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
     datastreamProducer.getEvents().clear();
 
     // delete the topic
-    deleteTopic(_zkUtils, yummyTopic);
+    deleteTopic(_adminClient, yummyTopic);
 
     // produce another event to SaltyPizza
     KafkaMirrorMakerConnectorTestUtils.produceEvents(saltyTopic, 1, _kafkaCluster);
@@ -611,7 +608,7 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
 
     String yummyTopic = "YummyPizza";
 
-    createTopic(_zkUtils, yummyTopic, 1);
+    createTopic(_adminClient, yummyTopic, 1);
 
     // create a datastream to consume from topics ending in "Pizza"
     Datastream datastream = KafkaMirrorMakerConnectorTestUtils.createDatastream("pizzaStream", _broker, "\\w+Pizza");
@@ -637,7 +634,7 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
         ImmutableSet.of(yummyTopic + "-0"));
 
     String saltyTopic = "SaltyPizza";
-    createTopic(_zkUtils, saltyTopic, 2);
+    createTopic(_adminClient, saltyTopic, 2);
 
     Assert.assertTrue(PollUtils.poll(() -> partitionChangeCalls.get() == 2, POLL_PERIOD_MS, POLL_TIMEOUT_MS));
     partitionInfo = connector.getDatastreamPartitions();
@@ -716,7 +713,7 @@ public class TestKafkaMirrorMakerConnector extends BaseKafkaZkTest {
   @Test
   public void testGetStateForMultipleTasks() throws Exception {
     List<String> topics = Arrays.asList("YummyPizza", "SaltyPizza", "HealthySalad");
-    topics.forEach(t -> createTopic(_zkUtils, t));
+    topics.forEach(t -> createTopic(_adminClient, t));
 
     // create a datastream to consume from topics ending in "Pizza"
     Datastream datastream = KafkaMirrorMakerConnectorTestUtils.createDatastream("pizzaStream", _broker, "\\w+Pizza");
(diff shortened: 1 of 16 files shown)


gradle/dependency-versions.gradle
```diff
@@ -21,4 +21,5 @@ ext {
     testngVersion = "7.1.0"
     zkclientVersion = "0.11"
```
surajkn, line 22: I should probably delete this as well

jzakaryan (commented):
@surajkn did a walkthrough of the changes and explained the differences in APIs. Waiting for him to confirm config changes for Kafka client before approving the PR.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTestUtils.java
```diff
@@ -61,20 +62,21 @@ private KafkaTestUtils() {
...
    * Note: The topic creation must be issued before this method is called.
    *
+   * @param adminClient Kafka Admin client to check if topic exists
    * @param topic the topic to wait for broker assignment
    * @param brokerList the brokers in the Kafka cluster
    * @throws IllegalStateException if the topic is not ready before the timeout ({@value #DEFAULT_TIMEOUT_MS} ms)
    */
-  public static void waitForTopicCreation(ZkUtils zkUtils, String topic, String brokerList) throws IllegalStateException {
-    Validate.notNull(zkUtils);
+  public static void waitForTopicCreation(AdminClient adminClient, String topic, String brokerList) throws IllegalStateException {
```
surajkn, line 70: I am not sure if we need this anymore. In this patch I replace calls to AdminUtils.createTopic with calls to AdminClient.createTopics method, and createTopics essentially returns a "KafkaFuture" object on which I currently do a blocking wait (to wait for topic to be created). So we probably don't need this anymore? I have currently removed all calls to this method in test cases and its working fine.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTestUtils.java
```diff
@@ -102,6 +104,24 @@ public static void waitForTopicCreation(ZkUtils zkUtils, String topic, String br
     throw new IllegalStateException("Topic was not ready within the timeout");
   }
 
+  /**
+   * Check if the given topic exists yet or not.
+   *
+   * @param adminClient AdminClient instance to check if topics exists
+   * @param topic Topic name
+   */
+  public static boolean topicExists(AdminClient adminClient, String topic) {
```
surajkn, line 113: I tried to import "topicExists" method from BaseKafkaZkTest class here but I kept getting "package com.linkedin.datastream.testutil does not exist" error so I copied the method here. Its probably some classpath issue but I am not sure how to resolve it in this repo.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
```diff
@@ -245,47 +243,90 @@ public void dropDestination(Datastream datastream) {
...
 
-        AdminUtils.createTopic(_zkUtils.get(), topicName, numberOfPartitions, replicationFactor, topicConfig, RackAwareMode.Disabled$.MODULE$);
-      } else {
+    try {
+      AdminClient adminClient = getAdminClient(datastream);
+      NewTopic newTopic = getTopic(topicName, numberOfPartitions, topicProperties);
+      // Removing from topic config since its passed as a direct argument
+      topicProperties.remove("replicationFactor");
+      // TopicExistsException is thrown if topic exists. Its a no-op.
+      adminClient.createTopics(Collections.singletonList(newTopic)).all().get();
```
surajkn, line 287: By doing ".all().get()" we have a blocking wait for the topic to be created. I am not sure what exactly was the behavior with AdminUtils and/or if doing a blocking wait is the correct thing to do here. I am guessing this is the correct way?

However, having said that. On repeated execution of all the unit tests, I noticed that the test "testCreateDatastreamHappyPathDefaultRetention" is a little flaky in that it randomly fails every now and then (very rarely though, hard to reproduce). This test calls "getRetention" which queries topic config and returns retention time if its exists in policy else null, and when this test fails its because getRetention returned null instead of actual expected value. I am not sure if this test was flaky even before. I am not sure how to explain this behavior.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
```diff
@@ -245,47 +243,90 @@ public void dropDestination(Datastream datastream) {
...
+        LOG.error(String.format("Creating topic %s failed with exception: ", topicName), e);
+        throw new DatastreamRuntimeException(e);
       }
     } catch (Throwable e) {
       LOG.error("Creating topic {} failed with exception {}", topicName, e);
       throw e;
     }
   }
 
+  private AdminClient getAdminClient(Datastream datastream) {
```
surajkn, line 301: Do we need to cache AdminClient instances based on the destinationBrokers ? Depends on how frequently this method is called and how expensive it is to re-instantiate AdminClient object, but I am not sure of it.

datastream-kafka/src/test/java/com/linkedin/datastream/kafka/TestKafkaTransportProvider.java
```diff
@@ -220,9 +220,9 @@ public void testSendMultipleEventsInSingleDatastreamProducerRecord() throws Exce
 
     DatastreamTask task = new DatastreamTaskImpl(Collections.singletonList(ds));
     TransportProvider transportProvider = provider.assignTransportProvider(task);
-    provider.createTopic(destinationUri, 1, new Properties());
+    provider.createTopic(destinationUri, 1, new Properties(), ds);
 
-    KafkaTestUtils.waitForTopicCreation(_zkUtils, topicName, _kafkaCluster.getBrokers());
+    //KafkaTestUtils.waitForTopicCreation(_adminClient, topicName, _kafkaCluster.getBrokers());
```
surajkn, line 225: Will remove this in following update to this patch. Same for another instance below.

gradle/dependency-versions.gradle
jzakaryan: If there are no usages you should

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTestUtils.java
surajkn: Yes, correct.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
```diff
@@ -245,47 +243,90 @@ public void dropDestination(Datastream datastream) {
...
-      } else {
+    try {
+      AdminClient adminClient = getAdminClient(datastream);
+      NewTopic newTopic = getTopic(topicName, numberOfPartitions, topicProperties);
+      // Removing from topic config since its passed as a direct argument
+      topicProperties.remove("replicationFactor");
+      // TopicExistsException is thrown if topic exists. Its a no-op.
+      adminClient.createTopics(Collections.singletonList(newTopic)).all().get();
+    } catch (InterruptedException | ExecutionException e) {
+      if (e instanceof TopicExistsException || e.getCause() instanceof TopicExistsException) {
```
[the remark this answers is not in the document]
surajkn: Actually I found this reference implementation in LiKafkaTransporterProviderAdmin implementation [here](https://jarvis.corp.linkedin.com/codesearch/result/?name=LiKafkaTransportProviderAdmin.java&path=brooklin-li-common%2Fbrooklin-kafka%2Fsrc%2Fmain%2Fjava%2Fcom%2Flinkedin%2Fdatastream%2Fkafka&reponame=brooklin%2Fbrooklin-li-common#587) 
But let me double check this once anyway.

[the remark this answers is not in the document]
surajkn: So looks like TopicExistsException has an inheritance hierarchy (chain) that inherits from java.lang.Exception and both InterruptedException and ExecutionException also inherit from same java.lang.Excepption.

vmaheshw (dismissed):
Can you please split this PR and separate out the admin client change? AdminClient change can be merged as is.

For the zk client migration part, we can see if we have a jfrog branching. Otherwise, we can duplicate some of the code and have different classes called using config knob and then later get rid of the knob, if things are working fine. It may duplicate some code, but will definitely unblock you and remove your dependency from jfrog branching.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTestUtils.java
shrinandthakkar: If it's a circular dep issue, then could we reference this function from BaseKafkaZkTest class's call to avoid duplicate code?

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnector.java
```diff
@@ -221,9 +220,7 @@ public void testValidateDatastreamUpdatePausedPartitions() throws Exception {
     coordinator.initializeDatastream(datastream);
 
     // create topic
-    if (!AdminUtils.topicExists(_zkUtils, topic)) {
-      AdminUtils.createTopic(_zkUtils, topic, 2, 1, new Properties(), null);
-    }
+    createTopic(_adminClient, topic, 2);
```
shrinandthakkar, line 223: how could we configure the replication and other properties with the newer createTopic method?

datastream-testcommon/src/main/java/com/linkedin/datastream/testutil/BaseKafkaZkTest.java
```diff
@@ -63,25 +69,42 @@ public void afterMethodTeardown() {
...
-    if (!AdminUtils.topicExists(zkUtils, topic)) {
-      Properties properties = new Properties();
-      properties.put("message.timestamp.type", "LogAppendTime");
-      AdminUtils.createTopic(zkUtils, topic, partitionCount, 1, properties, null);
+  protected static void createTopic(AdminClient adminClient, String topic, int partitionCount) {
+    NewTopic newTopic = new NewTopic(topic, partitionCount, (short) 1);
+    try {
+      adminClient.createTopics(Collections.singletonList(newTopic)).all().get();
+    } catch (InterruptedException | ExecutionException e) {
+      if (e instanceof TopicExistsException || e.getCause() instanceof TopicExistsException) {
```
shrinandthakkar, line 77: do we know here for sure that InterruptedException or ExecutionException is just a one-layer wrapper around the TopicExistsException? Or would we need to recursively check the causes of the caught exception to know if it is TopicExistsException?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
```diff
@@ -245,47 +243,90 @@ public void dropDestination(Datastream datastream) {
...
+
+    return null;
   }
 
   /**
    * Create Kafka topic based on the destination connection string, if it does not already exist.
    * @param connectionString connection string from which to obtain topic name
    * @param numberOfPartitions number of partitions
-   * @param topicConfig topic config to use for topic creation
+   * @param topicProperties topic config to use for topic creation
```
shrinandthakkar, line 272: nit: add Datastream param in java doc

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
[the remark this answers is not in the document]
jzakaryan: I understand that, my point was that `TopicExistsException` is neither a descendant of `InterruptedException` nor is it  a descendant of `ExecutionException`. That first condition is always `false`.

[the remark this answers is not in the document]
shrinandthakkar: yea so, we may need recursive calls to get the cause of the exceptions until we hit the topic exists right?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
jzakaryan: Summarizing the offline discussion here:

There's no callbacks and returned futures with `AdminUtils.createTopic`. No hints that it's an asynchronous operation. I'm not a scala expert, but from the looks of it it's a synchronous call.

As for the failing/flaky test:

1. Was this test flaky even before or it became flaky as a result of the API changes? We need to rule out changes in behavior of the system after switching from `AdminUtils` to `AdminClient`
2. Does adding a thread sleep after topic creation (and before querying the retention of the new topic) make the test pass? 

For (1), we'll reach out to Kafka to confirm that there are no changes in behavior in createTopic APIs. For (2) we can try fixing the flaky test in a separate PR.

jzakaryan (dismissed):

shrinandthakkar (approved):

jzakaryan (approved):
LGTM

PR #850, opened 2021-08-19
Update the DatabaseChunkedReader to take the Connection as input rather than the DataSource

somandal:
Currently the `DatabaseChunkedReader` takes a `DataSource` as input and calls its `getConnection()` function to obtain a connection to the DB. In the release resources code path, it does not clean up this `Connection` object, and this is potentially a leak and should be closed. The way we use the `DatabaseChunkedReader` in Oracle bootstrap is that we reuse the same `DataSource` for seeding multiple tables. Thus we can potentially expect the Oracle bootstrap task to know when it no longer needs the `Connection` object and close it. This would require passing the `Connection` object to the `DatabaseChunkedReader` instead of the `DataSource`.

the contribution guidelines.


DEEPTHIKORAT (approved):
lgtm

atoomula (approved):
LGTM

PR #851, opened 2021-08-24
Fix running task data structure logic in AbstractKafkaConnector for failing to stop task

surajkn:
In onAssignmentChange we prematurely remove task from running task structure without checking if
stopTask succeeded or not.
This patch fixes this by only removing task reference from running task structure only if stopTask
is successful.

the contribution guidelines.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -624,11 +645,13 @@ public AbstractKafkaBasedConnectorTask getConnectorTask() {
   public Map<String, Map<Integer, Long>> getConsumptionLagForTask(DatastreamTask task) throws IllegalArgumentException {
     Validate.notNull(task);
 
-    ConnectorTaskEntry taskEntry = _runningTasks.getOrDefault(task, null);
-    if (taskEntry == null) {
-      throw new IllegalArgumentException("Task not found.");
-    }
+    synchronized (_runningTasks) {
```
jzakaryan, line 648: Clarified with @surajkn. This change is not related to the bug that's being fixed. This is just extra synchronization code.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -147,15 +150,32 @@ public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
...
+              e.printStackTrace();
+            }
+            return null;
+          }).filter(Objects::nonNull).collect(Collectors.toList());
+
+          if (toRemoveTasks.size() > 0) {
+            synchronized (_runningTasks) {
+              toRemoveTasks.forEach(_runningTasks::remove);
+            }
+          }
```
jzakaryan, line 178: Earlier we had an issue with tasks that weren't stopped properly but were removed from running tasks. Now such tasks won't be removed. Other than logging that an attempt to stop a task failed or timed out, do we have an idea on how to deal with such "zombie" tasks? If they'll be living in-memory indefinitely, can they be a source for a memory leak?

surajkn: In the "restartDeadTasks" method we check again is a given task is "dead" and if yes it will attempt to stop it again. So if its a transient error eventually it should be stopped and removed. At least this is my understanding, pls correct me if I am missing something.

vmaheshw (dismissed):
AFAIK, this change will not work because the task to be removed is still part of runningTasks and if the restartDeadTasks logic will attempt to restart this task thread.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -147,32 +156,48 @@ public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
...
-          kafkaBasedConnectorTask.checkForUpdateTask(task);
-          // Make sure to replace the DatastreamTask with most up to date info
-          // This is necessary because DatastreamTaskImpl.hashCode() does not take into account all the
-          // fields/properties of the DatastreamTask (e.g. dependencies).
-          _runningTasks.remove(task);
-          _runningTasks.put(task, connectorTaskEntry);
-          continue; // already running
+      boolean toCallRestartDeadTasks = false;
+      // This is the only place where we synchronize on both _toStopTasks and _runningTasks. But if such a case
+      // arises elsewhere the order of synchronization should be _runningTasks and then _toStopTasks.
```
somandal, line 173: In the previous code block you are also accessing the lock for `_toStopTasks`, so this comment should be moved  up and modified to talk about accessing both in this function.
Is it really necessary to leave a  comment here since you've already left a comment where these are defined? In the future if someone adds another block where both are accessed, they may miss updating/removing this comment.

also nit to reword in case you decide to keep it:
"We only synchronize both _runningTasks and _toStopTasks in this function. If these need to be synchronized together in other parts of the code, please follow the correct lock ordering semantics."

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
somandal: For non-transient errors, we will never be able to start the task to replace this dead task which isn't able to stop. This can lead to stopped partitions problems and will be very hard to debug and RC.

Please ensure that you add a metric to track when we are unable to stop a task so what we can quickly check for this metric when we run into issues. Can you perhaps also add a comment explaining that this can lead to stopped partitions issue?

Also, please ensure that we fix pieces of code that don't honor the interrupt signal in the future so that we can minimize this scenario. @surajkn and I don't have the full context on what kind of exceptions are seen such that we are unable to stop tasks. We are also not aware of how often his happens, so any insights from @vmaheshw or @jzakaryan will be helpful here.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -147,32 +156,48 @@ public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
...
+              connectorTaskEntry = createKafkaConnectorTask(task);
+            }
+            _runningTasks.put(task, connectorTaskEntry);
+          }
         }
-
-        _runningTasks.put(task, createKafkaConnectorTask(task));
+      }
+      // If any tasks pending stop were re-assigned to this host we explicitly call restartDeadTasks to ensure
+      // stop and (re)start of the task. We do this to not wait for next periodic invocation of restartDeadTasks.
```
somandal, line 198: nit: reword: "We do this to avoid waiting for the next periodic invocation of restartDeadTasks."

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -231,8 +260,14 @@ protected void restartDeadTasks() {
         if (isTaskDead(connectorTaskEntry)) {
           _logger.warn("Detected that the kafka connector task is not running for datastream task {}. Restarting it",
               datastreamTask.getDatastreamTaskName());
-          boolean stopped = stopTask(datastreamTask, connectorTaskEntry);
-          if (stopped) {
+          // Connector task attempt stop will be set to false if stopTask is successful in stopping the connector task.
+          // If stopping the task fails the flag will remain True to attempt another stop before creating starting
+          // another connector task.
```
somandal, line 265: I'm not sure why the comments talk about setting things false or true here. It looks like stopTask() returns the DatastreamTask, so perhaps modify the comments to talk about null/not-null instead? Or maybe I didn't understand the comments too well, so can you explain what you're trying to say here?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -250,14 +285,75 @@ protected void restartDeadTasks() {
     }
   }
 
+  /**
+   * Returns the number of tasks yet to be stopped.
+   */
+  int getToStopTaskCount() {
```
somandal, line 291: nit: rename getTasksToStopCount() if you changed the variable name from _toStopTasks to _tasksToStop

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -250,14 +285,75 @@ protected void restartDeadTasks() {
...
+    synchronized (_toStopTasks) {
+      return _toStopTasks.size();
+    }
+  }
+
+  /**
+   * Returns the number of currently running tasks.
+   */
+  int getRunningTaskCount() {
```
somandal, line 300: nit: rename: getRunningTasksCount()

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -77,6 +81,8 @@
 
   private static final Duration CANCEL_TASK_TIMEOUT = Duration.ofSeconds(75);
   private static final Duration POST_CANCEL_TASK_TIMEOUT = Duration.ofSeconds(15);
+  // This should be at least equal to or greater than sum of cancel and post cancel timeouts
+  private static final Duration CANCEL_TASK_THREAD_TIMEOUT = Duration.ofSeconds(90);
```
somandal, line 85: Why not make variables for `75` and `15`, and add them here? That'll be the best way to enforce what you're intending with your comment.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -250,14 +285,75 @@ protected void restartDeadTasks() {
...
+            return stopTaskFuture.get(CANCEL_TASK_THREAD_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
+          } catch (ExecutionException | InterruptedException | TimeoutException e) {
+            _logger.warn("Stop task future failed with exception", e);
+          }
+          return null;
+        }).filter(Objects::nonNull).collect(Collectors.toList());
+
+        if (toRemoveTasks.size() > 0) {
+          synchronized (_toStopTasks) {
+            // Its possible that while stopping the task was pending there was another OnAssignmentChange event
```
somandal, line 337: nit: OnAssignmentChange -> onAssignmentChange

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -271,15 +367,16 @@ private boolean stopTask(DatastreamTask datastreamTask, ConnectorTaskEntry conne
         if (!connectorTask.awaitStop(POST_CANCEL_TASK_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)) {
           _logger.warn("Connector task for datastream task {} did not stop even after {} ms.", datastreamTask,
```
somandal, line 368: Add a metric to track how often we are unable to stop tasks within the two cancel timeouts.

somandal (commented):
@surajkn Please also add this as a potential STOPPED partition scenario to the BMM debugging runbook and point to the relevant metrics once they are added. Also ensure this is well tested as we discussed since this is a very critical piece of code.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestAbstractKafkaConnector.java
```diff
@@ -75,12 +77,72 @@ public void testConnectorRestartCalled() {
     connector.stop();
   }
 
+  @Test
+  public void testOnAssignmentChangeReassignment() {
```
somandal, line 81: Will it be possible to add a test for the new metric you added to ensure it gets incremented when a task stop fails?

somandal (approved):
Left a minor comment about adding a potential test case for the new metric. Please do ensure that you have captured the expected stopped partitions issue in the BMM debugging guide (and let the SREs know as well) so that people know that they should look for that as a potential RC.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/TestAbstractKafkaConnector.java
surajkn: I added a check for that in the test "testOnAssignmentChangeStopTaskFailure" below at line 133, I check for the metric count to be non-zero. Is that what you were suggesting?

jzakaryan (approved):

PR #853, opened 2021-09-24
Stop the tasks in parallel in AbstractKafkaConnector

vmaheshw:
Currently during container shutdown, the tasks are shutdown sequentially in `AbstractKafkaConnector`. The shutdown code waits for the task threads to be killed for 60+ sec. This increases the overall stop time, if the kafka producers have a lot pending to flush in scale environment. Ideally all the tasks can be stopped in parallel, similar to when the task assignment changes.

Using the shutdown executor service to stop the tasks. When the executor is stopped, it will force kill the ongoing operations. So, I'm changing the shutdown executor time to be same as the time it waits to stop a task which is currently 75 + 15 = 90 sec.


surajkn (approved):

jzakaryan (approved):
LGTM

PR #855, opened 2021-10-19
Add Stats to DatastreamTaskImpl

vmaheshw:
We frequently hear need to get some of the task level metrics for diagnostics that can be retrieved using the brooklin-service end-point.

 LoadBasedPartitionAssignmentStrategy distributes the partitions evenly based on the load. To be able to debug and validate the distribution, it is important to be able to pull out the metrics at task level and perform offline analytics on the data.

This PR exposes a new knob stats that can used to save the task level stats on the zookeeper and can be used to retrieve similar to other end-points.


vmaheshw:
> Just curious, why do we want to save these stats in ZK instead of simply reporting this as a task level metric ? Is it because in Ingraphs its hard/not possible to identify a specific task's metrics?

There is a limitation of number of metrics that we can emit from the container. Also, if we want to build a diagnostics command to collect the information from large clusters and analyze the data, it is difficult with the metrics. Also, these metrics are emitted only by the leader and on leader switch, these metrics will not emitted until the datastream is restarted.

PR #861, opened 2021-10-28
Fix restartDeadTask logic when the task thread has died

surajkn:
In restartDeadTask we check if a task is dead and attempt to cleanly stop the task before restarting
the task again. To check if the task is dead or not, one of the things we check is if the task
thread is alive or not. Even though this is a correct check, we don't need to attempt to stop the
task in this case as the task thread is already dead.
Another fix is in KafkaMirrorMakerConnectorTask to count down the latch in case we fail to acquire
the task lock. The bug is, that if the task thread fails to acquire the lock and exits with the exception
the latch count is not reduced to 0 and in the next period restartDeadTask check the connector
thinks that the task has not yet stopped and continues to retry stopTask.

A scenario in which the above can happen is, when a task gets reassigned to a different host and
the task thread in the new host fails to acquire the lock because the previous has still not fully
stopped and released the task lock. This may happen when there are a huge number of tasks
being reassigned.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -436,13 +436,9 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
         }
       }
       postShutdownHook();
-      releaseTaskLatch();
+      _stoppedLatch.countDown();
       _logger.info("{} stopped", _taskName);
     }
-  }
-
-  protected void releaseTaskLatch() {
-    _stoppedLatch.countDown();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -436,11 +436,15 @@ public void run() {
         }
       }
       postShutdownHook();
-      _stoppedLatch.countDown();
+      releaseTaskLatch();
```
vmaheshw, line 439: releaseTaskLatch does not sound right. At Abstract level, there is no concept of releaseTask. How about markTaskAsDown or markTaskAsStopped ? Also, since you got rid of the other half, you can skip creation of function from this and do it together as second half.

vmaheshw (approved):

jzakaryan (approved):

PR #863, opened 2021-11-03
Add retry to query retention time for destination topic

surajkn:
Since migrating to AdminClient to create destination topic it was observed that sometimes there is a
lag in topic metadata to be synced to all brokers. This may result in getting incorrect (null)
retention time. Adding a retry logic to account for delay in topic metadata sync.

the contribution guidelines.

### DIFF
--- a/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
+++ b/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
@@ -39,6 +39,7 @@ import com.linkedin.datastream.common.DatastreamMetadataConstants;
 import com.linkedin.datastream.common.DatastreamRuntimeException;
 import com.linkedin.datastream.common.DatastreamSource;
 import com.linkedin.datastream.common.DatastreamUtils;
+import com.linkedin.datastream.common.PollUtils;
 import com.linkedin.datastream.common.VerifiableProperties;
 import com.linkedin.datastream.metrics.BrooklinMetricInfo;
 import com.linkedin.datastream.server.DatastreamTask;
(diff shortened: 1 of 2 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProviderAdmin.java
```diff
@@ -250,23 +253,24 @@ public Duration getRetention(Datastream datastream) {
...
-      // to all the brokers yet and adminClient might query such a broker in which case topic retention will
-      // not be present in the topic config. To circumvent this we need to query only the controller node in
-      // Kafka cluster which will always return successful result, but that is more involved.
-      ConfigResource topicResource = new ConfigResource(ConfigResource.Type.TOPIC, topicName);
-      Map<ConfigResource, Config> topicConfigMap = adminClient.describeConfigs(Collections.singletonList(topicResource)).all().get();
-      Config topicConfig = topicConfigMap.get(topicResource);
-      ConfigEntry entry = topicConfig.get(TOPIC_RETENTION_MS);
-      if (entry != null) {
-        return Duration.ofMillis(Long.parseLong(entry.value()));
+    // In rare cases it may happen that even though topic has been created, its metadata hasn't synced
```
vmaheshw, line 256: What will happen if it is not synced in 10 sec?

surajkn: I am not sure the implication of simply returning a null here if we don't find retention time in topic config. The other alternative I can think of is to throw a runtime exception. In the existing code it looks like retention time being null is a valid case(?), or at least returning a null value for retention time from here should be handled in the rest of the code which depends on this retention time. 
So I guess my question is: Is there a valid scenario in which we return null from here. If yes, then how do we distinguish between a genuine null case vs where its a delay in topic config being synced? If no, then was the existing code returning null in case of not being able to get retention time wrong? 

Will add a couple of log messages here to indicate if polling multiple times to get the retention time in a following update to the PR.

surajkn: Discussed offline with Vaibhav. This change does not impact the already existing behavior, so we should be ok to proceed.

vmaheshw (approved):

jzakaryan (approved):
Looks good to me

PR #864, opened 2021-11-05
Make Default byte in rate and Msg in rate configurable

vmaheshw:
The default values set may need to be changed based on the specific scenario. Specifically, this brings a dependency that if the default throughput rate per task in 5MB and the default is 5KB, it implies that a task can only process 1000 partitions. Making this a configurable property will make it customizable based on the requirement.

Note: This change does not cover config at a datastream level. We can add it based on the requirement.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyConfig.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyConfig.java
@@ -15,15 +15,14 @@ import com.linkedin.datastream.common.VerifiableProperties;
  */
 public class LoadBasedPartitionAssignmentStrategyConfig extends PartitionAssignmentStrategyConfig {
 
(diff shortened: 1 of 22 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategyConfig.java
```diff
@@ -39,6 +40,9 @@
   private final int _throughputInfoFetchRetryPeriodMs;
   private final boolean _enableThroughputBasedPartitionAssignment;
   private final boolean _enablePartitionNumBasedTaskCountEstimation;
+  private final int _defaultPartitionBytesInKBRAte;
```
shrinandthakkar, line 43: nit: casing error in the name, _defaultPartitionBytesInKB**RAte**

shrinandthakkar (approved):

shrinandthakkar (approved):

jzakaryan (approved):
lgtm

PR #868, opened 2021-11-11
Refactor Stopping Tasks On Assignment Change of Tasks

shrinandthakkar:
Only trigger stopping for the tasks that need to be canceled per assignment change in the `onAssignmentChange` function. This change prevents calling stops on tasks recurrently in case of frequent assignment change triggers. 
________
the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
@@ -102,7 +102,10 @@ public abstract class AbstractKafkaConnector implements Connector, DiagnosticsAw
   // multiple concurrent threads. If access is required to both maps then the order of synchronization must be
   // _runningTasks followed by _tasksToStop to prevent deadlocks.
   private final Map<DatastreamTask, ConnectorTaskEntry> _runningTasks = new HashMap<>();
-  private final Map<DatastreamTask, ConnectorTaskEntry> _tasksToStop = new HashMap<>();
+
+  // _tasksPendingStop contains the tasks that are pending stop across various assignment changes. The periodic health
+  // check call will attempt to stop these tasks until they are not stopped / are stuck somewhere in stop path.
+  private final Map<DatastreamTask, ConnectorTaskEntry> _tasksPendingStop = new HashMap<>();
 
   // A daemon executor to constantly check whether all tasks are running and restart them if not.
   private final ScheduledExecutorService _daemonThreadExecutorService =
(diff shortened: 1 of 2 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -102,6 +102,8 @@
   // multiple concurrent threads. If access is required to both maps then the order of synchronization must be
   // _runningTasks followed by _tasksToStop to prevent deadlocks.
   private final Map<DatastreamTask, ConnectorTaskEntry> _runningTasks = new HashMap<>();
+
+  // _tasksToStop contains the tasks which are needed to be stopped across various assignment changes.
```
vmaheshw, line 106: There is more clarity on what this map is storing and how it is important.

It contains the list of tasks that are pending stop. The periodic scheduled health check call will try to stop these tasks again, with the assumption that that the task is stuck somewhere in stop path or the stop is not completed yet.

I feel it will be very useful to add this information in the comments and ensure the variable name is more meaningful for better code readability for other users.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -308,44 +311,50 @@ int getRunningTasksCount() {
...
-      // to stop, or are stuck indefinitely. A separate thread is spawned to handle this because the Coordinator
-      // requires that this step completely because we call this from onAssignmentChange() (assignment thread gets
-      // killed if it takes too long) and restartDeadTasks which must complete quickly.
-      List<Future<DatastreamTask>> stopTaskFutures = _tasksToStop.keySet().stream()
-          .map(task -> asyncStopTask(task, _tasksToStop.get(task)))
-          .collect(Collectors.toList());
+  /**
+   * Attempt to stop the unassigned tasks from the argument map.
+   */
+  private void scheduleStopping(Map<DatastreamTask, ConnectorTaskEntry> tasks) {
```
vmaheshw, line 323: the name of the method looks incomplete.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -154,6 +156,7 @@ public synchronized void onAssignmentChange(List<DatastreamTask> tasks) {
     _logger.info("onAssignmentChange called with tasks {}", tasks);
 
     synchronized (_runningTasks) {
```
vmaheshw, line 158: Do you have plans to fix the metric issue in the same PR or separate PR?

vmaheshw (dismissed):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
shrinandthakkar: Separate PR, https://github.com/linkedin/brooklin/pull/869

vmaheshw: PR #869 description does not tell how it is causing the metric issue. Also, in the situation where we saw metric issue was not related to InterruptedException path. Look at this code for more ideas.

`private DatastreamTask stopTask(DatastreamTask datastreamTask, ConnectorTaskEntry connectorTaskEntry) {`

vmaheshw (approved):
Thank you for fixing this.

surajkn (commented):
So the main objective of the change is to not call stop one more time while a current stop attempt is in progress. Correct? 
Is yes I think that is still possible since in "restartDeadTask" we still call "stopUnassignedTasks" which calls "scheduleTasksToStop" with "_tasksPendingStop"

shrinandthakkar:
> So the main objective of the change is to not call stop one more time while a current stop attempt is in progress. Correct? Is yes I think that is still possible since in "restartDeadTask" we still call "stopUnassignedTasks" which calls "scheduleTasksToStop" with "_tasksPendingStop"

Actually, the main objective of this change is to separate the two triggers of stopping the tasks. We want to avoid stopping the long-pending tasks whenever onAssignmentChange is invoked. We only want to trigger stop for the running tasks which are not part of the new assignment when the onAssignmentChange is called. 

Those long-pending tasks could still be re-triggered for stopping periodically from the restartDeadTask function, which should be the expected behavior I think.

vmaheshw (dismissed):

surajkn (approved):

surajkn (approved):

vmaheshw (approved):

PR #869, opened 2021-11-16
Catch interrupt signal in the connector task runnable

shrinandthakkar:
In the stopTasks logic, the thread might not get interrupted sometimes and thus could land up running forever. This change tries to catch the interrupt status in the runnable to eventually stop the task instead of recurring forever. 

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -388,7 +388,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
       consumerSubscribe();
 
       ConsumerRecords<?, ?> records;
-      while (!_shutdown) {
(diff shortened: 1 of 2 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -388,7 +388,7 @@ public void run() {
       consumerSubscribe();
 
       ConsumerRecords<?, ?> records;
-      while (!_shutdown) {
+      while (!_shutdown && !Thread.currentThread().isInterrupted()) {
```
vmaheshw, line 391: The code path trying to interrupt the thread, does it not call shutdown() method which sets _shutdown to true?

surajkn: Can you pls explain the code flow for how you think the interrupted exception is being lost (or more likely being caught and then ignored)?

vmaheshw:
@shrinandthakkar Can we close this PR if it is not relevant anymore?

PR #870, opened 2021-11-17
Replace 101tec ZkClient with Helix ZkClient

surajkn:
Use helix version 1.0.2. Extend common/zk/ZkClient.java from helix zookeeper client and refactoring
some code.

the contribution guidelines.

### DIFF
--- a/datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
+++ b/datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
@@ -6,19 +6,13 @@
 package com.linkedin.datastream.common.zk;
 
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
 import java.util.Stack;
 
-import org.I0Itec.zkclient.ZkConnection;
-import org.I0Itec.zkclient.exception.ZkInterruptedException;
-import org.I0Itec.zkclient.exception.ZkMarshallingError;
-import org.I0Itec.zkclient.exception.ZkNoNodeException;
-import org.I0Itec.zkclient.exception.ZkNodeExistsException;
-import org.I0Itec.zkclient.serialize.ZkSerializer;
-import org.apache.zookeeper.CreateMode;
-import org.apache.zookeeper.data.Stat;
+import org.apache.helix.zookeeper.zkclient.exception.ZkMarshallingError;
+import org.apache.helix.zookeeper.zkclient.exception.ZkNodeExistsException;
+import org.apache.helix.zookeeper.zkclient.serialize.ZkSerializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -26,7 +20,7 @@ import com.google.common.annotations.VisibleForTesting;
 
 
 /**
- * ZKClient is a wrapper of {@link org.I0Itec.zkclient.ZkClient}. It provides the following
+ * ZKClient is a wrapper of {@link org.apache.helix.zookeeper.impl.client.ZkClient}. It provides the following
  * basic features:
  * <ol>
  *  <li>tolerate network reconnects so the caller doesn't have to handle the retries</li>
@@ -34,7 +28,7 @@ import com.google.common.annotations.VisibleForTesting;
  *  <li>additional features like ensurePath to recursively create paths</li>
  * </ol>
  */
-public class ZkClient extends org.I0Itec.zkclient.ZkClient {
+public class ZkClient extends org.apache.helix.zookeeper.impl.client.ZkClient {
   public static final String ZK_PATH_SEPARATOR = "/";
   public static final int DEFAULT_CONNECTION_TIMEOUT = 60 * 1000;
   public static final int DEFAULT_SESSION_TIMEOUT = 30 * 1000;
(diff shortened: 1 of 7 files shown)


datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
```diff
@@ -86,71 +80,20 @@ public ZkClient(String zkServers, int sessionTimeoutMs, int connectionTimeoutMs,
...
-    try {
-      return retryUntilConnected(() -> _connection.getChildren(path, watch));
-    } finally {
-      long endT = System.nanoTime();
-      if (LOG.isTraceEnabled()) {
-        LOG.trace("getChildren, path: {}, time: {} ns", path, (endT - startT));
-      }
-    }
+    return super.getChildren(path, watch);
   }
```
jzakaryan, line 98: Is the only reason for keeping this class around `ensureReadData`? I was just wondering if there's a substitute for that method in the base apache helix zookeeper client, because it seems like a generic/common operation.

jzakaryan (approved):
lgtm

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
```diff
@@ -86,71 +80,20 @@ public ZkClient(String zkServers, int sessionTimeoutMs, int connectionTimeoutMs,
     _zkSessionTimeoutMs = sessionTimeoutMs;
   }
 
-  @Override
```
vmaheshw, line 89: What is the substitute for all the code that you are deleting?

gradle/dependency-versions.gradle
```diff
@@ -21,5 +21,5 @@ ext {
     testngVersion = "7.1.0"
     zkclientVersion = "0.11"
```
vmaheshw, line 22: Is `zkClientVersion` required?

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
surajkn: All those methods are implemented in https://github.com/apache/helix/blob/master/zookeeper-api/src/main/java/org/apache/helix/zookeeper/zkclient/ZkClient.java .

We seemed to be doing basically the same thing as this implementation of ZkClient so I thought to simply rely on the base implementation for these.

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
surajkn: I couldn't find a equivalent implementation for that in Helix. So I kept those around.

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
```diff
@@ -347,9 +216,12 @@ public void ensurePath(String path) {
     return (T) _zkSerializer.deserialize(data);
   }
 
+  /**
+   * Get Zk sessions session ID
```
vmaheshw, line 220: fix the javadoc.

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
```diff
@@ -347,9 +216,12 @@ public void ensurePath(String path) {
...
+   */
   @VisibleForTesting
   public long getSessionId() {
```
vmaheshw, line 223: Is this method required anymore? Can we directly use the super method?

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
```diff
@@ -86,71 +80,20 @@ public ZkClient(String zkServers, int sessionTimeoutMs, int connectionTimeoutMs,
...
-      }
-    }
+    return super.exists(path, watch);
   }
 
+  /**
+   * Get all children of zk path. Changes the access modified to public, its defined as protected in parent class.
+   */
   @Override
   public List<String> getChildren(final String path, final boolean watch) {
```
vmaheshw, line 95: You can probably get rid of this method itself since you are only calling super.()

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
```diff
@@ -212,80 +155,6 @@ public String ensureReadData(final String path) {
...
-      if (LOG.isTraceEnabled()) {
-        LOG.trace("writeData, path: {}, time: {} ns", path, (endT - startT));
-      }
-    }
-  }
-
-  @Override
-  public String create(final String path, Object data, final CreateMode mode) throws RuntimeException {
-    if (path == null) {
-      throw new IllegalArgumentException("path must not be null.");
```
vmaheshw, line 251: The new library throws `NullPointerException` instead of `IllegalArgumentException`. Can you plz check the contract of new methods ? Mostly the unit-tests are not covering it, otherwise the tests should have failed. May be a good idea to add those tests

vmaheshw: Also, this is something to bring up with Helix team that their contact says `IllegalArgumentException` but internally the method throws `NullPointerException`.

vmaheshw (commented):
Can you please address the comments from last review as well?

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
surajkn: Removed this method

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
surajkn: Removed the method. Not required anymore

datastream-utils/src/main/java/com/linkedin/datastream/common/zk/ZkClient.java
surajkn: Added unit test for NullPointerException. Will bring up the discrepancy with Helix team separately

PR #871, opened 2021-11-24
Use topic level throughput information when partition level information is unavailable.

vmaheshw:
If a partition level information is unavailable, the task estimation and assignment can utilize topic level information. It will not accurate but will be approximate and better than the default information. This will help in balancing larger set of topic partitions and better balance.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
@@ -95,7 +95,12 @@ public class LoadBasedPartitionAssigner implements MetricsAware {
         LoadBasedPartitionAssignmentStrategyConfig.DEFAULT_PARTITION_MESSAGES_IN_RATE, "");
     newPartitions.forEach((task, partitions) -> {
       int totalThroughput = partitions.stream()
-          .mapToInt(p -> partitionInfoMap.getOrDefault(p, defaultPartitionInfo).getBytesInKBRate())
+          .mapToInt(p ->  {
+            // default partition will be -1.
+            String wildcardTopicPartition = p.substring(0, p.lastIndexOf('-')) + '-' + "-1";
+            PartitionThroughputInfo defaultValue = partitionInfoMap.getOrDefault(wildcardTopicPartition, defaultPartitionInfo);
+            return partitionInfoMap.getOrDefault(p, defaultValue).getBytesInKBRate();
+          })
           .sum();
       taskThroughputMap.put(task, totalThroughput);
     });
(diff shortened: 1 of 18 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
```diff
@@ -65,8 +65,11 @@ public int getTaskCount(ClusterThroughputInfo throughputInfo, List<String> assig
         LoadBasedPartitionAssignmentStrategyConfig.DEFAULT_PARTITION_MESSAGES_IN_RATE, "");
     // total throughput in KB/sec
     int totalThroughput = allPartitions.stream()
-        .map(p -> throughputMap.getOrDefault(p, defaultThroughputInfo))
-        .mapToInt(PartitionThroughputInfo::getBytesInKBRate)
+        .mapToInt(p ->  {
+          String wildcardTopicPartition = p.substring(0, p.lastIndexOf('-')) + "-" + "-1";
```
jzakaryan, line 69: Appending `--1` to the topic name seems a bit arbitrary. If the intent is to save a topic level throughput info in the same map, how about just using the topic name (without appended `--1`)?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -95,7 +95,12 @@
         LoadBasedPartitionAssignmentStrategyConfig.DEFAULT_PARTITION_MESSAGES_IN_RATE, "");
     newPartitions.forEach((task, partitions) -> {
       int totalThroughput = partitions.stream()
-          .mapToInt(p -> partitionInfoMap.getOrDefault(p, defaultPartitionInfo).getBytesInKBRate())
+          .mapToInt(p ->  {
+            // default partition will be -1.
+            String wildcardTopicPartition = p.substring(0, p.lastIndexOf('-')) + '-' + "-1";
```
jzakaryan, line 100: Same as previous comment. Why `--1`?

jzakaryan (changes requested):

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedTaskCountEstimator.java
vmaheshw: Thank you for the suggestion. I did not think from this angle. It will work. I have refactored the code. Thanks again.

jzakaryan (approved):

shrinandthakkar (approved):

PR #872, opened 2021-11-29
Flushless producer without comparing offsets

shrinandthakkar:
Instead of using a priority queue and comparing offsets to move the checkpoint, using a simple deque-based approach that removes the dependency on comparable offsets for any underlying pub-sub framework.

the contribution guidelines.


vmaheshw:
Can we discard this one, if no longer valid

PR #876, opened 2021-12-04
Loadbased Partition assigner not using topic level metrics to recognize partitions

vmaheshw:
The load based estimator is considering the topic level metrics, but the assigner code is not using the information to distribute the partitions. Fixing the code to derive the partition level metrics.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
@@ -51,6 +51,19 @@ public class LoadBasedPartitionAssigner implements MetricsAware {
 
   private final Map<String, LoadBasedPartitionAssigner.PartitionAssignmentStats> _partitionAssignmentStatsMap =
       new ConcurrentHashMap<>();
+  private final int _defaultPartitionBytesInKBRate;
+  private final int _defaultPartitionMsgsInRate;
+
+  /**
+   * Constructor of LoadBasedPartitionAssigner
+   * @param defaultPartitionBytesInKBRate default bytesIn rate in KB for partition
+   * @param defaultPartitionMsgsInRate default msgsIn rate in KB for partition
+   */
+  public LoadBasedPartitionAssigner(int defaultPartitionBytesInKBRate, int defaultPartitionMsgsInRate) {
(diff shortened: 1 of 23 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -127,7 +123,13 @@ public LoadBasedPartitionAssigner(int defaultPartitionBytesInKBRate, int default
       if (partitionInfoMap.containsKey(partition)) {
         recognizedPartitions.add(partition);
       } else {
-        unrecognizedPartitions.add(partition);
+        String topic = extractTopicFromPartition(partition);
```
surajkn, line 126: A brief comment here as we discussed.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -80,40 +80,36 @@ public LoadBasedPartitionAssigner(int defaultPartitionBytesInKBRate, int default
...
-    newPartitions.forEach((task, partitions) -> {
+    newPartitionAssignmentMap.forEach((task, partitions) -> {
       int totalThroughput = partitions.stream()
           .mapToInt(p ->  {
-            int index = p.lastIndexOf('-');
-            String topic = p;
-            if (index > -1) {
-              topic = p.substring(0, index);
-            }
+            String topic = extractTopicFromPartition(p);
```
surajkn, line 112: Probably not a issue but for my understanding.
Below (for unassigned partitions for loop) we check if a partition is present in partitinInfoMap and if not check if partition's topic is present in partitionInfoMap. Why dont we do it that way here?

surajkn (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
vmaheshw: We are doing it, see the default value. This loop is for the partitions already assigned. That's why it is of no value to update the map.

surajkn (approved):

shrinandthakkar (approved):

PR #877, opened 2021-12-13
Fix Stopping Logic and Maintain Stopping Latch Counter

shrinandthakkar:
Fixed stop logic by preventing to stop dead-thread tasks and also maintaining stopping latch counter irrespective of success or failure in acquiring task lock.

the contribution guidelines.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -93,7 +93,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected volatile long _lastPolledTimeMillis = System.currentTimeMillis();
   protected volatile long _lastPollCompletedTimeMillis = 0;
   protected final CountDownLatch _startedLatch = new CountDownLatch(1);
-  protected final CountDownLatch _stoppedLatch = new CountDownLatch(1);
+  private final CountDownLatch _stoppedLatch = new CountDownLatch(1);
   private final AtomicBoolean _metricDeregistered = new AtomicBoolean(false);
 
   // config
@@ -438,9 +438,13 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
         }
       }
       postShutdownHook();
-      _stoppedLatch.countDown();
+      countDownStoppedLatch();
       _logger.info("{} stopped", _taskName);
     }
+  }
+
+  protected void countDownStoppedLatch() {
+    _stoppedLatch.countDown();
   }
 
   /**
@@ -1098,5 +1102,10 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   @VisibleForTesting
   public String getConsumerAutoOffsetResetConfig() {
     return _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "");
+  }
+
+  @VisibleForTesting
+  public long getStoppedLatchCount() {
+    return _stoppedLatch.getCount();
   }
 }
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
@@ -352,6 +352,9 @@ public class KafkaMirrorMakerConnectorTask extends AbstractKafkaBasedConnectorTa
         LOG.info("Trying to acquire the lock on datastreamTask: {}", _datastreamTask);
         _datastreamTask.acquire(LOCK_ACQUIRE_TIMEOUT);
       } catch (DatastreamRuntimeException ex) {
+        // setting _stoppedLatch count to 0 since the lock couldn't be acquired,
+        // as a non-zero stoppedLatch value won't let the task to be stopped.
+        countDownStoppedLatch();
         LOG.error(String.format("Failed to acquire lock for datastreamTask %s", _datastreamTask), ex);
         _dynamicMetricsManager.createOrUpdateMeter(generateMetricsPrefix(_connectorName, CLASS_NAME), _datastreamName,
             TASK_LOCK_ACQUIRE_ERROR_RATE, 1);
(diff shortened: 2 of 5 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -1099,4 +1103,9 @@ public static String getKafkaGroupId(DatastreamTask task, GroupIdConstructor gro
   public String getConsumerAutoOffsetResetConfig() {
     return _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "");
   }
+
+  @VisibleForTesting
+  public long getStoppedLatchCount() {
```
vmaheshw, line 1108: is `public` required for this method?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -352,6 +352,9 @@ public void run() {
         LOG.info("Trying to acquire the lock on datastreamTask: {}", _datastreamTask);
         _datastreamTask.acquire(LOCK_ACQUIRE_TIMEOUT);
       } catch (DatastreamRuntimeException ex) {
+        // setting _stoppedLatch count to 0 since the lock couldn't be acquired,
+        // as a non-zero stoppedLatch value won't let the task to be stopped.
+        countDownStoppedLatch();
         LOG.error(String.format("Failed to acquire lock for datastreamTask %s", _datastreamTask), ex);
         _dynamicMetricsManager.createOrUpdateMeter(generateMetricsPrefix(_connectorName, CLASS_NAME), _datastreamName,
             TASK_LOCK_ACQUIRE_ERROR_RATE, 1);
```
vmaheshw, line 360: Can you also add a comment that this exception is fatal and should not swallowed.

vmaheshw (dismissed):

vmaheshw:
@shrinandthakkar Can you please drive this to completion?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
shrinandthakkar: yes I am using this method in TestKafkaMirrorMakerConnectorTask.java

vmaheshw (approved):
LGTM

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
ryannedolan: But that test is in the same package. This should not be public, should just be default (package) scope:

`long getStoppedLatchCount() {`

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -438,11 +438,15 @@ public void run() {
         }
       }
       postShutdownHook();
-      _stoppedLatch.countDown();
+      countDownStoppedLatch();
       _logger.info("{} stopped", _taskName);
     }
   }
 
+  protected void countDownStoppedLatch() {
```
ryannedolan, line 446: This is a weird api to expose to subclasses. How would an implr know when it is safe to invoke this?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -352,9 +352,14 @@ public void run() {
         LOG.info("Trying to acquire the lock on datastreamTask: {}", _datastreamTask);
         _datastreamTask.acquire(LOCK_ACQUIRE_TIMEOUT);
       } catch (DatastreamRuntimeException ex) {
+        // setting _stoppedLatch count to 0 since the lock couldn't be acquired,
+        // as a non-zero stoppedLatch value won't let the task to be stopped.
+        countDownStoppedLatch();
```
ryannedolan, line 357: The exception is being re-thrown anyway, so why not handle it in the super class? I don't like how tightly the super and sub classes are being coupled here.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
surajkn, line 446: @vmaheshw I remember you had expressed concern over this way of doing the count down. IIRC your concern was that any subclass of this can forget to call countDownStopppedLatch (although in this patch Shrinand added a call to it in KafkaMirrorMakerConnectorTask). So just wanted to check if that was considered and concluded that this is the most reasonable way of doing it?
cc @shrinandthakkar

vmaheshw: Yes, I was not in support of doing it this way, but also did not have a better idea. So, we decided to do it, until we have a better idea.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
shrinandthakkar: @vmaheshw @surajkn @ryannedolan 

Let me take this up in a separate PR but to reduce the coupling here, could we do something like this in the AbstractKafkaConnector(from where the task threads are initiated).
```   
  private ConnectorTaskEntry createKafkaConnectorTask(DatastreamTask task) {
    _logger.info("Creating connector task for datastream task {}.", task);
    AbstractKafkaBasedConnectorTask connectorTask = createKafkaBasedConnectorTask(task);
    Thread taskThread = createTaskThread(connectorTask);
    taskThread.start();

    // task cleanup thread cleans up the task resources like stopped latch counts post shutdown of the task
    Thread taskCleanUpThread = new Thread(() -> {
      try {
        taskThread.join();
      } catch (InterruptedException exception) {
        _logger.error(String.format("Got interrupted exception while waiting for the completion of task : %s ", connectorTask), exception);
      } finally {
        connectorTask.postShutdownCleanUp();
      }
    });
    taskCleanUpThread.start();
    return new ConnectorTaskEntry(connectorTask, taskThread);
  } 
```
the `postShutdownCleanUp` method here would take care of counting the latch down so we don't have to handle it in `run()` functions of either derived or base class implementations.

vmaheshw: Basically it will create extra thread for each thread. I think it should work, but with the constraint of double task threads.

surajkn (approved):

PR #881, opened 2022-01-24
Kafka upgrade

srinagaraj:
The interface ConsumerRebalanceListener has changed in the 2.4 release, a new method is  added onPartitionLost(), if this is not provided the default behavior is to call onPartitionRevoked(). basically on consumer close, its calling revoked partitions causing testValidateFlushlessModeTaskDiesOnRewindFailure to fail. This was not the behavior prior to this version.
Add fix to introduce a flag to hold the failure state and use it to no commit the changes on onPartitionRevoked call.

The Kafka team plans to deprecate open source li-apache-kafka-clients (http://github.com/linkedin/li-apache-kafka-clients) to streamline the client stack. Open source Brooklin currently depends on li-apache-kafka-clients and will need to directly depend on the LinkedIn version of open source Kafka clients (http://github.com/linkedin/kafka) instead before its full deprecation.
-->Auditor, and LargeMessageSegment support is deprecated.


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -432,7 +434,10 @@ public void run() {
     } finally {
       if (null != _consumer) {
         try {
+          _logger.error("Before consumer close");
           _consumer.close();
+          _logger.error("After consumer close and resetting failure flag");
+          _failure = false;
```
vmaheshw, line 440: Don't clear this flag here. `OnPartitionsRevoked` is called from another thread and it can result in commit().  Although the close does not complete before this method is called in new kafka versions, it is better to add fail-safe in our code, in case they change some behavior in the future.

Also, get rid of the debug logs.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/EmbeddedKafkaCluster.java
```diff
@@ -179,7 +180,7 @@ public String toString() {
     return sb.toString();
   }
 
-  static class SystemTime implements Time {
+  static class EmbeddedSystemTime extends SystemTime implements Time {
```
vmaheshw, line 183: What is the reason behind this change?

vmaheshw (changes requested):
Overall looks good, except one comment.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -90,6 +90,7 @@
   // lifecycle
   private volatile Thread _connectorTaskThread;
   protected volatile boolean _shutdown = false;
+  protected volatile boolean _failure = false;
```
vmaheshw, line 93: also, does this flag need to be `protected` or can it be changed to `private` ?

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/EmbeddedKafkaCluster.java
srinagaraj: SystemTime is introduced as an interface into kafka lib. So we need to create a different one to override the behaviour.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
srinagaraj: I am thinking not clearing this flag here would be taken care of when the thread starts again after close with _failure=false. T
Tested that the tests pass by removing this flag clearing.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
srinagaraj: We can make it private.

vmaheshw (approved):
LGTM

build.gradle
```diff
@@ -150,7 +150,6 @@ project(':datastream-server-api') {
   dependencies {
     compile project(':datastream-common')
     compile project(':datastream-utils')
-    compile "com.linkedin.kafka.clients:li-apache-kafka-clients:$LIKafkaVersion"
```
vmaheshw, line 153: Just curious, why is this line getting removed instead of getting replaced with "com.linkedin.kafka:kafka-clients..."

vmaheshw (approved):

build.gradle
srinagaraj: There is no Open source version of this package. Its completely moved inside of Li. That is why we need to remove the dependency on this package. Will add more details to this rb. Thanks

atoomula (approved):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -766,7 +769,7 @@ protected void updateConsumerAssignment(Collection<TopicPartition> partitions) {
   public void onPartitionsRevoked(Collection<TopicPartition> topicPartitions) {
```
vmaheshw, line 769: `KafkaMirrorMakerConnectorTask` is overriding this method. You will have to fix this logic there as well. Without that this change is incomplete.

srinagaraj: KafkaMirrorMakerConnectorTask is calling super.Same with onPartitionsRevoked.  So I am thinking change in AbstractKafkaBasedConnectorTask should take care of it. COuld you pls confirm? 

 @Override
  public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
    super.onPartitionsRevoked(partitions);
    _topicManager.onPartitionsRevoked(partitions);
  }
@override
  public void run() {
    if (_enablePartitionAssignment) {
      try {
        LOG.info("Trying to acquire the lock on datastreamTask: {}", _datastreamTask);
        _datastreamTask.acquire(LOCK_ACQUIRE_TIMEOUT);
      } catch (DatastreamRuntimeException ex) {
        LOG.error(String.format("Failed to acquire lock for datastreamTask %s", _datastreamTask), ex);
        _dynamicMetricsManager.createOrUpdateMeter(generateMetricsPrefix(_connectorName, CLASS_NAME), _datastreamName,
            TASK_LOCK_ACQUIRE_ERROR_RATE, 1);
        throw ex;
      }
    }
    super.run();
  }

vmaheshw: At the current code state, yes, it will take care of it. But, it also has capability to create bigger problems in future. By not handling in the override class and relying that the logic is present in super class, in the future if someone changed the override class and remove the super call,  it will expose this problem. Since the current logic is very much dependent on how the code is written and it can change in future, it is critical to handle the scenario for easier debugging in future.

There are two ways to look at this:
a. TopicManager.onPartitionsRevoked will still be called which can add additional processing time based on what the implementation of that interface is. If it is decided that the subsequent calls should not be made, then additional checks are required in main class as well.
b. If it is not critical and we want to let the future users know about it, it is better to add a comment in KafkaMirrorMakerConnectorTask to not remove super.onPartitionsRevoked call or refactor the code to take care of the new kafka client scenario. 

I am very sure this knowledge will get lost in the future if there are not enough hints in the code.

srinagaraj: _topicManager.OnPartitionsRevoked is pretty much a noOp today. May be we could revisit this  overridden code onPartitionsRevoked in KafkaMirrorMakerConnectorTask later if needed completly.
For now, we can add comments to explicitly state the changes in upgraded kafka even in the overridden class as mentioned in the base class for clarity.
I will add this into a separate PR following this. Thanks!

PR #885, opened 2022-02-07
Added new datastream metadata field to include allowlist regex

shrinandthakkar:
Added allow-listing config in datastream metadata for easy access and configuration in BMM based operations. Until now the allow-listing value was appended into the connection string for the Kafka pub-sub system, but to use it for any other pub-sub systems differently, I wanted to expose that as separate metadata instead of doing string manipulations in the server code.

the contribution guidelines.


Tim-McNally:
Should we not also put the denylist in the metadata as well if we are putting the allowlist there?

shrinandthakkar:
hey @thomaslaw, until now the allowlisting value was appended into the connection string of the Kafka pub-sub system, but to use it for any other pub-sub systems differently, I wanted to expose that as separate metadata.

shrinandthakkar:
@Tim-McNally I am not sure if that is needed as of now. Will add that later if required.

PR #889, opened 2022-03-01
Add a test to catch minPartitions and maxPartitions per task issue

vmaheshw:
Add a test to catch minPartitions and maxPartitions per task issue


ryannedolan:
@vmaheshw can you rebase?

PR #890, opened 2022-03-03
Upgrade avro and move jackson from codehaus to fasterxml

vmaheshw:
Avro 1.9.2 upgrade pulls in fasterxml jackson versions which brings conflicting classes. Migrating the jackson classes from codeHause to fasterxml


ryannedolan:
diff looks good but commit log needs to be cleaned up. Maybe you need to rebase?

srinagaraj (approved):

PR #900, opened 2022-03-25
fix rebalancing-tasks bug and added tests

shrinandthakkar:
### Problem: 
Both the LoadBasedPartition and StickyPartition Strategy follow the StickyMulticastStrategy to assign the datastream tasks across the available instances. 

In the StickyMulticastStrategy, the last step of the process is to rebalance the tasks across instances such that the number of tasks across the instances adheres to the `imbalance threshold` value.  

> _Basically, the difference between the max #tasks assigned to any instance and the min #tasks assigned to any other instance should not exceed the imbalance threshold_. 

In the current implementation, there was a minor bug, which in some cases failed to achieve that proper rebalance. That issue causes validation failures and thus availability misses on the mirror-making service. 
____
**Failure Scenario:** Observed in production service with 6 instances connected to ZK and has 538 tasks in total.

We saw from the production logs that the tasks were distributed across the instances with counts: _89, 89, 89, 89, 93, 89_ 

And given that we enforce the imbalance threshold of 1, the actual distribution should have been something like 
_89, 89, 90, 90, 90, 90_

In the below pasted rebalance code, 

_tasksTotal = 538_
_minTasksPerInstance = 538 / 6 = 89_ 
And since the minimum number of tasks assigned == minTasksPerInstance, it won't go in the while loop to rebalance.

```
// STEP 3: Trigger rebalance if the number of different tasks more than the configured threshold

if (newAssignment.get(instancesBySize.get(instancesBySize.size() - 1)).size() - newAssignment.get(instancesBySize.get(0)).size() > _imbalanceThreshold) {
 int tasksTotal = newAssignment.values().stream().mapToInt(Set::size).sum();
 int minTasksPerInstance = tasksTotal / instances.size();

 // some rebalance to increase the task count in instances below the minTasksPerInstance
 while (newAssignment.get(instancesBySize.get(0)).size() < minTasksPerInstance) {
   ...
 }
}
```
____
### Solution:
Update the rebalancing code condition to only terminate when the **min #tasks assigned + the imbalance threshold is less than the max #tasks assigned**.

_____
the contribution guidelines.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyMulticastStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyMulticastStrategy.java
@@ -205,7 +205,8 @@ public class StickyMulticastStrategy implements AssignmentStrategy {
       int minTasksPerInstance = tasksTotal / instances.size();
 
       // some rebalance to increase the task count in instances below the minTasksPerInstance
-      while (newAssignment.get(instancesBySize.get(0)).size() < minTasksPerInstance) {
+      while (newAssignment.get(instancesBySize.get(0)).size() + _imbalanceThreshold < newAssignment.get(
+          instancesBySize.get(instancesBySize.size() - 1)).size()) {
         String smallInstance = instancesBySize.get(0);
         String largeInstance = instancesBySize.get(instancesBySize.size() - 1);
 
--- a/datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
+++ b/datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
@@ -7,9 +7,11 @@ package com.linkedin.datastream.server.assignment;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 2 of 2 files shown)


datastream-server/src/test/java/com/linkedin/datastream/server/assignment/TestStickyMulticastStrategy.java
```diff
@@ -499,6 +501,62 @@ public void testExtraTasksAreNotAssignedDuringReassignment() {
     }
   }
 
+  @Test
+  public void testReBalancingTasksWithThreshold() {
+    String[] instances = new String[]{"instance1", "instance2", "instance3"};
+    reBalancingTasksWithThresholdHelper(instances, new int[]{9, 9, 13}, 1);
```
vmaheshw, line 507: Suggestion to add some examples with even number of instances and only 1 instance.

vmaheshw (commented):
Can you improve the description explaining the problem in detail and how are you solving the problem?

shrinandthakkar:
> Can you improve the description explaining the problem in detail and how are you solving the problem?

Done

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyMulticastStrategy.java
```diff
@@ -205,7 +205,8 @@ public StickyMulticastStrategy(Optional<Integer> maxTasks, int imbalanceThreshol
       int minTasksPerInstance = tasksTotal / instances.size();
 
       // some rebalance to increase the task count in instances below the minTasksPerInstance
-      while (newAssignment.get(instancesBySize.get(0)).size() < minTasksPerInstance) {
+      while (newAssignment.get(instancesBySize.get(0)).size() + _imbalanceThreshold < newAssignment.get(
```
ryannedolan, line 208: This is a lot to parse in one line. Consider pulling out into a couple vars so the logic is more obvious.

ryannedolan (approved):

PR #903, opened 2022-04-17
Introduce broadcast API to TransportProvider

surajkn:
This API will allow to sending events to all consumers/clients of TransportProvider.
TransportProvider will be able to optionally implement broadcast API.
Example, for Kafka TransportProvider broadcast will send the event to all topic partitions.

the contribution guidelines.


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +143,38 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
...
+      listMetadataExceptionPair.add(new Pair<>(null, ex));
+      onSendComplete.onCompletion(listMetadataExceptionPair, partitionCount);
+      return;
+    }
+
+    LOG.debug("Broadcasting record {} to all {} partitions of destination {}", record, partitionCount, destinationUri);
+    for (int i = 0; i < partitionCount; i++) {
+      record.setPartition(i);
+      send(destinationUri, record, ((metadata, exception) -> {
+        listMetadataExceptionPair.add(new Pair<>(metadata, exception));
```
vmaheshw, line 172: Can you encapsulate the list from per partition, to just one? It will simplify the code.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +143,38 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
...
+      }));
+    }
+    onSendComplete.onCompletion(listMetadataExceptionPair, partitionCount);
```
vmaheshw, line 175: This should be inside the send() code, otherwise the oncompletion will get called before send completes.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
surajkn: Discussed this with Vaibhav. Will simplify the aggregation of send callback  results and make the broadcast callback simpler.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTestUtils.java
```diff
@@ -122,6 +133,40 @@ public static boolean topicExists(AdminClient adminClient, String topic) {
     return false;
   }
 
+  /**
+   * Consume broadcast messages from a given partition of a Kafka topic, using given BroadcastReaderCallback
```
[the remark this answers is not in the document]
surajkn: This method reads all the events from the destination topic and helps compare the read events to the send events. Basically you can use this method to read what you sent and compare any data points you might want.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +143,38 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
...
+      // Extract partition count from the topic using AdminClient
+      partitionCount =
+          adminClient.describeTopics(Collections.singleton(topicName)).all().get().get(topicName).partitions().size();
+    } catch (ExecutionException | InterruptedException ex) {
+      LOG.error("Failed to parse partition count for topic {} required for broadcast", topicName);
+      listMetadataExceptionPair.add(new Pair<>(null, ex));
+      onSendComplete.onCompletion(listMetadataExceptionPair, partitionCount);
+      return;
+    }
+
```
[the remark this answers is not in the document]
surajkn: Refactored as discussed. Pls review the updated logic.

datastream-server-api/src/main/java/com/linkedin/datastream/server/api/transport/TransportProvider.java
```diff
@@ -38,4 +38,28 @@ public interface TransportProvider {
...
+   * @param destination the destination topic to which the record should be broadcasted.
+   * @param record DatastreamEvent that needs to be broadcasted to the stream.
+   * @param onBroadcastComplete Callback that needs to be called when the broadcast attempt is complete. Any exception
+   *                            during broadcast attempt will be reported through callback. A callback must be called at
+   *                            the end of broadcast attempt
+   * @param onEventComplete Callback that will be called at send completion to each endpoint. This is an optional
+   *                        callback. For e.g., for Kafka this callback would be invoked when send to each partition
+   *                        completes.
+   */
+  default void broadcast(String destination, DatastreamProducerRecord record, SendCallback onBroadcastComplete,
```
vmaheshw, line 61: if it is a best-effort strategy, then what is the need for SendCallback. If book-keeping is the purpose, why do you need two callbacks? The interface is confusing.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTestUtils.java
```diff
@@ -45,6 +45,17 @@ public interface ReaderCallback {
     boolean onMessage(byte[] key, byte[] value) throws IOException;
   }
 
+  /**
+   * Interface for the callback invoked whenever broadcast messages are read
+   */
+  public interface BroadcastReaderCallback {
```
vmaheshw, line 51: Is this code still relevant?

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -200,13 +200,24 @@ private void validateEventRecord(DatastreamProducerRecord record) {
...
   /**
    * Send the event onto the underlying transport.
    * @param record the datastream event
-   * @param sendCallback the callback to be invoked after the event is sent to the destination
+   * @param sendBroadcastCallback the callback to be invoked after the event is broadcasted to the destination
+   * @param sendEventCallback the callback to be invoked after the event is sent to the destination
    */
-  @Override
-  public void send(DatastreamProducerRecord record, SendCallback sendCallback) {
+  public void helperSendOrBroadcast(DatastreamProducerRecord record, SendCallback sendBroadcastCallback,
```
ryannedolan, line 219: public?

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -231,9 +242,20 @@ public void send(DatastreamProducerRecord record, SendCallback sendCallback) {
...
       long recordEventsSourceTimestamp = record.getEventsSourceTimestamp();
       long recordEventsSendTimestamp = record.getEventsSendTimestamp().orElse(0L);
-      _transportProvider.send(destination, record,
-          (metadata, exception) -> onSendCallback(metadata, exception, sendCallback, recordEventsSourceTimestamp,
-              recordEventsSendTimestamp));
+      if (isBroadcast) {
+        _transportProvider.broadcast(destination, record, (metadata, exception) -> {
+          onBroadcastCallback(metadata, exception, sendBroadcastCallback);
+        }, ((metadata, exception) -> {
+          if (exception == null) {
```
ryannedolan, line 249: seems this would always be null.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +140,40 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
...
+            onEventComplete.onCompletion(metadata, exception);
+          }
+        }));
+        sentToPartitions.add(partition);
+      }
+      doOnBroadcastCallback(record, onBroadcastComplete, sentToPartitions, topicName, partitionCount, null);
+    } catch (DatastreamRuntimeException ex) {
+      LOG.error("Broadcast send failed for record {} at partition {}/{} because of exception: {} ",
+          record, partition, partitionCount, ex);
+      doOnBroadcastCallback(record, onBroadcastComplete, sentToPartitions, topicName, partitionCount, ex);
```
ryannedolan, line 173: Looks like the only exceptions we pass to the onBroadcastCallback is if send() throws (e.g. serde exceptions). But that doesn't capture the async exceptions that send() provides via callbacks (e.g. delivery timeouts). So we're swallowing any async exceptions.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -200,13 +200,24 @@ private void validateEventRecord(DatastreamProducerRecord record) {
     }
   }
 
+  @Override
+  public void broadcast(DatastreamProducerRecord record, SendCallback sendBroadcastCallback) {
```
ryannedolan, line 204: Seems there is no way for caller to know whether some/all/none of the msgs succeeded/failed.

ryannedolan (changes requested):
Think there are some bugs wrt exceptions and callbacks.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +140,40 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
         event.getMetadata().getOrDefault(BrooklinEnvelopeMetadataConstants.SOURCE_PARTITION, "-1"));
   }
 
+  @Override
+  public void broadcast(String destinationUri, DatastreamProducerRecord record,
+      SendCallback onBroadcastComplete, SendCallback onEventComplete) {
```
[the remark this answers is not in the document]
surajkn: With Future the responsibility will call on the caller (i.e. on the connector through EventProducer since most connectors call into TP through EventProducer), which will make connector logic (or probably EventProducer) more complicated where it will have to call this Future object. Probably in a separate thread so as to not make the send/broadcast blocking. 

Right now the most we do in callback is update metrics (in EventProducer) or log errors. Which makes it a good candidate for it being as a callback instead of Future. But thats just my understanding. 

In the current change above, broadcast callback is invoked when attempt to send to all partitions complete i.e. we call send for the record for each partition and in the broadcast callback we pass information of the total number of partitions and partitions to which send succeeded. However we dont wait for the send callback itself to complete.
onEventCompelete callback is the callback equivalent in send method. It is actually passed as the callback to send when we call send from broadcast.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +140,40 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
...
+    Validate.isTrue(record.isBroadcastRecord(), "Trying to broadcast a non-broadcast type record.");
+
+    // Currently destination topic partition count will be queried from datastream destination metadata. This implies
+    // a restriction that datastream destination partition should be updated for broadcast to work correctly.
+    int partitionCount = _datastreamTask.getDatastreamDestination().getPartitions();
```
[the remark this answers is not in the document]
surajkn: For now we are doing with a limitation that for broadcast to work correctly we expect the partition count in datastream destination metadata to be correctly updated. 

I had some logic around dynamically querying partition count using AdminClient but Vaibhav mentioned that it will be expensive to do if broadcast is used by a client as frequently as send. Apparently for low level BMM we handle this by having a separate thread where we periodically check for partition count (at 60 sec interval). In future we can do that as an enhancement and live with the restriction for now. It should be fine for now since partition count is not a frequent operation.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +140,40 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
...
+
+    LOG.debug("Broadcasting record {} to all {} partitions of destination {}", record, partitionCount, destinationUri);
+    int partition = 0;
+    List<Integer> sentToPartitions = new ArrayList<>();
+    try {
+      for (; partition < partitionCount; partition++) {
+        record.setPartition(partition);
+        send(destinationUri, record, ((metadata, exception) -> {
+          if (exception != null) {
+            LOG.error("Failed to broadcast record {} to partition {}", record, metadata.getPartition());
```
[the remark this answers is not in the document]
surajkn: Yeah I missed that. I should be invoking the callback here too. 

The partitions where the record was sent to as part of broadcast, we put this information in the DatastreamRecordMetadata in the broadcast callback.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
surajkn: The onEventCallback is where it would be handled.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
surajkn: Will make it private, thanks for pointing out. This should not be public.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -201,26 +201,28 @@ private void validateEventRecord(DatastreamProducerRecord record) {
...
    */
-  public void helperSendOrBroadcast(DatastreamProducerRecord record, SendCallback sendBroadcastCallback,
+  private DatastreamRecordMetadata helperSendOrBroadcast(DatastreamProducerRecord record,
       SendCallback sendEventCallback, boolean isBroadcast) {
+    DatastreamRecordMetadata broadcastMetadata = null;
+
     try {
       validateEventRecord(record);
-
       try {
```
ryannedolan, line 226: nested try? can probably simplify and have one try-catch pair per method.

ryannedolan (approved):
some nits but looks good

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
surajkn: It seems a little problematic to refactor this in a neat way.

The try block for serialization error has a catch all Exception and I wan't able to identify the correct set of specific exceptions that might occur, so I introduced a flag to differentiate the exception in the catch block. Other alternative if to have multiple try blocks instead of nested try block. I am personally not partial to either adding a flag or nested try block. But pls let me know if there is a cleaner way that I might have overlooked.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -267,6 +272,7 @@ public void helperSendOrBroadcast(DatastreamProducerRecord record, SendCallback
     if (_enableFlushOnSend && Instant.now().isAfter(_lastFlushTime.plus(_flushInterval))) {
       flush();
     }
+    return broadcastMetadata;
```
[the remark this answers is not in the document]
surajkn: This one is a little tricky. We do a conditional flush just above this return but we do broadcast above in the try block. And even there "send" has a void return type so I have to initialize "broadcastMetadata" to null. So probably seems cleaner to simply return broadcastMetadata. Also I have removed conditional assignment from the catch block of serialization error so hopefully it improved the readability a little. 

Other alternative is to not try to abstract out the common code between send and broadcast by doing this helper method and completely separate out the two. Let me know if overall that seems cleaner.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -369,6 +402,14 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
     _dynamicMetricsManager.createOrUpdateMeter(MODULE, _datastreamTask.getConnectorType(), EVENT_PRODUCE_RATE, 1);
   }
 
+  private void onBroadcastCallback(DatastreamRecordMetadata metadata, Exception exception, SendCallback sendCallback) {
```
vmaheshw, line 405: Is this method getting used in this code anywhere?

datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamProducerRecord.java
```diff
@@ -117,9 +124,22 @@ public Optional<Integer> getPartition() {
     return _partition;
   }
 
+  /**
+   * Set partition for the record. Can be used to set partition for broadcast record each time before sending.
+   * @param partition Topic partition to send the record to.
+   */
+  public void setPartition(int partition) {
```
ryannedolan, line 131: I think we need to keep this class immutable. Are you just trying to avoid a copy? I think that's not worth the broken semantics.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -136,6 +140,42 @@ private int getSourcePartitionFromEvent(BrooklinEnvelope event) {
...
+    // a restriction that datastream destination partition should be updated for broadcast to work correctly.
+    int partitionCount = _datastreamTask.getDatastreamDestination().getPartitions();
+    String topicName = KafkaTransportProviderUtils.getTopicName(destinationUri);
+
+    LOG.debug("Broadcasting record {} to all {} partitions of destination {}", record, partitionCount, destinationUri);
+    int partition = 0;
+    List<Integer> sentToPartitions = new ArrayList<>();
+    try {
+      for (; partition < partitionCount; partition++) {
+        record.setPartition(partition);
```
ryannedolan, line 157: I think modifying the record like this will get us into trouble. The caller would not expect the object to be modified, and the internal send() is probably expecting an immutable object. For example, what if send() were to buffer the record before looking at the partition?

I think we need to construct a new record for each send, either by shallow-copying into a new record (which is dangerous, cuz the fields to copy may change over time), or by wrapping the record in an outer class that overrides getPartition. Then you could do:

`send(new DatastreamProducerRecordWithPartititionOverride(record, newPartition))`

or smth.

ryannedolan (changes requested):

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
surajkn: Went the copy route

ryannedolan (approved):
thanks, lgtm!

datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamProducerRecord.java
```diff
@@ -20,11 +20,12 @@
  * List of brooklin events that need to be sent to the {@link com.linkedin.datastream.server.api.transport.TransportProvider}
  */
 public class DatastreamProducerRecord {
-  private final Optional<Integer> _partition;
+  private Optional<Integer> _partition;
```
vmaheshw, line 23: This change is not required.

vmaheshw (approved):

vmaheshw (approved):

PR #906, opened 2022-05-02
Dedupe tasks on LeaderDoAssignment

jogrogan:
Fixes crashing during `LeaderDoAssignment` by deduplicating tasks that contain the same task name.
This can happen if a leader change happens while saving an existing assignment change is already in progress.
If we reach this case, select one of the tasks as valid and remove the others.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
@@ -88,8 +88,8 @@ public class BroadcastStrategy implements AssignmentStrategy {
     Map<String, List<DatastreamTask>> reuseTaskMap =
         tasksAvailableToReuse.stream().collect(Collectors.groupingBy(DatastreamTask::getTaskPrefix));
     int instancePos = 0;
-    Set<String> uniqueDatastreamTaskNamesSet = new HashSet<>();
     for (DatastreamGroup dg : datastreams) {
+      Set<String> uniqueDatastreamTaskNamesSet = new HashSet<>();
       List<DatastreamTask> reuseTasksPerDg = reuseTaskMap.getOrDefault(dg.getTaskPrefix(), Collections.emptyList());
 
       int numTasks = getNumTasks(dg, instances.size());
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
@@ -88,11 +88,13 @@ public class BroadcastStrategy implements AssignmentStrategy {
     Map<String, List<DatastreamTask>> reuseTaskMap =
         tasksAvailableToReuse.stream().collect(Collectors.groupingBy(DatastreamTask::getTaskPrefix));
     int instancePos = 0;
+    Set<String> uniqueDatastreamTaskNames = new HashSet<>();
     for (DatastreamGroup dg : datastreams) {
       List<DatastreamTask> reuseTasksPerDg = reuseTaskMap.getOrDefault(dg.getTaskPrefix(), Collections.emptyList());
 
       int numTasks = getNumTasks(dg, instances.size());
-      for (int taskPos = 0; taskPos < numTasks; taskPos++) {
+      int taskPos = 0;
+      while (taskPos < numTasks) {
         String instance = instances.get(instancePos);
 
         DatastreamTask foundDatastreamTask = currentAssignmentCopy.get(instance)
@@ -102,10 +104,16 @@ public class BroadcastStrategy implements AssignmentStrategy {
             .orElseGet(() -> getOrCreateDatastreamTask(reuseTasksPerDg, dg));
 
         currentAssignmentCopy.get(instance).remove(foundDatastreamTask);
+
+        // Prevent duplicate datastream task names
+        if (!uniqueDatastreamTaskNames.add(foundDatastreamTask.getDatastreamTaskName())) {
+          continue;
+        }
         newAssignment.get(instance).add(foundDatastreamTask);
 
-        // Move to the next instance
+        // Move to the next task and instance
         instancePos = (instancePos + 1) % instances.size();
+        taskPos++;
       }
     }
 
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
@@ -88,7 +88,7 @@ public class BroadcastStrategy implements AssignmentStrategy {
     Map<String, List<DatastreamTask>> reuseTaskMap =
         tasksAvailableToReuse.stream().collect(Collectors.groupingBy(DatastreamTask::getTaskPrefix));
     int instancePos = 0;
-    Set<String> uniqueDatastreamTaskNames = new HashSet<>();
+    Set<String> uniqueDatastreamTaskNamesSet = new HashSet<>();
     for (DatastreamGroup dg : datastreams) {
       List<DatastreamTask> reuseTasksPerDg = reuseTaskMap.getOrDefault(dg.getTaskPrefix(), Collections.emptyList());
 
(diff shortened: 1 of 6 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -736,23 +736,36 @@ private void removeTaskNode(String instance, String name) {
    * act on the differences — that is: add new assignments, remove old assignments.
    * For ones that didn't change, do nothing.
    *
+   * Can receive duplicate task assignment. In this case select one of the tasks as
```
shrinandthakkar, line 739: If I understand this correctly,

In the `updateAllAssignments`, the leader coordinator adds tasks initially ([code link here](https://github.com/linkedin/brooklin/blob/87c7bec012a776cbb4a5be7d40903682be2d763e/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java#L758)) to the other instance based on the newly generated assignment. But if that leader instance crashes before removing that task from the previously assigned instance, there could be two different instances handling that same task.

And if that is the issue, I would prefer that we have a fix for that in the `performAssignment` method (in all implementations of this method) since I feel that although deduping it in the ZK code path would fix the issue for this case, but I think this should be handled by the `performAssignment` method. In the future, there could be other code paths that could use the `performAssignment` function, and they'll return a bad assignment (with the same tasks in two instances.)

And I think @vmaheshw has already fixed it for the StickyMulticastStrategy (https://github.com/linkedin/brooklin/pull/749) so maybe other strategies need a similar fix, or it's a different issue?

shrinandthakkar (changes requested):

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jogrogan: I believe it is a bit more nuanced than this. If a leader crash happens while saving an assignment, a new leader could come in and reuse a unique task name and create a whole new DataStreamTask with the same name. Thus the case that the StickyMulticastStrategy solves wouldn't matter here because doing an equality check between the two DataStreamTasks would not pass but a crash would still happen later when we expect a set of unique task names

shrinandthakkar: Reusing a unique task name for creating a newer task should be super rare(atleast in all strategies except LoadBalanced), which is not as probable as the issue described in the previous comment I feel, where the same task could land upon multiple hosts.

However, even if that happens, I would prefer the strategy handles cleaning it up instead of cleaning it in the ZK code path.

jogrogan: Updated the assignment of each strategy

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadbalancingStrategy.java
```diff
@@ -101,14 +101,19 @@ private List<DatastreamTask> getDatastreamTasks(List<DatastreamGroup> datastream
     List<DatastreamTask> currentlyAssignedDatastreamTasks = new ArrayList<>();
     currentAssignment.values().forEach(currentlyAssignedDatastreamTasks::addAll);
 
+    Set<String> uniqueDatastreamTaskNames = new HashSet<>();
     for (DatastreamGroup dg : datastreams) {
       Set<DatastreamTask> tasksForDatastreamGroup = currentlyAssignedDatastreamTasks.stream()
-          .filter(x -> x.getTaskPrefix().equals(dg.getTaskPrefix()))
+          .filter(x -> x.getTaskPrefix().equals(dg.getTaskPrefix()) && uniqueDatastreamTaskNames.add(x.getDatastreamTaskName()))
           .collect(Collectors.toSet());
```
shrinandthakkar, line 108: Since we are collecting all of them in a set anyway, why do we need a separate set to handle this?

shrinandthakkar (changes requested):

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadbalancingStrategy.java
jogrogan: Checking task name equality is a more general check than checking equality of two datastreams which will check a suite of fields associated with the datastreams

shrinandthakkar: I am good with either of those. But my only concern is that the test you have won't cover the case when the datastream tasks have different internal fields and the same task names. And that test would pass irrespective of the change you have in this PR for load balancing strategy.

But in my opinion, we would never hit that scenario for this issue since the previous leader would have copied the task to the other instance entirely, so all the internal fields should also match anyways.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -102,10 +104,16 @@ public Map<String, Set<DatastreamTask>> assign(List<DatastreamGroup> datastreams
             .orElseGet(() -> getOrCreateDatastreamTask(reuseTasksPerDg, dg));
 
         currentAssignmentCopy.get(instance).remove(foundDatastreamTask);
+
+        // Prevent duplicate datastream task names
+        if (!uniqueDatastreamTaskNames.add(foundDatastreamTask.getDatastreamTaskName())) {
```
vmaheshw, line 109: If a duplicate task name is found, can you not create a new one right here. You don't have to skip continue.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadbalancingStrategy.java
```diff
@@ -101,14 +101,19 @@ private List<DatastreamTask> getDatastreamTasks(List<DatastreamGroup> datastream
     List<DatastreamTask> currentlyAssignedDatastreamTasks = new ArrayList<>();
     currentAssignment.values().forEach(currentlyAssignedDatastreamTasks::addAll);
 
+    Set<String> uniqueDatastreamTaskNames = new HashSet<>();
```
vmaheshw, line 104: making this variable local to datastream group will eliminate the need of added code on line 113.

Also, can you rename with set

jogrogan: Is there any scenario where we could have a duplicate DatastreamGroup? It looks to be a list and not a set

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
jogrogan: This will create a new task. By moving this to a while loop and manually incrementing the taskPos we essentially skip the duplicate by not increasing the taskPos variable and instead create a new datastream on a following iteration. I don't think the code would be very clean if we wanted to duplicate that same logic inside the conditional

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadbalancingStrategy.java
jogrogan: Added tests for different internal fields

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/BroadcastStrategy.java
```diff
@@ -102,10 +104,16 @@ public Map<String, Set<DatastreamTask>> assign(List<DatastreamGroup> datastreams
             .orElseGet(() -> getOrCreateDatastreamTask(reuseTasksPerDg, dg));
 
         currentAssignmentCopy.get(instance).remove(foundDatastreamTask);
+
+        // Prevent duplicate datastream task names
+        if (!uniqueDatastreamTaskNamesSet.add(foundDatastreamTask.getDatastreamTaskName())) {
```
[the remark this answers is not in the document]
jogrogan: I don't think the code would be very clean if I added this. The instance will still get assigned a task on the next iteration  after `continue` while still utilizing the `reuseTaskMap`. If I instead just created a new task now we may not be reusing tasks optimally.

vmaheshw (approved):

shrinandthakkar (approved):

PR #914, opened 2022-09-28
Version 4.2.0-SNAPSHOT

ryannedolan:
Bumping version

### DIFF
--- a/gradle/maven.gradle
+++ b/gradle/maven.gradle
@@ -1,5 +1,5 @@
 allprojects {
-  version = "4.1.0"
+  version = "4.2.0-SNAPSHOT"
 }
 
 subprojects {
--- a/gradle/maven.gradle
+++ b/gradle/maven.gradle
@@ -1,5 +1,5 @@
 allprojects {
-  version = "4.0.0"
+  version = "4.1.0"
 }
 
 subprojects {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,66 @@
 All notable changes to this project will be documented in this file.  
 This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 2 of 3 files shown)


vmaheshw (approved):

surajkn (approved):

gradle/maven.gradle
```diff
@@ -1,5 +1,5 @@
 allprojects {
-  version = "4.0.0"
+  version = "4.2.0-SNAPSHOT"
```
vmaheshw, line 2: why a snapshot version?

ryannedolan: Per wiki https://github.com/linkedin/brooklin/wiki/Developer-Guide#update-mavengradle and per what's typical. Master/trunk should be a SNAPSHOT of the next release. Can ignore the intermediate 4.1.0 commit -- I'll squash it anyway.

Full disclosure, I tweaked the wiki there too, cuz it used to say to point master to the next PATCH release (which was wrong).

vmaheshw: So, while releasing, you will change this version and then make the release so that version 4.2.0 is released. Or the new version will be 4.2.0-SNAPSHOT, which will be referenced everywhere else.

vmaheshw (approved):

gradle/maven.gradle
ryannedolan: Right. During release process you just drop the SNAPSHOT part on your local machine. Then build and publish. Then increment the version to <next.version.0>-SNAPSHOT and submit a PR for the change. I had to do it a bit out of order this time bc the version in master was actually the previous release (should be the next release snapshot).

hshukla (approved):

jogrogan (approved):

PR #915, opened 2022-10-31
Add support for post datastream create,update,delete and state change action

hshukla:
- `Connector::postDatastreamStateChangeAction(Datastream stream)` Added to expose the hook for post datastream add/update/delete and state change action. The default has no implementation, any interested connectors need to add support by implementing this method.

- In` DatastreamResource`, call this newly added method in Coordinator after create, update, delete, state change(Pause, Resume, Stop).

- Add unit test to handle this new method functionality.

### DIFF
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -500,6 +500,14 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
 
     LOG.info("Completed request for stopping datastream {}", _store.getDatastream(datastream.getName()));
 
+    // above check guarantees us that datastreams will be in STOPPED state only, so need to fetch the latest state
+    // from store
+    datastreamsToStop.forEach(ds -> {
+      // set the status as STOPPED as the original status would be READY, PAUSED or STOPPING
+      ds.setStatus(DatastreamStatus.STOPPED);
+      invokePostDSStateChangeAction(ds);
+    });
+
     return new ActionResult<>(HttpStatus.S_200_OK);
   }
 
@@ -940,14 +948,14 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
     return null;
   }
 
-  private void invokePostDSStateChangeAction(Datastream datastream) throws DatastreamException {
+  private void invokePostDSStateChangeAction(Datastream datastream) {
     try {
       LOG.debug("Invoke post datastream state change action datastream={}", datastream);
       _coordinator.invokePostDataStreamStateChangeAction(datastream);
       LOG.info("Invoked post datastream state change action datastream={}", datastream);
     } catch (DatastreamException e) {
-      LOG.error("Failed to perform post datastream state change action datastream={}", datastream, e);
-      throw e;
+      _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_500_INTERNAL_SERVER_ERROR,
+          "Failed to perform post datastream state change action datastream=" + datastream, e);
     }
   }
 
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -500,8 +500,8 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
 
     LOG.info("Completed request for stopping datastream {}", _store.getDatastream(datastream.getName()));
 
-    // above check guarantees us that datastreams will be in STOPPED state only, so need to fetch the latest state
-    // from store
+    // above check guarantees us that datastreams will be in STOPPED state only, so not need to fetch the latest state
+    // from store explicitly
     datastreamsToStop.forEach(ds -> {
       // set the status as STOPPED as the original status would be READY, PAUSED or STOPPING
       ds.setStatus(DatastreamStatus.STOPPED);
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -742,7 +742,6 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
 
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, DELETE_CALL, 1);
       Instant startTime = Instant.now();
-      datastream.setStatus(DatastreamStatus.DELETING);
       _store.deleteDatastream(datastreamName);
       DELETE_CALL_LATENCY_MS.set(Duration.between(startTime, Instant.now()).toMillis());
 
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -734,6 +734,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
 
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, DELETE_CALL, 1);
       Instant startTime = Instant.now();
+      datastream.setStatus(DatastreamStatus.DELETING);
       _store.deleteDatastream(datastreamName);
       DELETE_CALL_LATENCY_MS.set(Duration.between(startTime, Instant.now()).toMillis());
 
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -303,6 +303,9 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
       // are updated before we touch the "assignments" node to avoid race condition
       for (String key : datastreamMap.keySet()) {
         _store.updateDatastream(key, datastreamMap.get(key), false);
+
+        // invoke post datastream state change action for recently updated datastream
+        invokePostDSStateChangeAction(datastreamMap.get(key));
       }
       _coordinator.broadcastDatastreamUpdate();
     } catch (DatastreamException e) {
@@ -359,6 +362,8 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
         if (DatastreamStatus.READY.equals(datastream.getStatus())) {
           d.setStatus(DatastreamStatus.PAUSED);
           _store.updateDatastream(d.getName(), d, true);
+          // invoke post datastream state change action for recently paused datastream
+          invokePostDSStateChangeAction(d);
         } else {
           LOG.warn("Cannot pause datastream {}, as it is not in READY state. State: {}", d, datastream.getStatus());
         }
@@ -467,12 +472,16 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
           d.setStatus(DatastreamStatus.STOPPING);
           _store.updateDatastream(d.getName(), d, true);
           _store.deleteDatastreamNumTasks(d.getName());
+          // invoke post datastream state change action for recently stopped datastream
+          invokePostDSStateChangeAction(d);
         } else if (DatastreamStatus.STOPPING.equals(d.getStatus())) {
           // this check helps in preventing any datastream from being stuck in STOPPING state indefinitely
           LOG.warn("Datastream {} is already in {} state. Notifying leader to initiate transition", d,
               d.getStatus());
           _store.updateDatastream(d.getName(), d, true);
           _store.deleteDatastreamNumTasks(d.getName());
+          // invoke post datastream state change action for recently stopped datastream
+          invokePostDSStateChangeAction(d);
         } else {
           LOG.warn("Cannot stop datastream {}, as it is not in READY/PAUSED state. State: {}", d, d.getStatus());
         }
@@ -528,6 +537,8 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
             DatastreamStatus.STOPPED.equals(datastream.getStatus())) {
           d.setStatus(DatastreamStatus.READY);
           _store.updateDatastream(d.getName(), d, true);
+          // invoke post datastream state change action for recently resumed datastream
+          invokePostDSStateChangeAction(d);
         } else {
           LOG.warn("Will not resume datastream {}, as it is not already in PAUSED/STOPPED state", d);
         }
@@ -712,7 +723,8 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
 
   @Override
   public UpdateResponse delete(String datastreamName) {
-    if (null == _store.getDatastream(datastreamName)) {
+    final Datastream datastream = _store.getDatastream(datastreamName);
+    if (null == datastream) {
       _errorLogger.logAndThrowRestLiServiceException(HttpStatus.S_404_NOT_FOUND,
           "Datastream requested to be deleted does not exist: " + datastreamName);
     }
@@ -724,6 +736,9 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
       Instant startTime = Instant.now();
       _store.deleteDatastream(datastreamName);
       DELETE_CALL_LATENCY_MS.set(Duration.between(startTime, Instant.now()).toMillis());
+
+      // invoke post datastream state change action for recently deleted datastream
+      invokePostDSStateChangeAction(datastream);
 
       return new UpdateResponse(HttpStatus.S_200_OK);
     } catch (Exception e) {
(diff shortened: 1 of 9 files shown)


datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -1348,6 +1348,77 @@ public void testValidateDatastreamsUpdate() throws Exception {
...
+    CachedDatastreamReader datastreamCache = new CachedDatastreamReader(zkClient, testCluster);
+    ZookeeperBackedDatastreamStore dsStore = new ZookeeperBackedDatastreamStore(datastreamCache, zkClient, testCluster);
+    DatastreamResources datastreamResources = new DatastreamResources(dsStore, coordinator1);
+    datastreamResources.update(datastream.getName(), datastream);
+
+    assertConnectorReceiveDatastreamUpdate(connector1, datastream);
+    Assert.assertTrue(PollUtils.poll(() -> connector1.getPostDSStatechangeActionInvokeCount() == 2, 200,
+        WAIT_TIMEOUT_MS));
+    Assert.assertTrue(PollUtils.poll(() -> connector2.getPostDSStatechangeActionInvokeCount() == 1, 200,
+        WAIT_TIMEOUT_MS));
```
vmaheshw, line 1405: Why should connector2 also publish an event? If there are 100 hosts and the tasks are assigned to all of them, an update operation will result in 100 events.

vmaheshw (commented):
In TestCoordinator, I just checked the update sequence and stopped at seeing multiple events published for one update. I did not evaluate the other tests.

Completed review for other source code.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1030,6 +1032,9 @@ private void handleDatastreamAddOrDelete() {
             _log.warn("Failed to update datastream: {} after initializing. This datastream will not be scheduled for "
                 + "producing events ", ds.getName());
             shouldRetry = true;
+          } else {
+            // invoke post datastream upsert/delete or state modify actions
```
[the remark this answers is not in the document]
hshukla: yes, comment is general, I will modify for the current operation. Regarding the datastream being stuck on `INITIALIZING `, I thought about that too. Once the DS is stuck on that, we retry by adding the task in the queue. so when the task gets executed, and if we publish to DH or have any other downstream actions on INITIALIZING, then it will get invoked on retry, which would not be good. You may want to take followup action only once the Datastream gets into READY state, at least that was the intention behind this logic.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -438,6 +438,7 @@ public void onDatastreamUpdate() {
             datastreamGroups.stream().filter(x -> x.getTaskPrefix().equals(task.getTaskPrefix())).findFirst();
         if (dg.isPresent()) {
           ((DatastreamTaskImpl) task).setDatastreams(dg.get().getDatastreams());
+          dg.get().getDatastreams().forEach(this::invokePostDataStreamStateChangeAction);
```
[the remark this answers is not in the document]
hshukla: yes, you are right. For the update, we do not have dedicated listener that runs only on a leader. I am thinking to make the postDS create, update, delete or state change not specific for DH publish. Like, if we want to allow any further followup actions, which requires task level knowledge, keeping logic here would be helpful and since we are running the post invocation logic as an async call, I thought it would be ok.

shrinandthakkar (commented):
Thinking out loud for a couple of cases, 

When we delete a datastream,
>  There would be two threads, 
> 1. The thread that handles the datastream resource request, updates the zookeeper and notifies the leader of the change. 
> 2. The second thread invokes postDatastreamStateChangeAction. 
> 
> Now, the leader would create a new assignment and invoke the respective followers to update their respective tasks. 
> 
> So, wouldn't there be chances of race conditions between deleting those tasks (for a delete request) and fetching the source information from the tasks for the postDatastreamStateChangeAction thread? 
> 

When we create a new datastream,
>  Similar race condition?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1089,6 +1090,24 @@ private void hardDeleteDatastream(Datastream ds, List<Datastream> activeStreams)
...
+   * Invokes post datastream state change action of connector for given datastream.
+   * @param datastream the datastream
+   * @throws DatastreamException if fails to perform post datastream action
+   */
+  public void invokePostDataStreamStateChangeAction(final Datastream datastream) throws DatastreamException {
+    _log.info("Invoke post datastream state change action");
+    try {
+      final String connectorName = datastream.getConnectorName();
+      final ConnectorInfo connectorInfo = _connectors.get(connectorName);
+      connectorInfo.getConnector().postDatastreamStateChangeAction(datastream);
```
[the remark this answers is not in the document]
hshukla: My thinking process was same when I started implementing this logic to have the async behavior here in coordinator so that connectors do not need to worry about it. However, if you think and compare it with other existing methods from connector interface, ideally it would be each connectors responsibility to implement the way it wants. For our use case we are ok with fire and forget call, what if some one else needs a blocking call, similar to validateUpdate or postDatastreamInitialization? This made be push the async logic inside the connector and not in the upper level(coordinator). What do you think?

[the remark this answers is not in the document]
shrinandthakkar: Although I understand that it would be difficult to manage the life cycle of the thread from the coordinator, I was only concerned as this new method is being called while handling a client request, and I think we shouldn't ideally be supporting a blocking call outside of this repo for that.

Can we somehow enforce a max timeout for these invocations in this implementation, so that we don't block the corresponding request due to unknown implementations outside of this repo?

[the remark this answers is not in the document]
hshukla: You are right, this would introduce yet extra operation during REST call, however, I don't think metadata publish usecase should determine the blocking vs non-blocking call for open source use case.

> Can we somehow enforce a max timeout for these invocations in this implementation, so that we don't block the corresponding request due to unknown implementations outside of this repo?

I don't think there would be a way unless we use a separate thread and make it a blocking call with timeout(1 or 2sec). Right now, the client request timeout will kick it, which I am not sure what it is (Ideally 5sec or default 30secs). I still feel that would be inside connectors as they will be interested in the result of this call and not the coordinator.  The new method java doc explicitly calls out the behavior of this method call and if you are worried about API call latency, make sure it gets addressed in connector implementation.

I don't think it would be difficult to manage the life cycle of thread. With this design I see this as yet another helper method for connectors to do their work. If they want it to be blocking call, they can do so. For us, right now we are talking about async call for metadata publish, but that would not be a case for other connectors or other use case. For metadata publish, we can make an async call in the respective connectors, if additional functionality is needed in the future which requires blocking call, this approach will let us do that.

I am not strongly opposing this idea, but design wise it make more sense to have this within connector.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -467,12 +472,16 @@ public ActionResult<Void> stop(@PathKeysParam PathKeys pathKeys,
...
+          // invoke post datastream state change action for recently stopped datastream
+          invokePostDSStateChangeAction(d);
         } else if (DatastreamStatus.STOPPING.equals(d.getStatus())) {
           // this check helps in preventing any datastream from being stuck in STOPPING state indefinitely
           LOG.warn("Datastream {} is already in {} state. Notifying leader to initiate transition", d,
               d.getStatus());
           _store.updateDatastream(d.getName(), d, true);
           _store.deleteDatastreamNumTasks(d.getName());
+          // invoke post datastream state change action for recently stopped datastream
+          invokePostDSStateChangeAction(d);
```
shrinandthakkar, line 484: The two invocations here for state change action are after setting the datastream status to STOPPING state, and we wait (in the polling loop below) in this stop rest call for the datastream to change to the STOPPED state before it is returned or, in the worst case, times out. 

So, instead of calling the invocation for the STOPPING state, should we call it after the poll loop so that the correct data stream status STOPPED is reflected? 

Also, for a much more consistent state change tracking, considering this invocation is lightweight, along with invoking these two invocations for the STOPPING state, we should have another invocation for the STOPPED state after the polling loop.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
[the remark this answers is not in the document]
shrinandthakkar: Synced up with @hshukla on this, as long as we somehow enforce that these are implemented asynchronously, within their connector implementation, we should be good. 

And in some case when some implementations are made blocking as such that they timeout the request, it won't cause much of any harm since the state change would have been performed as expected unless in cases when multiple rest calls are invoked (in a combined fashion, which is rare).

shrinandthakkar:
> Thinking out loud for a couple of cases,
> 
> When we delete a datastream,
> 
> > There would be two threads,
> > 
> > 1. The thread that handles the datastream resource request, updates the zookeeper and notifies the leader of the change.
> > 2. The second thread invokes postDatastreamStateChangeAction.
> > 
> > Now, the leader would create a new assignment and invoke the respective followers to update their respective tasks.
> > So, wouldn't there be chances of race conditions between deleting those tasks (for a delete request) and fetching the source information from the tasks for the postDatastreamStateChangeAction thread?
> 
> When we create a new datastream,
> 
> > Similar race condition?

Synced up with @hshukla on this, and we decided that we'd try to get all the required info for these cases from the connector level and not the task level to avoid these race conditions.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
hshukla: Valid points

> So, instead of calling the invocation for the STOPPING state, should we call it after the poll loop so that the correct data stream status STOPPED is reflected?

The reason I added here and not after the polling, current logic written such a way that if one of the update fails it will throw the exception and return 5xx to client. So, the polling logic will not be invoked and we will not be able to update to DH. For example, if there are 5 datastreams need to be stopped, and if it causes failure at 3rd, we will throw the exception, so even though first 2 got stopped successfully, we will not have it reflected into DH.

>Also, for a much more consistent state change tracking, considering this invocation is lightweight, along with invoking these two invocations for the STOPPING state, we should have another invocation for the STOPPED state after the polling loop.

I like this idea about invoking this again. The connector can then decide what to do based on the datastream status. But, by the time connector gets the call from the 1st invocation, the datastream may have moved from stopping to stopped, so it may make it redundant, what do you think?

shrinandthakkar: Yes, we better update it in both states, Stopping and Stopped state. And if the connector is implementing it in separate threads, then there will be multiple updates to any external system where anyone might need to send it, which is what we want, I think.

That way, all the incremental updates to the DS are tracked properly.

hshukla: Ok, it would still not be tracked properly as we fail out the request if anyone of the update gives error. Also, when we perform action in the new method based on stopping call, the DS could be in stopped state already. However, I don't see a way to make it 100% accurate so by providing ability to call it for each action, STOPPING and STOPPED at least connector has a chance to act on it.

shrinandthakkar: Yep but you can send a new object to the connectors instead of sending the reference, so it could be tracked properly.

hshukla: Just for reference: what I mean, when we call postDSState...() on STOPPING, by the time the connector run the logic, it could be in the STOPPED state, however, the object that we have passed in has STOPPING. Not a big issue as we will anyway call postDSState...() again with status STOPPED.

shrinandthakkar (approved):
LGTM

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw, line 1103: don't pass by reference, as the object may get modified in the connector. We have seen a bug in the past where the connector messed with the datastream cache.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1089,6 +1090,24 @@ private void hardDeleteDatastream(Datastream ds, List<Datastream> activeStreams)
...
+    } catch (DatastreamException e) {
+      _log.error("Failed to perform post datastream state change action datastream={}", datastream);
+      _metrics.updateKeyedMeter(CoordinatorMetrics.KeyedMeter.VALIDATE_DATASTREAMS_UPDATE_NUM_ERRORS, 1);
```
vmaheshw, line 1106: What is this metric for? Is the name correct or is this a copy-paste error?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1089,6 +1090,24 @@ private void hardDeleteDatastream(Datastream ds, List<Datastream> activeStreams)
...
+   * @throws DatastreamException if fails to perform post datastream action
+   */
+  public void invokePostDataStreamStateChangeAction(final Datastream datastream) throws DatastreamException {
+    _log.info("Invoke post datastream state change action");
+    try {
+      final String connectorName = datastream.getConnectorName();
+      final ConnectorInfo connectorInfo = _connectors.get(connectorName);
+      connectorInfo.getConnector().postDatastreamStateChangeAction(datastream);
+    } catch (DatastreamException e) {
+      _log.error("Failed to perform post datastream state change action datastream={}", datastream);
```
vmaheshw, line 1105: Just print the datastream name. In case of MM, the datastream objects are really huge and make the logs noisy. Please check all the logs in the patch and just print the name unless explicitly required.

hshukla: Ok, we print the whole object other places so I was not whether that was mandatory for open source project. refactored.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
hshukla: Good catch, forgot to update with the new constant.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: Thanks. We tried to fix as many places as we could. This one clearly can work with just the datastream name.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -1155,6 +1155,46 @@ public void testDatastreamWithBYOT() throws Exception {
...
+    Coordinator coordinator1 = createCoordinator(_zkConnectionString, testCluster);
+    coordinator1.addConnector(connectorType, connector1, new BroadcastStrategy(Optional.empty()),
+        false, new SourceBasedDeduper(), null);
+    coordinator1.start();
+    ZkClient zkClient = new ZkClient(_zkConnectionString);
+
+    // create datastream
+    Datastream[] list = DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, connectorType,
+        "datastream1");
+    Datastream datastream1 = list[0];
```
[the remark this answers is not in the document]
hshukla: With prev implementation this should have mattered but now we run this logic as part of the REST call, so in this method, this would be based on per datastream. The reason I added the test is that we have new public method. I have added it, you can see when we create datastreamResource we mention what coordinator to use.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -722,9 +742,13 @@ public UpdateResponse delete(String datastreamName) {
 
       _dynamicMetricsManager.createOrUpdateMeter(CLASS_NAME, DELETE_CALL, 1);
       Instant startTime = Instant.now();
+      datastream.setStatus(DatastreamStatus.DELETING);
```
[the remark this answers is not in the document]
hshukla: The status will be useful in connector to determine what kind of Metadata Platform event we need to generate. Yes, we have it set inside ZKDatastreamStore, but do we want to rely on that logic to have state change for the object? what if we make a copy of that object inside ZKDatastreamStore and then update the status, here we will not have status reflected?

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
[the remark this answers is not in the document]
vmaheshw: It was a very good testing scenario that verifies that  "Update on one datastream does not trigger event for another datastream".

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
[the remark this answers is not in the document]
vmaheshw: This does not make sense that ZKDatastreamStore will modify the state on another object and not the main object.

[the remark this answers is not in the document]
hshukla: yes, I went with theory that we should not rely on international implementation after seeing its split in two classes. However, you are right, it's ideally datastore responsibility. reverted.

vmaheshw (approved):
LGTM

shrinandthakkar (approved):

PR #919, opened 2022-12-07
BMM Restart Improvements Part 1. Leader Coordinator Issuing Assignment Tokens

jzakaryan:
This pull request is part of a series of changes that are meant to improve BMM Restart and make it easier to debug restart failures and trace them to the faulty hosts. It introduces the concept of assignment tokens and adds support for issuing those tokens by the leader coordinator. This feature can be toggled by the newly introduced `brooklin.server.coordinator.enableAssignmentTokens` configuration.

The subsequent PRs will deal with the following aspects:
Part 2 – Changes to the followers' `onAssignmentChange` to make them claim the tokens issued by the leader.
Part 3 – Changes to the leader to make it poll the ZooKeeper and wait for the assignment change (stop) to be propagated and executed by the cluster. This will also include code for handling assignment failures and signals to the client that problem(s) occurred in the form of log messages and a new metric.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentToken.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentToken.java
@@ -12,7 +12,7 @@ import com.linkedin.datastream.common.JsonUtils;
  * Data structure to store assignment tokens. These are used as a mechanism for followers to signal the leader that
  * they handled assignment change
  */
-public class AssignmentToken {
+public final class AssignmentToken {
   private String _issuedBy;
   private String _issuedFor;
   private long  _timestamp;
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentToken.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentToken.java
@@ -1,0 +1,91 @@
+/**
+ *  Copyright 2022 LinkedIn Corporation. All rights reserved.
+ *  Licensed under the BSD 2-Clause License. See the LICENSE file in the project root for license information.
+ *  See the NOTICE file in the project root for additional information regarding copyright ownership.
+ */
+package com.linkedin.datastream.server;
+
+import com.linkedin.datastream.common.JsonUtils;
+
+
+/**
+ * Data structure to store assignment tokens. These are used as a mechanism for followers to signal the leader that
+ * they handled assignment change
+ */
+public class AssignmentToken {
+  private String _issuedBy;
+  private String _issuedFor;
+  private long  _timestamp;
+
+  /**
+   * Constructor for {@link AssignmentToken}
+   */
+  public AssignmentToken(String issuedBy, String issuedFor) {
+    _issuedBy = issuedBy;
+    _issuedFor = issuedFor;
+    _timestamp = System.currentTimeMillis();
+  }
+
+  /**
+   * Default constructor for {@link AssignmentToken}, required for json ser/de
+   */
+  public AssignmentToken() {
+
+  }
+
+  /**
+   * Creates {@link AssignmentToken} instance from JSON
+   */
+  public static AssignmentToken fromJson(String json) {
+    return JsonUtils.fromJson(json, AssignmentToken.class);
+  }
+
+  /**
+   * Converts the object to JSON
+   */
+  public String toJson() {
+    return JsonUtils.toJson(this);
+  }
+
+  /**
+   * Gets the name of the leader host that issued the token
+   */
+  public String getIssuedBy() {
+    return _issuedBy;
+  }
+
+  /**
+   * Gets the name of the host for which the token was issued
+   */
+  public String getIssuedFor() {
+    return _issuedFor;
+  }
+
+  /**
+   * Gets the timestamp (in UNIX epoch format) for when the token was issued
+   */
+  public long getTimestamp() {
+    return _timestamp;
+  }
+
+  /**
+   * Sets the name of the leader host that issued the token
+   */
+  public void setIssuedBy(String issuedBy) {
+    _issuedBy = issuedBy;
+  }
+
+  /**
+   * Sets the name of the host for which the token was issued
+   */
+  public void setIssuedFor(String issuedFor) {
+    _issuedFor = issuedFor;
+  }
+
+  /**
+   * Sets the timestamp (in UNIX epoch format) for when the token was issued
+   */
+  public void setTimestamp(long timestamp) {
+    _timestamp = timestamp;
+  }
+}
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/zk/KeyBuilder.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/zk/KeyBuilder.java
@@ -22,6 +22,8 @@ public final class KeyBuilder {
   private static final String INSTANCE_ASSIGNMENT = "/%s/instances/%s/assignments/%s";
   private static final String DATASTREAMS = "/%s/dms";
   private static final String DATASTREAM = "/%s/dms/%s";
+  private static final String DATASTREAM_ASSIGNMENT_TOKENS = "/%s/dms/%s/assignmentTokens";
+  private static final String DATASTREAM_ASSIGNMENT_TOKEN_FOR_INSTANCE = "/%s/dms/%s/assignmentTokens/%s";
   private static final String CONNECTORS = "/%s/connectors";
   private static final String CONNECTOR = "/%s/connectors/%s";
 
@@ -198,6 +200,25 @@ public final class KeyBuilder {
    */
   public static String datastream(String cluster, String stream) {
     return String.format(DATASTREAM, cluster, stream);
+  }
+
+  /**
+   * Get the ZooKeeper znode for a datastream's assignment tokens in a Brooklin cluster
+   * @param cluster Brooklin cluster name
+   * @param stream Datastream name
+   */
+  public static String datastreamAssignmentTokens(String cluster, String stream) {
+    return String.format(DATASTREAM_ASSIGNMENT_TOKENS, cluster, stream);
+  }
+
+  /**
+   * Get the ZooKeeper znode for a datastream's assignment token for an instance in a Brooklin cluster
+   * @param cluster Brooklin cluster name
+   * @param stream Datastream name
+   * @param instance Instance name
+   */
+  public static String datastreamAssignmentTokenForInstance(String cluster, String stream, String instance) {
+    return String.format(DATASTREAM_ASSIGNMENT_TOKEN_FOR_INSTANCE, cluster, stream, instance);
   }
 
   /**
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
@@ -752,6 +752,7 @@ public class ZkAdapter {
         collect(Collectors.toMap(DatastreamGroup::getTaskPrefix, Function.identity()));
 
     // For each stopping datastream group, find all the instances that currently have tasks assigned
+    Map<DatastreamGroup, Set<String>> stoppingDgInstances = new HashMap<>();
     currentAssignment.keySet()
         .forEach(i -> currentAssignment.get(i).stream()
             .filter(t -> stoppingDatastreamTaskPrefixes.contains(t.getTaskPrefix()))
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
@@ -29,6 +29,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.Validate;
 import org.apache.helix.zookeeper.zkclient.IZkChildListener;
 import org.apache.helix.zookeeper.zkclient.IZkDataListener;
 import org.apache.helix.zookeeper.zkclient.IZkStateListener;
@@ -741,6 +742,9 @@ public class ZkAdapter {
    */
   public void updateAllAssignmentsAndIssueTokens(Map<String, List<DatastreamTask>> assignmentsByInstance,
       List<DatastreamGroup> stoppingDatastreamGroups) {
+    Validate.notNull(assignmentsByInstance);
+    Validate.notNull(stoppingDatastreamGroups);
+
     Map<String, Set<DatastreamTask>> currentAssignment = getAllAssignedDatastreamTasks();
     Set<String> stoppingDatastreamTaskPrefixes = stoppingDatastreamGroups.stream().
         map(DatastreamGroup::getTaskPrefix).collect(toSet());
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
@@ -5,6 +5,8 @@
  */
 package com.linkedin.datastream.server.zk;
 
+import com.linkedin.datastream.server.AssignmentToken;
+import com.linkedin.datastream.server.DatastreamGroup;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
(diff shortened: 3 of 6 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+   * <ol>
+   *    *  <li>{@code /<cluster>/dms/<datastream-name>/assignmentTokens/<token1>,<token2>...}</li>
+   * </ol>
+   */
+  public void updateAllAssignmentsAndIssueTokens(Map<String, List<DatastreamTask>> assignmentsByInstance,
+      List<DatastreamGroup> stoppingDatastreamGroups) {
+    Validate.notNull(assignmentsByInstance);
+    Validate.notNull(stoppingDatastreamGroups);
+
+    Map<String, Set<DatastreamTask>> currentAssignment = getAllAssignedDatastreamTasks();
```
shrinandthakkar, line 748: The java doc of `getAllAssignedDatastreamTasks` says that there may be uncleaned dead instance – task assignments left which will be cleaned up later in the `handleLeaderDoAssignment` routine (line #1245).

So, if the leader issues tokens to those dead hosts, those would never be picked up?

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentToken.java
```diff
@@ -0,0 +1,91 @@
...
+
+
+/**
+ * Data structure to store assignment tokens. These are used as a mechanism for followers to signal the leader that
+ * they handled assignment change
+ */
+public class AssignmentToken {
+  private String _issuedBy;
+  private String _issuedFor;
+  private long  _timestamp;
```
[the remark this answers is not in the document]
jzakaryan: These can't be marked final. Getters/setters are required for JSON ser/de

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentToken.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Gets the timestamp (in UNIX epoch format) for when the token was issued
+   */
+  public long getTimestamp() {
+    return _timestamp;
+  }
+
+  /**
+   * Sets the name of the leader host that issued the token
```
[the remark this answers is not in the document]
jzakaryan: I left them out on purpose. I saw that javadocs for return type and parameters are omitted in similar simple object types (see `HostTargetAssignment` for reference)

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+          stoppingDgInstances.computeIfAbsent(datastreamGroup, k -> new HashSet<>()).add(instance);
+        }
+      }
+
+    String hostname = "localhost";
+    try {
+      hostname = InetAddress.getLocalHost().getHostName();
+    } catch (UnknownHostException ex) {
+      LOG.warn("Unable to obtain hostname for leader");
+    }
```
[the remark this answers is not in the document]
jzakaryan: This same assumption is made in other parts of the open source brooklin. If we're planning on addressing this, it needs to be done separately and in all places.

Example classes that make use of `InetAddress.getLocalHost().getHostName()`
`ServerComponentHealthAggregator`
`AbstractKafkaConnector`

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+   *
+   * In addition to the work done by updateAllAssignments, this method writes to
+   * <ol>
+   *    *  <li>{@code /<cluster>/dms/<datastream-name>/assignmentTokens/<token1>,<token2>...}</li>
+   * </ol>
+   */
+  public void updateAllAssignmentsAndIssueTokens(Map<String, List<DatastreamTask>> assignmentsByInstance,
+      List<DatastreamGroup> stoppingDatastreamGroups) {
+    Validate.notNull(assignmentsByInstance);
+    Validate.notNull(stoppingDatastreamGroups);
```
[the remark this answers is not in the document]
ehoner: I am not sure that these checks are providing the "correct" level of "validation" here. `Validate.nonNull` for `List<>` and `Map<>` types validates the objects are not `null`. It does not operate on the elements. 

Are there conditions where either the keys can be empty and/or the objects `null` and the operation can continue? Why would this be potentially called with `null`, ie. is this a pro-forma check?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+        String assignmentTokensPath = KeyBuilder.datastreamAssignmentTokens(_cluster, stoppingStream.getName());
+        _zkclient.ensurePath(assignmentTokensPath);
+
+        Set<String> instances = stoppingDgInstances.get(stoppingGroup);
+        LOG.info("Issuing assignment tokens for stream {} and {} instance(s)", stoppingStream.getName(), instances.size());
+        for (String instance : instances) {
+          String assignmentTokenPath = KeyBuilder.datastreamAssignmentTokenForInstance(_cluster,
+              stoppingStream.getName(), instance);
+          AssignmentToken token = new AssignmentToken(hostname, instance);
+          _zkclient.create(assignmentTokenPath, token.toJson(), CreateMode.PERSISTENT);
```
shrinandthakkar, line 793: could something bad happen here if two concurrent stop requests occur? Like should we have a check routine if an assignment token already exists for a host, then we don't create it?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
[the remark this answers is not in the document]
ehoner: If there isn't already an abstraction that identifies each instance in a cluster uniquely, we can leave this and add an identifier abstraction in separate work.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+
+    Map<String, Set<DatastreamTask>> currentAssignment = getAllAssignedDatastreamTasks();
+    Set<String> stoppingDatastreamTaskPrefixes = stoppingDatastreamGroups.stream().
+        map(DatastreamGroup::getTaskPrefix).collect(toSet());
+    Map<String, DatastreamGroup> taskPrefixDatastreamGroups = stoppingDatastreamGroups.stream().
+        collect(Collectors.toMap(DatastreamGroup::getTaskPrefix, Function.identity()));
+
+    // For each stopping datastream group, find all the instances that currently have tasks assigned
+    Map<DatastreamGroup, Set<String>> stoppingDgInstances = new HashMap<>();
+    for (String instance : currentAssignment.keySet()) {
```
hshukla, line 756: I think you can replace this for loop with following block, this way you do not need to create intermediate Set for instanceTasks and List for stoppingTasks and iterate on that list again.
```suggestion
    currentAssignment.keySet()
        .forEach(i -> currentAssignment.get(i).stream()
            .filter(t -> stoppingDatastreamTaskPrefixes.contains(t.getTaskPrefix()))
            .forEach(st -> {
              DatastreamGroup datastreamGroup = taskPrefixDatastreamGroups.get(st.getTaskPrefix());
              stoppingDgInstances.computeIfAbsent(datastreamGroup, k -> new HashSet<>()).add(i);
            }));
```

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+      LOG.warn("Unable to obtain hostname for leader");
+    }
+
+    // Issue assignment tokens for each instance and each stopping datastream
+    for (DatastreamGroup stoppingGroup : stoppingDatastreamGroups) {
+      for (Datastream stoppingStream : stoppingGroup.getDatastreams()) {
+        String path = KeyBuilder.datastream(_cluster, stoppingStream.getName());
+
+        if (!_zkclient.exists(path)) {
+          LOG.warn("Trying to issue assignment tokens for non-existing stream: " + stoppingStream.getName());
```
hshukla, line 780:
```suggestion
          LOG.warn("Trying to issue assignment tokens for non-existing stream={}", stoppingStream.getName());
```

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
hshukla: looks like `_zkclient.create` can throw exception, in that case are we okay with the partial assignmentToken written? Does leader know that there needs to be N assignmentToken assigned and there is only X, so even though those X have said that I have finished my assignments, leaders will not consider this as a success?

datastream-server/src/test/java/com/linkedin/datastream/server/zk/TestZkAdapter.java
```diff
@@ -1059,6 +1061,59 @@ public void testUpdateAssignmentDeleteUnusedTasks() {
     adapter.disconnect();
   }
 
+  @Test
+  public void testUpdateAllAssignmentAndIssueTokens() throws Exception {
```
hshukla, line 1065: Can we add a test for failure cases too? Like in your code https://github.com/linkedin/brooklin/pull/919/files#diff-c0704e6fe31835d24d153416a6d6efb590a4458c334d83326cd61e64c6ce9e8eR759 you have filter for stopping DS tasks only. Can we add test where we have mix of task(stopping and non-stopping) and make sure the code takes only stopping tasks?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+    for (String instance : currentAssignment.keySet()) {
+        Set<DatastreamTask> instanceTasks = currentAssignment.get(instance);
+        List<DatastreamTask> stoppingTasks = instanceTasks.stream().
+            filter(t -> stoppingDatastreamTaskPrefixes.contains(t.getTaskPrefix())).
+            collect(Collectors.toList());
+        for (DatastreamTask stoppingTask : stoppingTasks) {
+          DatastreamGroup datastreamGroup = taskPrefixDatastreamGroups.get(stoppingTask.getTaskPrefix());
+          stoppingDgInstances.computeIfAbsent(datastreamGroup, k -> new HashSet<>()).add(instance);
+        }
+      }
```
hshukla, line 765: I think you can replace this for loop with following block, this way you do not need to create intermediate Set for instanceTasks and List for stoppingTasks and iterate on that list again.
```suggestion
    currentAssignment.keySet()
        .forEach(i -> currentAssignment.get(i).stream()
            .filter(t -> stoppingDatastreamTaskPrefixes.contains(t.getTaskPrefix()))
            .forEach(st -> {
              DatastreamGroup datastreamGroup = taskPrefixDatastreamGroups.get(st.getTaskPrefix());
              stoppingDgInstances.computeIfAbsent(datastreamGroup, k -> new HashSet<>()).add(i);
            }));
```

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: @shrinandthakkar assignment token path is created under the datasream. Your previous work with datastream states already guarantees that no two concurrent stop requests on a datastream can happen at any given time in the cluster (state transitions are rejected if datastream is in STOPPING state).

@hshukla Zookeeper writes are atomic - there's no way we can end up with a partially written token. If the concern is more in the lines of "what if Zookeeper fails and leader writes subset of the tokens", then it's a cluster problem. One such example is session expiry. It is a critical failure and a lot of work went into handling it. No need for extra retry loops or checks here imo.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: Even if they never get picked up, leader will end up revoking them. I'll see if it's possible to distinguish these dead instances and exclude them from the error log message and metrics (no need to alert for dead instances). Those changes are in the next PR.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: Your version is not necessarily better than string concatenation in mine. Concatenation is usually faster than formatting a string too. I prefer formatting when there are more variables involved and when you need finer control wrt how those variables are formatted in the final string.

Interesting SO post that I found:
https://stackoverflow.com/questions/925423/is-it-better-practice-to-use-string-format-over-string-concatenation-in-java

datastream-server/src/main/java/com/linkedin/datastream/server/zk/KeyBuilder.java
```diff
@@ -22,6 +22,8 @@ public final class KeyBuilder {
   private static final String INSTANCE_ASSIGNMENT = "/%s/instances/%s/assignments/%s";
   private static final String DATASTREAMS = "/%s/dms";
   private static final String DATASTREAM = "/%s/dms/%s";
+  private static final String DATASTREAM_ASSIGNMENT_TOKENS = "/%s/dms/%s/assignmentTokens";
```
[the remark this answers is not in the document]
jzakaryan: I like the suggestion, but this is a nit and I don't want my 2 string constants to stand out in this file. Maybe that can be done in the whole file as a separate change.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
hshukla: I don't think the performance would be a problem here, I would go with consistency and maintain the logging format.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/KeyBuilder.java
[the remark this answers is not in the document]
hshukla: Ok. We might never come back. I would not think that would be a big deal, but again, not a blocker.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: We have mixed uses for logging API calls everywhere. And again, this is more of a personal preference.

datastream-server/src/test/java/com/linkedin/datastream/server/zk/TestZkAdapter.java
jzakaryan: That's what my current test is doing. It's creating 2 datastreams and only one of them is stopping. Please refer to the assertions I have there.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -728,6 +731,73 @@ private void removeTaskNode(String instance, String name) {
...
+          String assignmentTokenPath = KeyBuilder.datastreamAssignmentTokenForInstance(_cluster,
+              stoppingStream.getName(), instance);
+          AssignmentToken token = new AssignmentToken(hostname, instance);
+          _zkclient.create(assignmentTokenPath, token.toJson(), CreateMode.PERSISTENT);
+        }
+      }
+    }
+
+    updateAllAssignments(assignmentsByInstance);
+  }
```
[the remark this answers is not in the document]
jzakaryan: Done

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
hshukla: > "what if Zookeeper fails and leader writes subset of the tokens",
Yes, I was referring this.  There could be a transient failures too does not mean the ZK is down. If there are 5 tokens to write, and it fails after 3, what happens here? will the request fail or if not, then  when leader polls the result does it consider that as a successful restart?

shrinandthakkar: @jzakaryan, as far as I understand, even after my work on ensuring that the leader would create an assignment when there's a stop request, there could still be a chance that two concurrent stop or restart requests occur. Still, since the reassignment is the same, the other request would be considered a no-op. 

So, my question here is if two concurrent stop requests occur, is there something that might create issues for the code in `updateAllAssignmentsAndIssueTokens` before it calls `updateAllAssignments`? 
1. What if a host claimed a token and a concurrent stop request created a new token for that same host until that host finished the request?  
2. ZK-related issues could also occur, but those would be cluster-level issues, which could not be handled here, I understand.

ehoner (approved):
I can approve again if you want to change the json serialization.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: @shrinandthakkar just to make sure I understand your concern, are you claiming that 2 concurrent assignment requests for the **same datastream** can occur?

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentToken.java
```diff
@@ -0,0 +1,91 @@
...
+  public void setIssuedFor(String issuedFor) {
+    _issuedFor = issuedFor;
+  }
+
+  /**
+   * Sets the timestamp (in UNIX epoch format) for when the token was issued
+   */
+  public void setTimestamp(long timestamp) {
+    _timestamp = timestamp;
+  }
```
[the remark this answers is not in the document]
jzakaryan: Thanks for sharing the API usage example. With the next set of changes I'll change my JSON class to make use of these and keep the class otherwise immutable.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: @hshukla Zookeeper client will throw a runtime exception and it will be caught by the `Coordinator` which will schedule a retry.
https://github.com/linkedin/brooklin/blob/71a9dea076531981cde7be52693cd98be1e34318/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java#L1242

The only potential problem here is that the consecutive attempts at assignment change can result in some hosts ending up with more than one token. I'll make sure to handle that in Part 4 PR (cleanup and edge cases).

shrinandthakkar: That is exactly my concern: we need to have a no-op behavior implemented in the hosts handling these tokens if they are already working on stopping the producer for datastreams. If you are handling that in the upcoming PR, I am good with this.

jzakaryan: @shrinandthakkar let's sync up offline on this to make sure I understand everything, and I'll handle this in the edge cases PR

shrinandthakkar (approved):

PR #921, opened 2022-12-17
BMM Restart Improvements Part 2. Followers claim their tokens after handling stop assignments

jzakaryan:
This pull request is part of a series of changes that are meant to improve BMM Restart and make it easier to debug restart failures and trace them to the faulty hosts. Part 2 includes changes to the follower's code to make them claim assignment tokens for the datastreams that have been successfully stopped. It also includes the following improvements that were suggested in #919:

- `AssignmentToken` class was made immutable thanks to @ehoner pointing out the JSON library APIs.
- Tokens now carry instance names of issuing host instead of the host name. There's a slight semantic distinction between a host name and a live instance name and the latter is more useful. This was a suggestion by @vmaheshw.

Part 1 – Introduction of assignment tokens and support for issuing tokens by the leader coordinator (#919)
**Part 2 – Changes to the followers' handleAssignmentChange to make them claim the tokens issued by the leader.**
Part 3 – Changes to the leader to make it poll the ZooKeeper and wait for the assignment change (stop) to be propagated and executed by the cluster. This will also include code for handling assignment failures and signals to the client that problem(s) occurred in the form of log messages and a new metric.


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -710,6 +710,12 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
       retryHandleAssignmentChange(isDatastreamUpdate);
     }
 
+    if (_config.getEnableAssignmentTokens()) {
+      List<DatastreamGroup> stoppingDatastreamGroups =
+          fetchDatastreamGroupsWithStatus(Collections.singletonList(DatastreamStatus.STOPPING));
```
vmaheshw, line 715: I don't think this API will work for the follower. The followers do not listen to the datastream change list. So, it does not maintain the latest snapshot of all the datastreams.  

The other challenge is: if you will try to fetch it, it is going to be an extremely expensive operation and will put pressure on zk server.

vmaheshw (commented):
Brooklin does not maintain the latest datastream cache and does not listen to incremental updates. So, this PR may not work. Please recheck if your tests contains multiple instances and the follower only instance is able to claim the token.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: I'm not sure how else to distinguish stopping streams other than having the followers fetch the datastream status from ZK. Let me see if there are any workarounds.

vmaheshw: Consider setting listener and notification for followers.

jzakaryan: That will still be an expensive operation. What I did instead is I wrote code that figures out the stopping streams from the diff of the old vs new assignment. The idea is that if all the tasks belonging to a stream are removed, then most likely the stream is stopped. Follower coordinator will try to claim the token for such streams.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -767,6 +769,71 @@ private Future<Boolean> dispatchAssignmentChangeIfNeeded(String connectorType, L
     return null;
   }
 
+  private List<DatastreamTask> getRemovedTasks(List<DatastreamTask> newAssignment, List<DatastreamTask> oldAssignment) {
+    List<DatastreamTask> removedTasks = new ArrayList<>(oldAssignment);
+    removedTasks.removeAll(newAssignment);
```
vmaheshw, line 774: Isn't this an expensive operation? Since the code does similar searches multiple times, should you convert it to set?

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -798,24 +797,31 @@ private void issueAssignmentTokensForStoppingDatastreams(List<DatastreamGroup> s
...
-    String hostname = "localhost";
-    try {
-      hostname = InetAddress.getLocalHost().getHostName();
-    } catch (UnknownHostException ex) {
-      LOG.warn("Unable to obtain hostname for leader");
+  public void claimAssignmentTokensForDatastreams(List<Datastream> datastreams, String instance) {
+    for (Datastream stream : datastreams) {
+      String streamName = stream.getName();
+      String tokenPath = KeyBuilder.datastreamAssignmentTokenForInstance(_cluster, streamName, instance);
+      if (_zkclient.exists(tokenPath)) {
```
vmaheshw, line 816: You need the `else` path. Also add more logging around this.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -767,6 +769,71 @@ private Future<Boolean> dispatchAssignmentChangeIfNeeded(String connectorType, L
...
+  }
+
+  private List<DatastreamTask> getAddedTasks(List<DatastreamTask> newAssignment, List<DatastreamTask> oldAssignment) {
+    List<DatastreamTask> addedTasks = new ArrayList<>(newAssignment);
+    addedTasks.removeAll(oldAssignment);
+    return addedTasks;
+  }
+
+  @VisibleForTesting
+  void maybeClaimAssignmentTokensForStoppingStreams(List<DatastreamTask> newAssignment,
```
vmaheshw, line 785: Add logs in both info and debug mode for debugging. If there is any issue in production in this code path, we will not be able to identify it easily.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -767,6 +769,71 @@ private Future<Boolean> dispatchAssignmentChangeIfNeeded(String connectorType, L
...
+    // stopping stream, the attempt to claim the token will be a no-op.
+    for (String connector : allConnectors) {
+      List<DatastreamTask> oldTasks = oldAssignmentPerConnector.getOrDefault(connector, Collections.emptyList());
+      List<DatastreamTask> newTasks = newAssignmentPerConnector.getOrDefault(connector, Collections.emptyList());
+      List<DatastreamTask> removedTasks = getRemovedTasks(newTasks, oldTasks);
+
+      List<Datastream> stoppingStreams = inferStoppingDatastreamsFromAssignment(newTasks, removedTasks);
+      List<String> stoppingStreamNames = stoppingStreams.stream().map(Datastream::getName).collect(Collectors.toList());
+
+      if (!stoppingStreamNames.isEmpty()) {
```
vmaheshw, line 815: You probably need `else` from the logging/debugging aspect.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -767,6 +769,71 @@ private Future<Boolean> dispatchAssignmentChangeIfNeeded(String connectorType, L
...
+        _log.info("Trying to claim assignment tokens for streams: {}", stoppingStreamNames);
+        _adapter.claimAssignmentTokensForDatastreams(stoppingStreams, _adapter.getInstanceName());
+      }
+    }
+  }
+
+  @VisibleForTesting
+  static List<Datastream> inferStoppingDatastreamsFromAssignment(List<DatastreamTask> newAssignment,
```
vmaheshw, line 823: Need logging here as well.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw, line 815: Add metrics that help to understand how many inferences were correct?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -710,6 +712,10 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
       retryHandleAssignmentChange(isDatastreamUpdate);
     }
 
+    if (_config.getEnableAssignmentTokens()) {
+      maybeClaimAssignmentTokensForStoppingStreams(newAssignment, oldAssignment);
```
vmaheshw, line 716: This does not look right. The assignment token must be claimed after all the assigned tasks are stopped. Claiming after submission without knowing that the task has stopped will have a rippling effect. The leader will mark the datastream "STOPPED" and attempt to "RESUME" immediately. With low level, it will result in the same partition getting processed by two tasks, as the task dependency is not maintained outside the restart boundry.

shrinandthakkar: Vaibhav, trying to understand your comment here. The follower instance performs this claim after all the assignment futures are waited for, so wouldn't that make sure that the corresponding unassigned tasks in this assignment are stopped within that timeout?

There are a few failure scenarios to think about,
1. If the futures are timing out, we are retrying endlessly, which won't let the instance claim the tokens. 
2. If there's an interrupted exception, we need to handle whether we need the hosts to claim the tokens or not. 

Another optimization may be to think about is,
In case when multiple streams are stopped, and a task for one particular stream cannot be stopped, should we handle the futures per stream/prefix to not block other stopping streams?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: +1

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: I checked with Jhora, if that's not the case. Completable Future only confirms that the task state is changed to shutdown. It does not check if the task thread has died or not. The task thread must die to avoid duplicate processing. ([Code block](https://github.com/linkedin/brooklin/blob/2ab7ce0065058913908660b576d93758a0b59594/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java#L770))

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: I didn't have an "else" path for this because I wanted it to be a no-op (i.e. attempt to claim non-existing token). Will add a debug log in the else path.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: No need for additional logging. The calling method `maybeClaimAssignmentTokensForStoppingStreams` already logs the result of this. Also I made this static for testing, can't use instance logger here.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Adding

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: This is covered with tests and already has logs in the important path (i.e. when it infers what streams are potentially stopping). I don't think there's value in adding logs in the first section where it breaks down the old and new assignments per connector.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Added a TODO comment to address metrics later

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -788,4 +788,18 @@ public Map<String, Map<Integer, Long>> getConsumptionLagForTask(DatastreamTask t
       return taskEntry.getConnectorTask().getKafkaTopicPartitionTracker().getConsumptionLag();
     }
   }
+
+  public List<String> getActiveTasks() {
+    List<DatastreamTask> tasks = new ArrayList<>();
+
+    synchronized (_runningTasks) {
```
jzakaryan, line 795: Tasks are synchronized across `AbstractKafkaConnector` when making transitions from running tasks to stopping tasks and finally being discarded. I'm locking on both of these collections here and getting tasks in both.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaConnector.java
```diff
@@ -788,4 +788,18 @@ public Map<String, Map<Integer, Long>> getConsumptionLagForTask(DatastreamTask t
...
+      synchronized (_tasksPendingStop) {
+        tasks.addAll(_runningTasks.keySet());
+        tasks.addAll(_tasksPendingStop.keySet());
+      }
+    }
+
+    _logger.debug("Found {} active tasks assigned to the connector", tasks.size());
+    return tasks.stream().map(DatastreamTask::getId).collect(Collectors.toList());
```
jzakaryan, line 803: I'm exposing task IDs instead of Task objects on purpose. Exposing Task carries risks of state corruption etc.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: This was copied over from the original code (i.e. code that calculates removed/added tasks). I did a conversion to HashSet to address this, but I think a better solution is to refactor this code. Most of these types can be converted to `Collection<DatastreamTask>` unless there's a specific need to use Set/List APIs. That will reduce the overall number of conversions in the code. It's out of scope from this PR, hence not touching it here.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -767,6 +769,71 @@ private Future<Boolean> dispatchAssignmentChangeIfNeeded(String connectorType, L
...
+    // The follower nodes don't keep an up-to-date view of the datastreams Zk node and making them listen to changes
+    // to datastreams node risks overwhelming the ZK server with reads. Instead, the follower is inferring the stopping
+    // streams from the task assignment for each connector type. Even if that inference results in a falsely identified
+    // stopping stream, the attempt to claim the token will be a no-op.
+    for (String connector : allConnectors) {
+      List<DatastreamTask> oldTasks = oldAssignmentPerConnector.getOrDefault(connector, Collections.emptyList());
+      List<DatastreamTask> newTasks = newAssignmentPerConnector.getOrDefault(connector, Collections.emptyList());
+      List<DatastreamTask> removedTasks = getRemovedTasks(newTasks, oldTasks);
+
+      List<Datastream> stoppingStreams = inferStoppingDatastreamsFromAssignment(newTasks, removedTasks);
```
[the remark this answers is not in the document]
jzakaryan: I think the word "infer" has enough semantic weight to hint at that.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -798,24 +797,33 @@ private void issueAssignmentTokensForStoppingDatastreams(List<DatastreamGroup> s
...
+        if (instance.equals(_instanceName)) {
+          LOG.info("Claiming assignment token for datastream: {}", streamName);
+        } else {
+          LOG.info("Revoking assignment token for datastream: {}, instance: {}", streamName, instance);
+        }
+        _zkclient.delete(tokenPath);
```
shrinandthakkar, line 822: We should have a ZkException catcher to know if any updates fail.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -767,6 +777,92 @@ private Future<Boolean> dispatchAssignmentChangeIfNeeded(String connectorType, L
...
+      Set<String> stoppingStreamNames = stoppingStreams.stream().map(Datastream::getName).collect(Collectors.toSet());
+
+      if (!stoppingStreamNames.isEmpty()) {
+        _log.info("Trying to claim assignment tokens for connector {}, streams: {}", connector, stoppingStreamNames);
+
+        Set<String> stoppingDatastreamTasks = removedTasks.stream().
+            filter(t -> stoppingStreamNames.contains(t.getTaskPrefix())).
+            map(DatastreamTask::getId).collect(Collectors.toSet());
+
+        if (PollUtils.poll(() -> connectorTasksHaveStopped(connector, stoppingDatastreamTasks),
```
shrinandthakkar, line 831: Should we instead put this in a loop for each datastream? Not a strong requirement, but this will prevent having dependencies on other stopping streams for a particular stream to stop.

shrinandthakkar (commented):
The overall code flow LGTM, just a couple of comments.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: `_zkclient` doesn't throw an exception but it returns a boolean telling the caller if the delete was successful or not. It also logs a warning message if the delete fails:

```
public boolean delete(final String path) {
    long startT = System.currentTimeMillis();
    boolean var13 = false;

    boolean success;
    try {
      try {
        var13 = true;
        this.retryUntilConnected(new Callable<Object>() {
          public Object call() throws Exception {
            ZkClient.this.getConnection().delete(path);
            return null;
          }
        });
        success = true;
      } catch (ZkNoNodeException var14) {
        success = false;
        if (LOG.isDebugEnabled()) {
          LOG.debug("zkclient {}, Failed to delete path {}, znode does not exist!", this._uid, path);
        }
      }

      this.record(path, (byte[])null, startT, AccessType.WRITE);
      var13 = false;
    } catch (Exception var15) {
      this.recordFailure(path, AccessType.WRITE);
      LOG.warn("zkclient {}, Failed to delete path {}! ", new Object[]{this._uid, path, var15});
      throw var15;
    } finally {
      if (var13) {
        long endT = System.currentTimeMillis();
        if (LOG.isTraceEnabled()) {
          LOG.trace("zkclient {} delete, path: {}, time {} ms", new Object[]{this._uid, path, endT - startT});
        }

      }
    }
```
I don't think there's much we need to do here.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: It feels like a micro-optimization. The poll is already parallel for connectors, and also, this whole method is scheduled to run asynchronously in `handleAssignmentChange`.

shrinandthakkar: In some scenarios, a stuck pending-to-stop task for a particular stream could block a successful stopping datastream for the same connector, hence I suggested that.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
shrinandthakkar: The code you pasted does seem to throw the exception "throw var15;". And I was concerned here since there's no catch block in the code chain after that if that exception is thrown. 

And since the scheduled executor is running that thread, that executor thread would die, causing work stoppage, I guess if I am not wrong. Also, in this case, it does not harm as a new task to this executor would create a new thread, but catching that exception should make things simpler.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Stop requests usually affect one stream, there's no support for bulk stops/restarts, so in production scenarios multiple stopping streams (or rather, streams that are inferred as stopping) are virtually unexpected. I can't think of a scenario that would trigger a bulk stop/restart on the server. 
It's a tradeoff of complicating the code and the value we get out of it. It's not hard for me to add it, I'll let @vmaheshw be the tie breaker here.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
jzakaryan: Good catch. I didn't realize it's getting re-thrown. Added exception handling and error logging in ZkAdapter.

shrinandthakkar (approved):

datastream-server-api/src/main/java/com/linkedin/datastream/server/api/connector/Connector.java
```diff
@@ -66,6 +66,15 @@ public interface Connector extends MetricsAware, DatastreamChangeListener {
    */
   void initializeDatastream(Datastream stream, List<Datastream> allDatastreams) throws DatastreamValidationException;
 
+  /**
+   * Returns a list with IDs for tasks that are active (i.e. running or pending stop)
+   * @return A list with task IDs that are currently running on the coordinator
+   * @throws UnsupportedOperationException if not implemented by Connector classes.
+   */
+  default List<String> getActiveTasks() {
+    throw new UnsupportedOperationException("Running tasks API is not supported unless implemented by coordinators");
```
vmaheshw, line 75: Plz fix the exception message.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -3256,6 +3460,11 @@ public String toString() {
     public List<BrooklinMetricInfo> getMetricInfos() {
       return null;
     }
+
+    @Override
+    public List<String> getActiveTasks() {
```
vmaheshw, line 3465: Have you added a test that covers the following scenarios:
a) Task list has some tasks for the datastream that eventually get cleared.
b) Task list has some tasks for the datastream that the task thread continues to run.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -710,6 +714,16 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
       retryHandleAssignmentChange(isDatastreamUpdate);
     }
 
+    if (_config.getEnableAssignmentTokens()) {
+      try {
+        // Queue assignment token claim task
+        _executor.schedule(() -> maybeClaimAssignmentTokensForStoppingStreams(newAssignment, oldAssignment),
```
vmaheshw, line 720: This `_executor` object is used for CoordinatorEventQueue. Calling this API using the same executor and any delay in this can result in a delay in a heartbeat and other operations. Was this intentional?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -742,19 +756,15 @@ private Future<Boolean> dispatchAssignmentChangeIfNeeded(String connectorType, L
...
-    List<DatastreamTask> removedTasks;
     List<DatastreamTask> oldAssignment = _assignedDatastreamTasks.values()
         .stream()
         .filter(t -> t.getConnectorType().equals(connectorType))
         .collect(Collectors.toList());
+    List<DatastreamTask> addedTasks = getAddedTasks(assignment, oldAssignment);
+    List<DatastreamTask> removedTasks = getRemovedTasks(assignment, oldAssignment);
 
     // if there are any difference in the list of assignment. Note that if there are no difference
     // between the two lists, then the connector onAssignmentChange() is not called.
```
vmaheshw, line 767: The refactored code got moved before the comment, is that intentional?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: Parallel restarts, or stopping all the datastreams, usually happens during outage recovery. In that scenario, if the stopping group of datastreams is stuck because of one, do we have an alternate way to recover (maybe force-stopping the datastream)?

I agree with both of you here. Let's add this as comments to start with in case this becomes a pain in the future. We can implement it then.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Lines 745, 746, 754 through 757 were removed because I introduced methods that get added/removed tasks. This is done intentionally.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: This was intentional. I didn't want to put the code on the thread that executes `handleAssignmentChange` and I wanted to make the checks for stopped tasks and token claims asynchronous. `maybeClaimAssignmentTokensForStoppingStreams` has a blocking wait for tasks to stop. What would be a better way to execute this on a separate thread? Do you want me to introduce a new executor service?

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
jzakaryan: (a) is covered in `testClaimAssignmentTokensForStoppingStreams`. Here I use the TestHookConnector to simulate that tasks have stopped (active tasks list is empty).
(b) is covered in `testTokensNotClaimedForConnectorThatFailedToStop`. Here the mock for Connector always keeps some of the tasks "running"

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: A new executor service.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -716,11 +716,11 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
 
     if (_config.getEnableAssignmentTokens()) {
       try {
+        ExecutorService executor = Executors.newSingleThreadExecutor();
```
vmaheshw, line 719: I know that we don't follow this practice from the beginning, but we started this once we got feedback to have meaningful name in the threaddump and looking at review process of Samza/Cruise Control.

Please use the other API which helps with naming the executor thread. 

For eg:

 ` // An executor to spawn threads to close the producer.
  private final ExecutorService _producerCloseExecutorService = Executors.newSingleThreadExecutor(
      new ThreadFactoryBuilder().setNameFormat("KafkaProducerWrapperClose-%d").build());`

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -360,6 +366,14 @@ public void stop() {
       }
     }
 
+    // Shutdown executor services
+    if (_scheduledExecutor != null) {
+      _scheduledExecutor.shutdown();
```
jzakaryan, line 371: I'm calling the more graceful shutdown path here. No new submissions will be accepted but the tasks that have already been submitted will be lingering until completion. I could block and wait for all tasks to complete here. I could also force terminate everything (which is a best effort operation). I don't think more oppressive measures are required here.

vmaheshw (approved):

shrinandthakkar (approved):

PR #922, opened 2023-01-13
BMM Restart Improvements Part 3. Leader polls for unclaimed tokens before declaring a datastream STOPPED

jzakaryan:
This pull request is part of a series of changes that are meant to improve BMM Restart and make it easier to debug restart failures and trace them to the faulty hosts. Part 3 introduces changes to the leader that make it poll the tokens for stop assignment and make sure that all followers claimed their tokens before declaring a stream `STOPPED`. If at least one follower failed to claim their token, a detailed error message is logged and a metric is emitted. Unclaimed tokens are revoked by the leader. Poll timeout is configurable.

Part 1 – Introduction of assignment tokens and support for issuing tokens by the leader coordinator (https://github.com/linkedin/brooklin/pull/919)
Part 2 – Changes to the followers' handleAssignmentChange to make them claim the tokens issued by the leader. (https://github.com/linkedin/brooklin/pull/921)
**Part 3 – Changes to the leader to make it poll the ZooKeeper and wait for the assignment change (stop) to be propagated and executed by the cluster.**

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -1351,7 +1351,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
             map(AssignmentToken::getIssuedFor).collect(Collectors.toSet());
         _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
             _config.getStopPropagationTimeout(), failedStreams, hosts);
-        revokeUnclaimedAssignmentTokens(unclaimedTokens);
+        revokeUnclaimedAssignmentTokens(unclaimedTokens, stoppingDatastreamGroups);
         _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_FAILED_STOPS, failedStreams.size());
       }
 
@@ -1401,13 +1401,22 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
     }
   }
 
-  private void revokeUnclaimedAssignmentTokens(Map<String, List<AssignmentToken>> unclaimedTokens) {
+  private void revokeUnclaimedAssignmentTokens(Map<String, List<AssignmentToken>> unclaimedTokens,
+      List<DatastreamGroup> stoppingDatastreamGroups) {
     _log.info("Revoking unclaimed tokens");
-    for (String stream : unclaimedTokens.keySet()) {
-      List<String> instances = unclaimedTokens.get(stream).stream().map(AssignmentToken::getIssuedFor).
+    Map<String, Datastream> datastreamMap = new HashMap<>();
+    stoppingDatastreamGroups.forEach(dg -> dg.getDatastreams().forEach(ds -> datastreamMap.put(ds.getName(), ds)));
+    for (String streamName : unclaimedTokens.keySet()) {
+      Datastream stream = datastreamMap.get(streamName);
+
+      if (stream == null) {
+        _log.warn("Failed to claim token for unknown datastream: {}", streamName);
+        continue;
+      }
+
+      List<String> instances = unclaimedTokens.get(streamName).stream().map(AssignmentToken::getIssuedFor).
           collect(Collectors.toList());
-      // TODO Uncomment after #921
-      //instances.forEach(i -> _adapter.claimAssignmentTokensForDatastreams(Collections.singletonList(stream), i));
+      instances.forEach(i -> _adapter.claimAssignmentTokensForDatastreams(Collections.singletonList(stream), i));
     }
   }
 
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -1240,7 +1240,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
             map(AssignmentToken::getIssuedFor).collect(Collectors.toSet());
         _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
             _config.getStopPropagationTimeout(), failedStreams, hosts);
-        // TODO Revoke tokens that have remained unclaimed
+        revokeUnclaimedAssignmentTokens(unclaimedTokens);
         _metrics.updateKeyedMeter(CoordinatorMetrics.KeyedMeter.HANDLE_LEADER_DO_ASSIGNMENT_NUM_FAILED_STOPS,
             failedStreams.size());
       }
@@ -1288,6 +1288,16 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
         _eventQueue.put(CoordinatorEvent.createLeaderDoAssignmentEvent(cleanUpOrphanNodes));
         _leaderDoAssignmentScheduled.set(false);
       }, _config.getRetryIntervalMs(), TimeUnit.MILLISECONDS);
+    }
+  }
+
+  private void revokeUnclaimedAssignmentTokens(Map<String, List<AssignmentToken>> unclaimedTokens) {
+    _log.info("Revoking unclaimed tokens");
+    for (String stream : unclaimedTokens.keySet()) {
+      List<String> instances = unclaimedTokens.get(stream).stream().map(AssignmentToken::getIssuedFor).
+          collect(Collectors.toList());
+      // TODO Uncomment after #921
+      //instances.forEach(i -> _adapter.claimAssignmentTokensForDatastreams(Collections.singletonList(stream), i));
     }
   }
 
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -2190,6 +2190,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
      */
     public enum Meter {
       NUM_REBALANCES("numRebalances"),
+      NUM_FAILED_STOPS("numFailedStops"),
       NUM_ASSIGNMENT_CHANGES("numAssignmentChanges"),
       NUM_PARTITION_ASSIGNMENTS("numPartitionAssignments"),
       NUM_PARTITION_MOVEMENTS("numPartitionMovements"),
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -166,6 +166,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
 
   private static final long EVENT_THREAD_LONG_JOIN_TIMEOUT = 90000L;
   private static final long EVENT_THREAD_SHORT_JOIN_TIMEOUT = 3000L;
+  // how long should the leader wait between consecutive polls to zookeeper to confirm that the datastreams have stopped
   private static final long STOP_PROPAGATION_RETRY_MS = 5000L;
 
   private static final Duration ASSIGNMENT_TIMEOUT = Duration.ofSeconds(90);
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -56,6 +56,7 @@ import com.linkedin.datastream.common.DatastreamStatus;
 import com.linkedin.datastream.common.DatastreamTransientException;
 import com.linkedin.datastream.common.DatastreamUtils;
 import com.linkedin.datastream.common.ErrorLogger;
+import com.linkedin.datastream.common.PollUtils;
 import com.linkedin.datastream.common.VerifiableProperties;
 import com.linkedin.datastream.metrics.BrooklinCounterInfo;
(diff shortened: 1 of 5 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1225,12 +1228,33 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
         _adapter.updateAllAssignments(newAssignmentsByInstance);
       }
 
+      Set<String> failedStreams = Collections.emptySet();
+      // Poll the zookeeper to ensure that hosts claimed assignment tokens for stopping streams
+      if (_config.getEnableAssignmentTokens() &&
+          !PollUtils.poll(() -> _adapter.getNumUnclaimedTokensForDatastreams(stoppingDatastreamGroups) == 0,
```
shrinandthakkar, line 1234: Let's listen to any exceptions that `updateAllAssignmentsAndIssueTokens` might throw and only perform this poll in the non-exception path, otherwise in worst-case scenarios, if the ZK is not updated with the tokens in `updateAllAssignmentsAndIssueTokens`, the count would always be 0, and this poll condition would meet.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1225,12 +1228,33 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
...
+        failedStreams = unclaimedTokens.keySet();
+        Set<String> hosts = unclaimedTokens.values().stream().flatMap(List::stream).
+            map(AssignmentToken::getIssuedFor).collect(Collectors.toSet());
+        _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
+            _config.getStopPropagationTimeout(), failedStreams, hosts);
+        revokeUnclaimedAssignmentTokens(unclaimedTokens);
+        _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_FAILED_STOPS, failedStreams.size());
+      }
+
+      // TODO Explore introduction of UNKNOWN/WARN state for streams that failed to stop
```
shrinandthakkar, line 1247: Should we keep the streams in the 'STOPPING" state if the tokens cannot be claimed?

Should we fall back to the current way of stopping streams if any of this ZK business goes wrong in worst-case scenarios? And since we have a 60 seconds timeout anyway, should we mark the stream as Stopped (as we currently do)?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1225,12 +1228,33 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
         _adapter.updateAllAssignments(newAssignmentsByInstance);
       }
 
+      Set<String> failedStreams = Collections.emptySet();
+      // Poll the zookeeper to ensure that hosts claimed assignment tokens for stopping streams
+      if (_config.getEnableAssignmentTokens() &&
+          !PollUtils.poll(() -> _adapter.getNumUnclaimedTokensForDatastreams(stoppingDatastreamGroups) == 0,
+            STOP_PROPAGATION_RETRY_MS, _config.getStopPropagationTimeout())) {
```
shrinandthakkar, line 1235: The current timeout in the stop API handler (in DatastreamResources) is also 60 seconds, so I think we should either update that or either update this `getStopPropagationTimeout` (since this is also 60 seconds) otherwise there could be inconsistencies between what's returned by the stop API and the current state of the datastream.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: On a second thought, I don't think we should do that.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -163,6 +164,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
 
   private static final long EVENT_THREAD_LONG_JOIN_TIMEOUT = 90000L;
   private static final long EVENT_THREAD_SHORT_JOIN_TIMEOUT = 3000L;
+  private static final long STOP_PROPAGATION_RETRY_MS = 5000L;
```
vmaheshw, line 167: Should we start the practice to add the reasoning for the new timer values?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -196,6 +198,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   // handleEvent is synchronized and downstream code can misbehave.
   private final Duration _heartbeatPeriod;
 
+
```
vmaheshw, line 201: nit: empty line

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: I can add comments for the newly introduced ones. Would that help?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Removing

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: I was having the same debate. I think it's too risky to change that behavior. I'm planning to get back to this and introduce a new state in the state machine after testing the rest of it (at least the metrics etc need to work really well).

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1225,12 +1228,33 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
         _adapter.updateAllAssignments(newAssignmentsByInstance);
       }
 
+      Set<String> failedStreams = Collections.emptySet();
+      // Poll the zookeeper to ensure that hosts claimed assignment tokens for stopping streams
+      if (_config.getEnableAssignmentTokens() &&
```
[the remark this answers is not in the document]
jzakaryan: I know I'm making two checks for the `enableAssignmentTokens` config where I can make just one, but the current structure makes more sense to me. Lines 1220-1228 deal with assignment, and the section below does the token poll business. The comments also describe what each section does.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: - If there's no tokens written to Zk (in the worst case scenario that you described), the poll will exit immediately. The condition for the if statement, however, won't be met, and the body of the if statement won't execute. This is expected behavior (no tokens were written, nothing to do here).
- In case of Zk create failure, I don't like the idea of handling Zk Exceptions in `handleLeaderDoAssignment`. We already have generic exception handling in `handleEvent` and it logs the exception messages. There's not much we can do in `handleLeaderDoAssignment`

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Help me understand how these two timeouts being 60 seconds poses a problem. Is the concern here that the timeout in rest.li API will hit earlier than the stop propagation timeout?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Link to exception handling code for reference:

https://github.com/linkedin/brooklin/blob/8bc670bd8bad20a6c2ff1fb335f3f3018c5b9b7b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java#L933

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Discussed offline. We need to increase the timeout in the rest.li DMS service to give the request handler enough time to hear back from the cluster leader. This change will be included in the next PR.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1378,6 +1401,16 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
     }
   }
 
+  private void revokeUnclaimedAssignmentTokens(Map<String, List<AssignmentToken>> unclaimedTokens) {
+    _log.info("Revoking unclaimed tokens");
+    for (String stream : unclaimedTokens.keySet()) {
+      List<String> instances = unclaimedTokens.get(stream).stream().map(AssignmentToken::getIssuedFor).
+          collect(Collectors.toList());
+      // TODO Uncomment after #921
```
shrinandthakkar, line 1409: will you uncommenting this in this PR ?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Done

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Done

shrinandthakkar (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorConfig.java
```diff
@@ -32,10 +32,16 @@ public final class CoordinatorConfig {
...
+
+  // how long should the leader poll the zookeeper to confirm that stopping datastreams have stopped before giving up
+  public static final String CONFIG_STOP_PROPAGATION_TIMEOUT_MS = PREFIX + "stopPropagationTimeout";
+  // how long should the coordinator wait for a connector's tasks to stop before giving up
   public static final String CONFIG_TASK_STOP_CHECK_TIMEOUT_MS = PREFIX + "taskStopCheckTimeoutMs";
+  // how long should the coordinator wait in between two consecutive calls to check if a connector's tasks have stopped
   public static final String CONFIG_TASK_STOP_CHECK_RETRY_PERIOD_MS = PREFIX + "taskStopCheckRetryPeriodMs";
 
   public static final int DEFAULT_MAX_ASSIGNMENT_RETRY_COUNT = 100;
+  public static final long DEFAULT_STOP_PROPAGATION_TIMEOUT_MS = 60000;
```
vmaheshw, line 44: nit: 60000 to 60 * 1000 (simiar to others)

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1337,12 +1339,33 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
         _adapter.updateAllAssignments(newAssignmentsByInstance);
       }
 
+      Set<String> failedStreams = Collections.emptySet();
+      // Poll the zookeeper to ensure that hosts claimed assignment tokens for stopping streams
+      if (_config.getEnableAssignmentTokens() &&
+          !PollUtils.poll(() -> _adapter.getNumUnclaimedTokensForDatastreams(stoppingDatastreamGroups) == 0,
+            STOP_PROPAGATION_RETRY_MS, _config.getStopPropagationTimeout())) {
+        Map<String, List<AssignmentToken>> unclaimedTokens =
+            _adapter.getUnclaimedAssignmentTokensForDatastreams(stoppingDatastreamGroups);
```
vmaheshw, line 1348: If there is a bug and the list of hosts that need to claim the token is incorrect, how will we debug it?

What happens if the instance dies in between and is not able to claim the token? and then if the task moves to another host?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1337,12 +1339,33 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
...
-          datastream.setStatus(DatastreamStatus.STOPPED);
-          if (!_adapter.updateDatastream(datastream)) {
-            _log.warn("Failed to update datastream: {} to stopped state", datastream.getName());
-            succeeded = false;
+          // Only streams that were confirmed to have stopped successfully will be transitioned to STOPPED state
+          if (!failedStreams.contains(datastream.getName())) {
+            datastream.setStatus(DatastreamStatus.STOPPED);
+            if (!_adapter.updateDatastream(datastream)) {
+              _log.warn("Failed to update datastream: {} to stopped state", datastream.getName());
+              succeeded = false;
```
vmaheshw, line 1367: @shrinandthakkar Can you double check that the recent issue that you saw with the previous assignment getting replaced by new assignment and some cleanup is not completed, will not reoccur in this scenario?

vmaheshw (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: The cleanup code will be in Part 4. I'll see what other logs/metrics I can emit to help us debug issues with hosts dying.

PR #924, opened 2023-01-26
BMM Restart Improvements Part 4. Edge cases, cleanup and general improvements

jzakaryan:
This pull request is part of a series of changes that are meant to improve BMM Restart and make it easier to debug restart failures and trace them to the faulty hosts. Part 4 handles edge cases such as leader failover. It also introduces the following improvements suggested in the previous parts:

- A metric for the number of datastreams that are inferred as stopping has been added to help debug the feature (@vmaheshw's suggestion)
- Timeout for the rest.li request handler node waiting on the leader to mark the datastream as STOPPED has been increased from 60 seconds to 90 seconds. This is done to make sure the leader will have time to handle stop propagation (which has a 60 second timeout of its own). (@shrinandthakkar's suggestion)

Part 1 – Introduction of assignment tokens and support for issuing tokens by the leader coordinator (https://github.com/linkedin/brooklin/pull/919)
Part 2 – Changes to the followers' handleAssignmentChange to make them claim the tokens issued by the leader. (https://github.com/linkedin/brooklin/pull/921)
Part 3 – Changes to the leader to make it poll the ZooKeeper and wait for the assignment change (stop) to be propagated and executed by the cluster. ((https://github.com/linkedin/brooklin/pull/922)
**Part 4 – Edge cases and cleanup and general improvements**

### DIFF
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -483,7 +483,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
             // leader failed to do so. This needs to be used only if attempts to stop a stream regularly fail
             LOG.info("Force stop for datastream {} requested. Setting the datastream to STOPPED state and invoking cleanup",
                 d.getName());
-            _store.deleteAssignmentTokens(d.getName());
+            _store.forceCleanupDatastream(d.getName());
             d.setStatus(DatastreamStatus.STOPPED);
             _store.updateDatastream(d.getName(), d, false);
           } else {
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -479,6 +479,10 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
           LOG.warn("Datastream {} is already in {} state. Notifying leader to initiate transition", d,
               d.getStatus());
           if (force) {
+            // force stop recover the setup (i.e. set the datastream to STOPPED state) and do cleanup in case
+            // leader failed to do so. This needs to be used only if attempts to stop a stream regularly fail
+            LOG.info("Force stop for datastream {} requested. Setting the datastream to STOPPED state and invoking cleanup",
+                d.getName());
             _store.deleteAssignmentTokens(d.getName());
             d.setStatus(DatastreamStatus.STOPPED);
             _store.updateDatastream(d.getName(), d, false);
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -478,7 +478,13 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
           // this check helps in preventing any datastream from being stuck in STOPPING state indefinitely
           LOG.warn("Datastream {} is already in {} state. Notifying leader to initiate transition", d,
               d.getStatus());
-          _store.updateDatastream(d.getName(), d, true);
+          if (force) {
+            _store.deleteAssignmentTokens(d.getName());
+            d.setStatus(DatastreamStatus.STOPPED);
+            _store.updateDatastream(d.getName(), d, false);
+          } else {
+            _store.updateDatastream(d.getName(), d, true);
+          }
           _store.deleteDatastreamNumTasks(d.getName());
           // invoke post datastream state change action for recently stopped datastream
           invokePostDSStateChangeAction(d);
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -102,7 +102,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
   // To support retries on the request timeouts
   public static final String CONFIG_STOP_TRANSITION_TIMEOUT_MS = "stopTransitionTimeoutMs";
   public static final String CONFIG_STOP_TRANSITION_RETRY_PERIOD_MS = "stopTransitionRetryPeriodMs";
-  private static final Long STOP_TRANSITION_TIMEOUT_MS_DEFAULT = Duration.ofMillis(60000).toMillis();
+  private static final Long STOP_TRANSITION_TIMEOUT_MS_DEFAULT = Duration.ofMillis(90000).toMillis();
   private static final Long STOP_TRANSITION_RETRY_PERIOD_MS_DEFAULT = Duration.ofMillis(1000).toMillis();
 
   private final DatastreamStore _store;
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
@@ -71,5 +71,5 @@ public interface DatastreamStore {
    * Delete the assignment tokens (if any) associated with the provided key
    * @param key Name of the datastream whose assignment tokens have to be deleted
    */
-  void deleteAssignmentTokens(String key);
+  void forceCleanupDatastream(String key);
 }
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
@@ -66,4 +66,10 @@ public interface DatastreamStore {
    * @param key Name of the original datastream whose numTasks znode is to be deleted
    */
   void deleteDatastreamNumTasks(String key);
+
+  /**
+   * Delete the assignment tokens (if any) associated with the provided key
+   * @param key Name of the datastream whose assignment tokens have to be deleted
+   */
+  void deleteAssignmentTokens(String key);
 }
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
@@ -13,6 +13,7 @@ import java.util.stream.Stream;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.Validate;
+import org.apache.helix.zookeeper.exception.ZkClientException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -243,7 +244,7 @@ public class ZookeeperBackedDatastreamStore implements DatastreamStore {
   }
 
   @Override
-  public void deleteAssignmentTokens(String key) {
+  public void forceCleanupDatastream(String key) {
     String assignmentTokensPath = KeyBuilder.datastreamAssignmentTokens(_cluster, key);
 
     if (!_zkClient.exists(assignmentTokensPath)) {
@@ -251,7 +252,11 @@ public class ZookeeperBackedDatastreamStore implements DatastreamStore {
       return;
     }
 
-    _zkClient.deleteRecursively(assignmentTokensPath);
+    try {
+      _zkClient.deleteRecursively(assignmentTokensPath);
+    } catch (ZkClientException ex) {
+      LOG.error("Failed to cleanup assignment tokens for {}", key, ex);
+    }
   }
 
   private void notifyLeaderOfDataChange() {
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
@@ -242,6 +242,18 @@ public class ZookeeperBackedDatastreamStore implements DatastreamStore {
     _zkClient.delete(path);
   }
 
+  @Override
+  public void deleteAssignmentTokens(String key) {
+    String assignmentTokensPath = KeyBuilder.datastreamAssignmentTokens(_cluster, key);
+
+    if (!_zkClient.exists(assignmentTokensPath)) {
+      LOG.info("Assignment tokens path clear for datastream {}. Nothing to delete", key);
+      return;
+    }
+
+    _zkClient.deleteRecursively(assignmentTokensPath);
+  }
+
   private void notifyLeaderOfDataChange() {
     String dmsPath = KeyBuilder.datastreams(_cluster);
     // Update the /dms to notify that coordinator needs to act on a deleted or changed datastream.
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -1298,10 +1298,10 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   }
 
   /*
-   * If cleanUpOrphanNodes is set to true, it cleans up the orphan connector tasks not assigned to
+   * If isNewlyElectedLeader is set to true, it cleans up the orphan connector tasks not assigned to
    * any instance after old unused tasks are cleaned up.
    */
-  private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
+  private void handleLeaderDoAssignment(boolean isNewlyElectedLeader) {
     boolean succeeded = true;
     List<String> liveInstances = Collections.emptyList();
     Map<String, Set<DatastreamTask>> previousAssignmentByInstance = Collections.emptyMap();
@@ -1314,7 +1314,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
       _log.info("stopping datastreams: {}", stoppingDatastreamGroups);
       onDatastreamChange(datastreamGroups);
 
-      if (cleanUpOrphanNodes) {
+      if (isNewlyElectedLeader) {
         performPreAssignmentCleanup(datastreamGroups);
       }
 
@@ -1349,10 +1349,18 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
         failedStreams = unclaimedTokens.keySet();
         Set<String> hosts = unclaimedTokens.values().stream().flatMap(List::stream).
             map(AssignmentToken::getIssuedFor).collect(Collectors.toSet());
-        _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
-            _config.getStopPropagationTimeout(), failedStreams, hosts);
+
+        if (!isNewlyElectedLeader) {
+          _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
+              _config.getStopPropagationTimeout(), failedStreams, hosts);
+          _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_FAILED_STOPS, failedStreams.size());
+        } else {
+          _log.warn("Stop may have failed to propagate within {}ms for streams: {}. The newly elected leader was " +
+              "expecting the hosts {} to claim tokens but they didn't",
+              _config.getStopPropagationTimeout(), failedStreams, hosts);
+        }
+
         revokeUnclaimedAssignmentTokens(unclaimedTokens, stoppingDatastreamGroups);
-        _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_FAILED_STOPS, failedStreams.size());
       }
 
       // TODO Explore introduction of UNKNOWN/WARN state for streams that failed to stop
@@ -1383,7 +1391,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
       instances.add(PAUSED_INSTANCE);
       _adapter.cleanUpDeadInstanceDataAndOtherUnusedTasks(previousAssignmentByInstance,
           newAssignmentsByInstance, instances);
-      if (cleanUpOrphanNodes) {
+      if (isNewlyElectedLeader) {
         performCleanupOrphanNodes();
       }
       _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_REBALANCES, 1);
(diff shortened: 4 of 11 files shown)


datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -102,7 +102,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
   // To support retries on the request timeouts
   public static final String CONFIG_STOP_TRANSITION_TIMEOUT_MS = "stopTransitionTimeoutMs";
   public static final String CONFIG_STOP_TRANSITION_RETRY_PERIOD_MS = "stopTransitionRetryPeriodMs";
-  private static final Long STOP_TRANSITION_TIMEOUT_MS_DEFAULT = Duration.ofMillis(60000).toMillis();
+  private static final Long STOP_TRANSITION_TIMEOUT_MS_DEFAULT = Duration.ofMillis(90000).toMillis();
```
ehoner, line 105: discussion: Was the original value of the timeout based, whole or in part, on Nurse? My understanding is that part of motivation for this work was that the current `STOP`/`START`/`RESTART` process is unreliable given our automation will timeout after 60s. Can this timeout be lowered to 60s and any of its dependents lowered accordingly?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1349,10 +1349,18 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
         failedStreams = unclaimedTokens.keySet();
         Set<String> hosts = unclaimedTokens.values().stream().flatMap(List::stream).
```
ehoner, line 1350: out-of-curiousity: Is there a reason why entry into this block is predicated on `getNumUnclaimedTokensForDatastreams` (time bounded) but requires `getUnclaimedAssignmentTokensForDatastreams` (unbounded)? These two requests could have different results. Any non-zero list is an equivalent test and an error could be interpreted as "success not confirmed". Since there is no recovery here, it seems or important to make a best-effort attempt to verify no unclaimed tokens are present and honor the time constraint.

I realize this was added in #922  and not here, but this question wasn't addressed in any of comments there.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
jzakaryan: This is the timeout of the rest.li node handling the stop request, i.e. how long will it wait for the leader to mark the datastream STOPPED. I don't think the choice of this was based on the Nurse, @shrinandthakkar can correct me here if I'm wrong.

The reason we decided to increase this is because now the leader also has a 60 second timeout to hear from followers that they stopped the stream. With this change we're making sure the leader's 60 seconds poll times out before the rest.li requests' timeout. Eventually, with Phase 2, this will be revisited.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Can you explain how `getNumUnclaimedTokensForDatastreams` is time bounded? What do you mean by time bounded here?

The reason I have `getNumUnclaimedTokensForDatastreams` in the if statement condition is because it is faster. It doesn't involve reading the zookeeper data and deserializing it. I understand that the results of `getNumUnclaimedTokensForDatastreams` and `getUnclaimedAssignmentTokensForDatastreams` can be inconsistent in the edge cases. By the time the leader calls `getUnclaimedAssignmentTokensForDatastreams` their number may be lower than what was earlier reported. It can even be zero. This, however, won't affect the outcome.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1349,10 +1349,18 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
         failedStreams = unclaimedTokens.keySet();
```
shrinandthakkar, line 1349: Do you think we should be handling this better? When a stop request fails, the datastream will be stuck in the STOPPING state. Since the leader will update the ZK with the newer assignment (removing the stopping tasks) and would also clean up the _liveTaskMap of the zkAdapter, a subsequent stopping request would not be handled properly since the leader would not have the host to task assignment for that stuck STOPPING state datastream to issue tokens for the hosts hosting tasks for stopping datastream. Thus the stream would forever be stuck in the STOPPING state.

I am trying to think if we can somehow recover from this in code, like should we at least somehow maintain a state where we can retrigger the stop request so that the leader could reattempt to stop this datastream.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1349,10 +1349,18 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
...
         Set<String> hosts = unclaimedTokens.values().stream().flatMap(List::stream).
             map(AssignmentToken::getIssuedFor).collect(Collectors.toSet());
-        _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
-            _config.getStopPropagationTimeout(), failedStreams, hosts);
+
+        if (!isNewlyElectedLeader) {
+          _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
+              _config.getStopPropagationTimeout(), failedStreams, hosts);
+          _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_FAILED_STOPS, failedStreams.size());
```
shrinandthakkar, line 1356: In case of multiple stopping datastreams, even if one datastream fails to stop, all of them will be marked as failed stop here. Should we fix that?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -102,7 +102,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
   // To support retries on the request timeouts
```
shrinandthakkar, line 102: Do we already have or should we have some test cases around these failure-causing / edge-case scenarios? 
1. Leader failure / kill
2. Follower failure / kill
3. Multiple stopping streams.
4. Stuck in the STOPPING state.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: oops, I think I overlooked this. this is correct!

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: Added a section regarding this in the supporting document. Setting up a meeting to discuss this.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
ehoner: > Can you explain how `getNumUnclaimedTokensForDatastreams` is time bounded? What do you mean by time bounded here?

I was referring to the `PollUtils` wrapper and timeout. This wrapped request will be interrupted if incomplete while the result from `getUnclaimedAssignmentTokensForDatastreams` will not.
```java
if (_config.getEnableAssignmentTokens() &&
        !PollUtils.poll(() -> _adapter.getNumUnclaimedTokensForDatastreams(stoppingDatastreamGroups) == 0,
        STOP_PROPAGATION_RETRY_MS, _config.getStopPropagationTimeout())) {
    Map<String, List<AssignmentToken>> unclaimedTokens =
            _adapter.getUnclaimedAssignmentTokensForDatastreams(stoppingDatastreamGroups);
````

----

I agree that the `..Num..` is faster and do not believe there is anything incorrect with either method; but I am not sure the faster option is valuable though. For the happy path, no unassigned tokens, the request will be fast, because `zkclient.getChildren()` will be be empty. In the failure case, identifying the failures, unclaimed tokens, is required. If queried with `...Num...`, a second query is necessary to retrieve the unassigned tokens. However, the second query may be have a different result from the first. It may be better to query once and capture state clearly if we plan on debugging unclaimed token issues. 

To reclaim some performance, the deserialization can be deferred and return a `List<String>`. The error handling block or `revokeUnclaimedAssignmentTokens` can convert to `AssignmentToken`s. (I don't have a preference for deferring deserialization.)

File: datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java (hunk 3 above)
Comment (jzakaryan) [reply datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java:1349]: Force stop path was added as a recovery for stream stuck in STOPPING state. Also a configuration is added that stops the streams even if stop failed/timed out on some nodes. Both of these mechanisms are interim solutions to avoid any regressions. A more comprehensive solution to be implemented as part of Phase 2 (a restart endpoint).

File: datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java (hunk 5 above)
Comment (jzakaryan) [reply datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java:102]: Added a test case that simulates leader failover. I also added tests for the new force stop mechanism in DatastreamResources.

For follower failure/multiple stopping streams, I'm not sure what else to add. If there's a particular piece of code you want to have covered with tests (and which is not covered yet), let'd discuss.

File: datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java (hunk 2 above)
Comment (jzakaryan) [reply datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java:1350]: @ehoner `getUnclaimedAssignmentTokens` might return fewer tokens than `..Num..`, because the invocation order is `..Num..` followed by `getUnclaimed...`. So the fact that these are two separate methods won't change the overall outcome. At the end of the day, the remaining tokens (whatever is returned by the last call) will get processed.

I would want to avoid exposing List<String> and leaking serialized tokens outside of the `ZkAdapter`. Classes that use ZkAdapter shouldn't worry about ser/de aspect of ZK storage.

Comment (ehoner) [reply datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java:1350]: Discussed offline. 
tldr; The issue of detecting different results between the two queries isn't of concern.

Differing results between the first request to `..Num..` and token retrieval does not produce any material behavior changes. If unclaimed token(s) were detected by the first query, but claimed before the second, the only actions are logging and `revokeUnclaimedTokens`. The log statement will show an empty set and `revoke..` will be a noop. More importantly, though no person will be signaled to investigate and no "corrective" actions would be required.

**Review by ehoner — APPROVED** [review]

File: datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java (hunk 5 above)
Comment (shrinandthakkar) [reply datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java:102]: For follower failure, I was inferring the scenario where a follower fails to claim tokens, and then the leader cleans up for it. And for multiple stopping streams case, to see if more than one stream stopping is handled as expected.

File: datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java [hunk 6]
```diff
@@ -478,7 +478,13 @@ public ActionResult<Void> stop(@PathKeysParam PathKeys pathKeys,
           // this check helps in preventing any datastream from being stuck in STOPPING state indefinitely
           LOG.warn("Datastream {} is already in {} state. Notifying leader to initiate transition", d,
               d.getStatus());
-          _store.updateDatastream(d.getName(), d, true);
+          if (force) {
+            _store.deleteAssignmentTokens(d.getName());
+            d.setStatus(DatastreamStatus.STOPPED);
+            _store.updateDatastream(d.getName(), d, false);
+          } else {
+            _store.updateDatastream(d.getName(), d, true);
```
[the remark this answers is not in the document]
jzakaryan: The purpose of this call is to recover the setup and to do cleanup of pending tokens in case leader failed to do so. This is a last resort call and this should be used sparingly.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
vmaheshw: We need a way to set dependencies on the other timers in the code itself. The reasoning for 90 second related to another 60 second timeout, should also reflect in the code. 

for eg: TimeoutA= 60sec TimeoutB = TimeoutA + 30sec rather than TimeoutA=60sec and TimeoutB=30sec as former helps in understanding the dependency.

Also, please add comments.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -478,7 +478,13 @@ public ActionResult<Void> stop(@PathKeysParam PathKeys pathKeys,
           // this check helps in preventing any datastream from being stuck in STOPPING state indefinitely
           LOG.warn("Datastream {} is already in {} state. Notifying leader to initiate transition", d,
               d.getStatus());
-          _store.updateDatastream(d.getName(), d, true);
+          if (force) {
```
vmaheshw, line 481: 1. Logs should reflect that it was a force stop.
2. Add comments explaining the rationale.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
```diff
@@ -66,4 +66,10 @@ void updatePartitionAssignments(String key, Datastream datastream, HostTargetAss
    * @param key Name of the original datastream whose numTasks znode is to be deleted
    */
   void deleteDatastreamNumTasks(String key);
+
+  /**
+   * Delete the assignment tokens (if any) associated with the provided key
+   * @param key Name of the datastream whose assignment tokens have to be deleted
+   */
+  void deleteAssignmentTokens(String key);
```
vmaheshw, line 74: You are leaking implementation details in the DatastreamStore interface. Can you think of another generic name?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -242,6 +242,18 @@ public void deleteDatastreamNumTasks(String key) {
...
+  @Override
+  public void deleteAssignmentTokens(String key) {
+    String assignmentTokensPath = KeyBuilder.datastreamAssignmentTokens(_cluster, key);
+
+    if (!_zkClient.exists(assignmentTokensPath)) {
+      LOG.info("Assignment tokens path clear for datastream {}. Nothing to delete", key);
+      return;
+    }
+
+    _zkClient.deleteRecursively(assignmentTokensPath);
```
vmaheshw, line 254: This method can throw an exception. Can you please check once that it will not break anything? @shrinandthakkar is evaluating the other instances of this call in the code already.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1349,18 +1349,26 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
         failedStreams = unclaimedTokens.keySet();
         Set<String> hosts = unclaimedTokens.values().stream().flatMap(List::stream).
             map(AssignmentToken::getIssuedFor).collect(Collectors.toSet());
-        _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
-            _config.getStopPropagationTimeout(), failedStreams, hosts);
+
+        if (!isNewlyElectedLeader) {
```
vmaheshw, line 1353: This `if-else` block is unclear to me. Can you please add the comments explaining the logic?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -242,6 +242,18 @@ public void deleteDatastreamNumTasks(String key) {
     _zkClient.delete(path);
   }
 
+  @Override
+  public void deleteAssignmentTokens(String key) {
```
vmaheshw, line 246: This method should get called from the regular cleanup (when the leader claims the unclaimed tokens) as well, and then you may need thread safety constructs around it. Currently, there are two separate code flows for force token claim and leader claiming the tokens.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: I already documented this and we went over this. We're skipping emitting metrics in case of leader failover because the new leader doesn't have a way to know if streams that are pending to stop have claimed their tokens. So the new leader issues new tokens and revokes them. Emitting metric in this case can cause false alerts.

Added inline comments explaining this.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
jzakaryan: `Coordinator` can't import `ZookeeperBackedDatastreamStore`. It's relying on `ZkAdapter` instead to modify the state in ZK. I understand this adds a duplicate code path. But I think it's a cleaner way to do this rather than adding a new dependency and moving classes between packages to solve circular dependency issues.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
jzakaryan: I cant think of a better name. I'm open to suggestions.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
jzakaryan: The exception will be caught in `DatastreamResources` and propagated to the client as a rest.li service exception.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
jzakaryan: Done

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
vmaheshw: Without synchronization, both threads can try to delete the tokens simultaneously. One of the code blocks may see exceptions.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
vmaheshw: Something related to forcing datastream stop maybe..

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
jzakaryan: `claimAssignmentTokensForDatastreams`, which is the main path, has the following code:

```
if (_zkclient.exists(tokenPath)) {
        if (instance.equals(_instanceName)) {
          LOG.info("Claiming assignment token for datastream: {}", streamName);
        } else {
          LOG.info("Revoking assignment token for datastream: {}, instance: {}", streamName, instance);
        }
        try {
          _zkclient.delete(tokenPath);
        } catch (Exception ex) {
          LOG.error("Failed to delete token {}", tokenPath, ex);
        }
      } else {
        LOG.debug("Attempt to claim non-existing assignment token {}", tokenPath);
      }
```
Even if there's an Exception , it will print an error and swallow it. Also, Zookeeper delete is idempotent - it won't throw an exception if there are concurrent attempts to delete a node, it will simply return `false`.

I'm adding similar kind of `Exception` handling in the second path.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamStore.java
jzakaryan: renamed it to `forceCleanupDatastream`. Not particularly fond of it, but slightly better

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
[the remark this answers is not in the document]
shrinandthakkar: The only issue i have is since you [fetch the stopping datastreams from the _liveTaskMap](https://github.com/linkedin/brooklin/blob/091ce5d1e60106c72cf015363b997e2744d5e05e/datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java#L767), it will be empty when this call is made. 

Would that create any issues? Since it will not re-issue tokens I think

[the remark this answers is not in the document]
jzakaryan: I'm trying to make sure that I understand this correctly. The code link you pasted points to the `getStoppingDatastreamGroupInstances` method, which is used by the leader `Coordinator` to get the hosts that are processing tasks for a stopping stream.

The force stop that I added is supposed to be called only if the stream is stuck in STOPPING state indefinitely. Here's a snapshot of the cluster in that bad state:

Leader did a best effort to stop the stream. It issued tokens for the instances in the `liveTaskMap`, it then updated the `liveTaskMap` (in case of a stopping stream that would amount to removing the stream's tasks from the map). Leader polled the followers and waited for them to acknowledge that they stopped their tasks, and it didn't hear from at least one host. It revokes the said hosts' tokens, alerts that "something went wrong with this host" and refuses to initiate a STOPPING -> STOPPED transition.

At this stage all the token issue/claim/revoke business is done. If someone is calling force stop on a stream, the intent is not to force reissue tokens. It is to reset the state of the stream, cleanup leftover garbage in ZK, and let subsequent operations such as resume go through.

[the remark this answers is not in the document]
jzakaryan: Discussed offline. @shrinandthakkar's question was whether we need the old call to `_store.datastreamUpdate` that notifies the leader to initiate an assignment change. While it's true that second and subsequent calls to `stop()` without the force flag won't reissue new tokens and the leader won't have a way of getting feedback from followers, we can't simply remove/modify this call. This call is still useful in case the assignment tokens feature is disabled. Modifying this code path may introduce regressions.

shrinandthakkar (approved):

vmaheshw (approved):

PR #925, opened 2023-02-04
Moved the datastream stop check poll to a separate thread to avoid blocking the event queue

jzakaryan:
- After issuing a stop stream assignment, the leader polls the zookeeper to confirm that tasks have stopped. Currently that poll happens in the main event thread. 
- There is also a requirement that the coordinator should process all of its events in less than 30 seconds. If this requirement is not met the coordinator won't be able to punch heartbeats.
- If the coordinator is configured to wait for stop propagation for a prolonged time, it will block the event queue for more than 30 seconds.

This change fixes the problem by moving the task to wait for stop request to propagate to a separate thread.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -1335,39 +1335,16 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
       // and schedule a retry. The retry should be able to diff the remaining ZooKeeper work
       if (_config.getEnableAssignmentTokens()) {
         _adapter.updateAllAssignmentsAndIssueTokens(newAssignmentsByInstance, stoppingDatastreamGroups);
+        try {
+          _tokenClaimExecutor.submit(() -> waitForStopToPropagateAndMarkDatastreamsStopped(stoppingDatastreamGroups));
+        } catch (RejectedExecutionException ex) {
+          _log.error("handleLeaderDoAssignment: Failed to submit the task that waits for stop propagation");
+          succeeded = false;
+        }
+
       } else {
         _adapter.updateAllAssignments(newAssignmentsByInstance);
-      }
-
-      Set<String> failedStreams = Collections.emptySet();
-      // Poll the zookeeper to ensure that hosts claimed assignment tokens for stopping streams
-      if (_config.getEnableAssignmentTokens() &&
-          !PollUtils.poll(() -> _adapter.getNumUnclaimedTokensForDatastreams(stoppingDatastreamGroups) == 0,
-            STOP_PROPAGATION_RETRY_MS, _config.getStopPropagationTimeout())) {
-        Map<String, List<AssignmentToken>> unclaimedTokens =
-            _adapter.getUnclaimedAssignmentTokensForDatastreams(stoppingDatastreamGroups);
-        failedStreams = unclaimedTokens.keySet();
-        Set<String> hosts = unclaimedTokens.values().stream().flatMap(List::stream).
-            map(AssignmentToken::getIssuedFor).collect(Collectors.toSet());
-        _log.error("Stop failed to propagate within {}ms for streams: {}. The following hosts failed to claim their token(s): {}",
-            _config.getStopPropagationTimeout(), failedStreams, hosts);
-        revokeUnclaimedAssignmentTokens(unclaimedTokens, stoppingDatastreamGroups);
-        _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_FAILED_STOPS, failedStreams.size());
-      }
-
-      // TODO Explore introduction of UNKNOWN/WARN state for streams that failed to stop
-
-      for (DatastreamGroup datastreamGroup : stoppingDatastreamGroups) {
-        for (Datastream datastream : datastreamGroup.getDatastreams()) {
-          // Only streams that were confirmed to have stopped successfully will be transitioned to STOPPED state
-          if (!failedStreams.contains(datastream.getName())) {
-            datastream.setStatus(DatastreamStatus.STOPPED);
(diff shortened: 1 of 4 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1335,39 +1335,16 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
...
-      // TODO Explore introduction of UNKNOWN/WARN state for streams that failed to stop
-
-      for (DatastreamGroup datastreamGroup : stoppingDatastreamGroups) {
-        for (Datastream datastream : datastreamGroup.getDatastreams()) {
-          // Only streams that were confirmed to have stopped successfully will be transitioned to STOPPED state
-          if (!failedStreams.contains(datastream.getName())) {
-            datastream.setStatus(DatastreamStatus.STOPPED);
-            if (!_adapter.updateDatastream(datastream)) {
-              _log.warn("Failed to update datastream: {} to stopped state", datastream.getName());
-              succeeded = false;
```
jzakaryan, line 1367: My main concern with this pull request is that earlier failures to update Zookeeper state would cause the assignment change to be rescheduled.

Now that I moved this code to a separate thread, I can't propagate the result back to handleLeaderDoAssignment to schedule a retry. I'm not sure how valuable this check is. It seems that if Zookeeper failed during a stream update, a retry won't do much help here.

If you think that this retry path is indeed necessary, I can find a workaround to solve this. One way is to introduce a new type of event for STOPPING -> STOPPED state transition. I can schedule that in the main event queue and have it rescheduled until it succeeds.

Share your thoughts @vmaheshw @shrinandthakkar

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1335,39 +1335,16 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
...
-            }
-          }
-        }
+        succeeded = markDatastreamsStopped(stoppingDatastreamGroups, Collections.emptySet());
```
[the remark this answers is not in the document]
jzakaryan: I left the succeeded flag assignment here to not introduce regressions when `enableAssignmentTokens` is disabled (i.e. to make sure the behavior is unchanged)

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1391,14 +1368,55 @@ private void handleLeaderDoAssignment(boolean cleanUpOrphanNodes) {
...
-      _leaderDoAssignmentScheduled.set(true);
-      _leaderDoAssignmentScheduledFuture = _scheduledExecutor.schedule(() -> {
-        _eventQueue.put(CoordinatorEvent.createLeaderDoAssignmentEvent(cleanUpOrphanNodes));
-        _leaderDoAssignmentScheduled.set(false);
-      }, _config.getRetryIntervalMs(), TimeUnit.MILLISECONDS);
+      scheduleLeaderDoAssignmentRetry(cleanUpOrphanNodes);
+    }
+  }
+
+  void scheduleLeaderDoAssignmentRetry(boolean cleanUpOrphanNodes) {
```
vmaheshw, line 1375: this method should be `private`

vmaheshw, line 1375: You changed the variable name `cleanUpOrphanNodes` in #924

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: The retry path is necessary. I like the workaround approach. @shrinandthakkar is working on some changes that will be conflicting with this path.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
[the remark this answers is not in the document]
vmaheshw: (a) If succeed flag is true, then `cleanUpDeadInstanceDataAndOtherUnusedTasks` removes the dead instances from zk.
(b) ` _adapter.updateAllAssignments(newAssignmentsByInstance)` save the newAssignment in the cache and get rid of the older assignment.

Both operations should happen atomically. But, there is a bug in this path introduced because of the datastream stopping transition; if the datastream transition to stopped failed, then only (b) will happen. This will result in the next leader seeing stale assignments and going into a non-recoverable state.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: I will merge those changes when #924 is checked in @vmaheshw

jzakaryan: Renamed the parameter

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
jzakaryan: After assessing the changes to add a new event type for STOPPING -> STOPPED transition, I realized that it is a large change and has risks of introducing regressions.
Instead of doing that I added a retry in the new separate thread that confirms the streams have stopped and updates the Zookeeper. This change won't affect the old behavior/code since it will only trigger when the assignment tokens feature is enabled. This will unblock me and let me test the feature.

jzakaryan: Discussed offline with @vmaheshw. Decided to go forward with retry poll for `STOPPED` → `STOPPING` transition instead of a separate event type. Added improvements.

atoomula (approved):

PR #926, opened 2023-02-28
BMM Restart Improvements: Bug fixes and logging improvements

jzakaryan:
This pull request addresses bugs and makes logging improvements in Assignment Tokens Feature (#919 #921 #922 #924)

* Addressed a bug where a stream with no tasks causes the assignment handler to fail.
* Addressed a bug where for tokens assigned to itself, the leader would print "Claiming token" instead of "Revoking token".
* Addressed a bug which was causing a failure in followers' code for inferring stopping streams.
* Added additional log messages for debugging.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -168,6 +168,9 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   private static final long EVENT_THREAD_SHORT_JOIN_TIMEOUT = 3000L;
   // how long should the leader wait between consecutive polls to zookeeper to confirm that the datastreams have stopped
   private static final long STOP_PROPAGATION_RETRY_MS = 5000L;
+  // how many threads will the token claims executor use for assignment tokens feature. There's a risk that this will get
+  // exhausted when there are more concurrent stop requests than threads in the thread pool
+  private static final int TOKEN_CLAIM_THREAD_POOL_SIZE = 8;
 
   private static final Duration ASSIGNMENT_TIMEOUT = Duration.ofSeconds(90);
 
@@ -268,8 +271,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
     // Initializing executor services
     _scheduledExecutor = Executors.newSingleThreadScheduledExecutor(
         new ThreadFactoryBuilder().setNameFormat("CoordinatorScheduledExecutor-%d").build());
-    // TODO Assess whether having a single threaded executor for token claim tasks is sufficient or it will be exhausted
-    _tokenClaimExecutor = Executors.newSingleThreadExecutor(
+    _tokenClaimExecutor = Executors.newFixedThreadPool(TOKEN_CLAIM_THREAD_POOL_SIZE,
         new ThreadFactoryBuilder().setNameFormat("CoordinatorTokenClaimExecutor-%d").build());
 
     for (String connectorType : _connectors.keySet()) {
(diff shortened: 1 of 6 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -268,8 +271,7 @@ public void start() {
     // Initializing executor services
     _scheduledExecutor = Executors.newSingleThreadScheduledExecutor(
         new ThreadFactoryBuilder().setNameFormat("CoordinatorScheduledExecutor-%d").build());
-    // TODO Assess whether having a single threaded executor for token claim tasks is sufficient or it will be exhausted
-    _tokenClaimExecutor = Executors.newSingleThreadExecutor(
+    _tokenClaimExecutor = Executors.newFixedThreadPool(TOKEN_CLAIM_THREAD_POOL_SIZE,
```
jzakaryan, line 274: A single threaded token claim executor will get backed up in the following cases:
* The leader will try to schedule 2 tasks in this executor upon a stream stop: (1) a task for polling the Zookeeper to check whether stop has propagated and completed across the cluster (2) a task for claiming the token assigned to itself.
* Concurrent requests to stop different streams.

A single threaded executor won't work here. For now I increased the thread pool size to 8 threads. That will let the cluster process up to 8 concurrent requests without being backed up (note that token claim tasks are short lived unlike the poll tasks, so they won't be the bottleneck here). I'll monitor how this performs in EI and make adjustments accordingly. A config property can also be added in the future if we end up tinkering with this.

shrinandthakkar: great catch!

shrinandthakkar, line 274: I think it's debatable what should be the ideal value for the pool size here. 

Only in the worst case when 8 concurrent stop requests occupy the thread pool of the leader and tokens assigned to itself are never scheduled for claiming, resulting in a deadlock situation. But this sounds very rare unless we do scripted restarts of datastreams in clusters.

jzakaryan: It's not even a deadlock situation. What happens is the executor gets backed up, and the task to claim tokens has to wait for threads to be available in order to be executed. So all tasks eventually get executed.
In case of a single threaded executor the token claim task will be executed after the poll task times out. So by the time it tries to claim the tokens leader already revokes them and declares a stop failure.

jzakaryan: As for the thread pool side, I'd rather be on the safer side, that's why I put 8.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -834,27 +842,40 @@ void maybeClaimAssignmentTokensForStoppingStreams(List<DatastreamTask> newAssign
...
+        stoppingStreamNames = stoppingStreams.stream().map(Datastream::getName).collect(Collectors.toSet());
+      } catch (Exception ex) {
+        _log.error("Failed to infer stopping streams. ", ex);
+      }
 
       if (!stoppingStreamNames.isEmpty()) {
+        _metrics.updateMeter(CoordinatorMetrics.Meter.NUM_INFERRED_STOPPING_STREAMS, stoppingStreams.size());
         _log.info("Trying to claim assignment tokens for connector {}, streams: {}", connector, stoppingStreamNames);
 
+        Set<String> finalStoppingStreamNames = stoppingStreamNames;
```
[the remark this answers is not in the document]
jzakaryan: Because when using it in a lambda, it has to be final or effectively final.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -735,6 +737,12 @@ private void handleAssignmentChange(boolean isDatastreamUpdate) throws TimeoutEx
     if (_config.getEnableAssignmentTokens()) {
       try {
         // Queue assignment token claim task
+        List<String> oldAssignmentTaskNames = oldAssignment.stream().map(DatastreamTask::getDatastreamTaskName).
+            collect(Collectors.toList());
+        List<String> newAssignmentTaskNames = newAssignment.stream().map(DatastreamTask::getDatastreamTaskName).
+            collect(Collectors.toList());
+        _log.debug("Claiming assignment tokens. Old assignment: {}", oldAssignmentTaskNames);
```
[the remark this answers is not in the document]
jzakaryan: I just wanted these printed before scheduling the task for claiming tokens. These logs helped me figure out the bug with the dying tokens claim thread. I'm also thinking to clean these up once the feature is tested more and rolled out in production.

vmaheshw (approved):

shrinandthakkar (approved):

PR #928, opened 2023-03-16
Handling Metrics and SLA Reporting for Throughput Violating Topics via Datastream Update API Part 1

shrinandthakkar:
The EventProducer of every DatastreamTask reports SLA and latency metrics for every datastream record. But when topics (at least one partition) have higher throughput than the thresholds, it introduces latency and SLA misses in the mirroring pipeline.

This pull request is the first part of changes to handle the metrics and SLA reporting of throughput-violating topics via the datastream update API. It introduces the following changes:

1. The datastream update endpoint shall simultaneously accept a datastream and a list of throughput-violating topics via its datastream metadata. The ZkAdapter would persist this information in the DatastreamStore.
2. The update API touches every server host in its normal code path, and in that code path, every host (Coordinator) maintains a shared cache (Datastream -> Violating Topics Map) with the latest violations for every datastream.
3. A new callback for the eventProducer to fetch the latest list of offending topics from the Coordinator. We ensure that the correct set of topics is excluded from reporting the metrics and SLAs for every record.

***
Part 2 of this series would take care of:
1. Excluding the reporting of metrics for these throughput-violating topics within EventProducer.
2. Introducing separate metrics for throughput-violating topics.

***

the contribution guidelines.

### DIFF
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
@@ -38,7 +38,8 @@ import static com.linkedin.datastream.server.Coordinator.PAUSED_INSTANCE;
 public class ZookeeperBackedDatastreamStore implements DatastreamStore {
 
   private static final Logger LOG = LoggerFactory.getLogger(ZookeeperBackedDatastreamStore.class.getName());
-
+  private static final Double ONE_MEBIBYTE = (double) (1024 * 1024);
+  private static final Double ZNODE_BLOB_SIZE_LIMIT = ONE_MEBIBYTE;
   private final ZkClient _zkClient;
   private final String _cluster;
   private final CachedDatastreamReader _datastreamCache;
@@ -127,6 +128,12 @@ public class ZookeeperBackedDatastreamStore implements DatastreamStore {
       throw new DatastreamException("Datastream does not exists, can not be updated: " + key);
     }
 
+    // As this limit is ZK specific, adding this validation check specifically in ZookeeperBackedDatastreamStore.
+    double datastreamBlobSizeInMBs = getBlobSizeInMBs(DatastreamUtils.toJSON(datastream));
+    Validate.isTrue(datastreamBlobSizeInMBs <= ZNODE_BLOB_SIZE_LIMIT,
+        String.format("Encoded blob size: %.2f exceeded the znode threshold of %.2f. Datastream: %s cannot be updated.",
+            datastreamBlobSizeInMBs, ZNODE_BLOB_SIZE_LIMIT, datastream.getName()));
+
     Objects.requireNonNull(datastream.getMetadata()).remove("numTasks");
     String json = DatastreamUtils.toJSON(datastream);
     _zkClient.writeData(getZnodePath(key), json);
@@ -139,6 +146,12 @@ public class ZookeeperBackedDatastreamStore implements DatastreamStore {
   public void createDatastream(String key, Datastream datastream) {
     Validate.notNull(datastream, "null datastream");
     Validate.notNull(key, "null key for datastream" + datastream);
+
+    // As this limit is ZK specific, adding this validation check specifically in ZookeeperBackedDatastreamStore.
+    double datastreamBlobSizeInMBs = getBlobSizeInMBs(DatastreamUtils.toJSON(datastream));
+    Validate.isTrue(datastreamBlobSizeInMBs <= ZNODE_BLOB_SIZE_LIMIT,
+        String.format("Encoded blob size: %.2f exceeded the znode threshold of %.2f. Datastream: %s cannot be created.",
+            datastreamBlobSizeInMBs, ZNODE_BLOB_SIZE_LIMIT, datastream.getName()));
 
     String path = getZnodePath(key);
     if (_zkClient.exists(path)) {
@@ -263,5 +276,13 @@ public class ZookeeperBackedDatastreamStore implements DatastreamStore {
     String dmsPath = KeyBuilder.datastreams(_cluster);
     // Update the /dms to notify that coordinator needs to act on a deleted or changed datastream.
     _zkClient.updateDataSerialized(dmsPath, old -> String.valueOf(System.currentTimeMillis()));
+  }
+
+  // Gets the size of the encoded blob (default charset UTF) in MBs
+  private double getBlobSizeInMBs(String stringEncodedBlob) {
+    if (Objects.nonNull(stringEncodedBlob) && !stringEncodedBlob.isEmpty()) {
+      return ((double) stringEncodedBlob.getBytes(ZkClient.ZK_STRING_SERIALIZER_CHARSET).length) / ONE_MEBIBYTE;
+    }
+    return 0;
   }
 }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -31,6 +31,9 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -223,6 +226,15 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   private Future<?> _leaderDoAssignmentScheduledFuture = null;
   private volatile boolean _zkSessionExpired = false;
 
+  // Cache all the throughput violating topics per datastream
+  private final Map<String, Set<String>> _throughputViolatingTopicsMap = new HashMap<>();
+  private final Function<DatastreamTask, Set<String>> _throughputViolatingTopicsProvider;
+
+  // As the _throughputViolatingTopicsMap may be updated by watcher thread and read by multiple task threads concurrently.
+  private final ReadWriteLock _throughputViolatingTopicsMapReadWriteLock = new ReentrantReadWriteLock();
+  private final Lock _throughputViolatingTopicsMapWriteLock = _throughputViolatingTopicsMapReadWriteLock.writeLock();
+  private final Lock _throughputViolatingTopicsMapReadLock = _throughputViolatingTopicsMapReadWriteLock.readLock();
+
   /**
    * Constructor for coordinator
    * @param datastreamCache Cache to maintain all the datastreams in the cluster.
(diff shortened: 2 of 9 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -952,7 +1031,15 @@ private void initializeTask(DatastreamTask task, Set<DatastreamTask> failedDatas
...
       TransportProviderAdmin tpAdmin = _transportProviderAdmins.get(task.getTransportProviderName());
       TransportProvider transportProvider = tpAdmin.assignTransportProvider(task);
-      EventProducer producer = new EventProducer(task, transportProvider, _cpProvider, _eventProducerConfig, customCheckpointing);
+
+      Function<DatastreamTask, Set<String>> throughputViolatingTopicsProvider =
+          _config.getEnableThroughputViolatingTopicsHandling() ?
+              (t) -> getThroughputViolatingTopics(t.getDatastreams()) :
+              (t) -> new HashSet<>();
+
+      EventProducer producer =
```
shrinandthakkar, line 1040: From @vmaheshw 
```
Will it make sense to simplify this at Datastream level rather than DatastreamTask level?
For eg: in digest, there is only one datastream. This function will result in 10 copies for 
the same datastream.

Also, hashing based on datastream name will be cheaper and more efficient.
```

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -223,6 +226,14 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
...
   private volatile boolean _zkSessionExpired = false;
 
+  // Cache all the throughput violating topics per datastream
+  private final Map<String, Set<String>> _throughputViolatingTopicsMap = new HashMap<>();
+
+  // As the _throughputViolatingTopicsMap may be updated by watcher thread and read by multiple task threads concurrently.
+  private final ReadWriteLock _throughputViolatingTopicsMapReadWriteLock = new ReentrantReadWriteLock();
+  private final Lock _throughputViolatingTopicsMapWriteLock = _throughputViolatingTopicsMapReadWriteLock.writeLock();
+  private final Lock _throughputViolatingTopicsMapReadLock = _throughputViolatingTopicsMapReadWriteLock.readLock();
+
```
shrinandthakkar, line 236: From @vmaheshw 
```
Do we need to go to low-level locks for this? Can we not rely on ConcurrentHashMap? 
Violation calculation does not have to be precise in the close second.
```

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -466,10 +478,60 @@ public void onDatastreamUpdate() {
...
+            "received an exception while populating the datastream violating topics to host level cache from the "
+                + "datastream objects on the update trigger", exception);
+      }
+    }
     queueHandleAssignmentOrDatastreamChangeEvent(CoordinatorEvent.createHandleDatastreamChangeEvent(), true);
     _log.info("Coordinator::onDatastreamUpdate completed successfully");
   }
 
+  // populates the violations to local cache from the datastream metadata on every update call.
+  private void populateThroughputViolatingTopicsMap(List<DatastreamGroup> datastreamGroups) {
```
shrinandthakkar, line 498: From @vmaheshw 
```
Also, mention that the behavior is replace all, rather than incremental update.
```

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: Instead of moving this at the Datastream level, I refactored to only maintaining a single initialization of this callback per instance and only maintaining the set of violations per init of EventProducer.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: I would prefer not to use the concurrent hash map since while performing "replace-all" to this map, there may be chances that these violating topics' metrics and SLAs get reported. Hence to keep things consistent, I went with this approach.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: done

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -38,7 +38,8 @@
 public class ZookeeperBackedDatastreamStore implements DatastreamStore {
 
   private static final Logger LOG = LoggerFactory.getLogger(ZookeeperBackedDatastreamStore.class.getName());
-
+  private static final Double ONE_MEBIBYTE = (double) (1024 * 1024);
+  private static final Double ZNODE_BLOB_SIZE_LIMIT = ONE_MEBIBYTE;
```
vmaheshw, line 42: Do you really need 2 variables?

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
```diff
@@ -127,6 +128,12 @@ public void updateDatastream(String key, Datastream datastream, boolean notifyLe
       throw new DatastreamException("Datastream does not exists, can not be updated: " + key);
     }
 
+    // As this limit is ZK specific, adding this validation check specifically in ZookeeperBackedDatastreamStore.
+    double datastreamBlobSizeInMBs = getBlobSizeInMBs(DatastreamUtils.toJSON(datastream));
+    Validate.isTrue(datastreamBlobSizeInMBs <= ZNODE_BLOB_SIZE_LIMIT,
```
vmaheshw, line 133: The validation check is tricky, especially in the case of a Programmatic update. This will block the datastream update until the logic in the caller is fixed. Can you think of a less disruptive way?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
vmaheshw: I recommend simplicity, especially since this is not on a critical path.

vmaheshw (commented):
I am okay with the READ-WRITE lock if the other approver is also fine.

Please address the valid check comment.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
jzakaryan: What are your concerns @vmaheshw? I understand that we need to add similar validation on the client side as well. If a client request is breaching the ZK node size limit, it will eventually fail the update anyway, it's just the failure will happen in the ZkAdapter/ZkClient layer. This to me is a more explicit and descriptive way to fail.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -223,6 +226,15 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   private Future<?> _leaderDoAssignmentScheduledFuture = null;
   private volatile boolean _zkSessionExpired = false;
 
+  // Cache all the throughput violating topics per datastream
```
jzakaryan, line 229: Can you please confirm that no 2 streams can have the same name, even across different connectors? If that's not the case might be worth storing Connector + Stream Name tuple as the key here.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -466,10 +484,61 @@ public void onDatastreamUpdate() {
...
     }
+
+    if (_config.getEnableThroughputViolatingTopicsHandling()) {
+      _log.info(
+          "Populating the datastream violating topics to host level cache from the datastream objects on the update trigger");
+      try {
+        populateThroughputViolatingTopicsMap(datastreamGroups);
+      } catch (Exception exception) {
+        _log.error(
+            "received an exception while populating the datastream violating topics to host level cache from the "
```
jzakaryan, line 495: Nit. Start message with capital letter like everywhere else

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -466,10 +484,61 @@ public void onDatastreamUpdate() {
...
+    } finally {
+      _throughputViolatingTopicsMapWriteLock.unlock();
+    }
+  }
+
+  /**
+   * Fetch all the throughput violating topics per datastream from cache.
+   * */
+  public Set<String> getThroughputViolatingTopics(List<Datastream> datastreams) {
+    _throughputViolatingTopicsMapReadLock.lock();
```
jzakaryan, line 531: Does this need to be public? If you don't plan to expose this other than for tests, consider making it package private.

jzakaryan (changes requested):
Left a few comments. Overall looks good.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
vmaheshw: I initially thought that it can impact the datastream restart, but I was wrong.

However, in the scenario, the limit goes >1MB because of the violation list, we will not be able to disable this at the server level and will have to rely on the external service to disable this feature. Until then, any other update/allowlisting will not go through.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
shrinandthakkar: The purpose of these variables are little different
- ZNODE_BLOB_SIZE_LIMIT ––> Validation of data size per znode based on the znode limit (=1MB) 
- ONE_MEBIBYTE  --> Converting bytes to MB to get encoded data size in MBs. 

Ideally, I can use the same variable, but I kept them both for better understanding purposes of two different operations happening on the same value.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: I think we can safely assume that datastream names are unique across a cluster. I am using a map with the key as the datastream name here since we'll be storing this information within ZK similarly under the /DMS/ path with datastream as the key.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: done

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
shrinandthakkar: done

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/ZookeeperBackedDatastreamStore.java
shrinandthakkar: @vmaheshw understood your concern. 

I have added steps in both update and create paths to only adhere to the throughputViolatingTopic metadata if the corresponding config is enabled in our server.

jzakaryan (approved):
Going forward, please retain the commit history in your PRs. Don't rebase/squash those commits locally. Having a commit history retained in the PR helps reviewers see changes over time.

vmaheshw (approved):

PR #931, opened 2023-04-11
Added Support and Handling For Generic Headers

shrinandthakkar:
## Summary
BrooklinEnvelope supports transporting the source header data to the destination, but is currently bounded to Kafka's header type. This change replaces that with generic header to let any other publisher-subscriber transfer events & headers via brooklin. 

***
## Testing
Added a unit test comparing the source and destination event's headers.

### DIFF
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
@@ -11,7 +11,6 @@ import java.util.Optional;
 
 import org.apache.avro.reflect.Nullable;
 import org.apache.commons.lang.Validate;
-import org.apache.kafka.common.header.Headers;
 
 
 /**
@@ -30,7 +29,7 @@ public class BrooklinEnvelope {
 
   private Map<String, String> _metadata;
 
-  private Headers _headers;
+  private Object _headers;
 
   /**
    * Construct a BrooklinEnvelope using record key, value, and metadata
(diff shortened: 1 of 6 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
```diff
@@ -112,7 +112,7 @@ private ProducerRecord<byte[], byte[]> convertToProducerRecord(String topicName,
     Headers headers = null;
     if (event instanceof BrooklinEnvelope) {
       BrooklinEnvelope envelope = (BrooklinEnvelope) event;
-      headers = envelope.getHeaders();
+      headers = envelope.getHeaders() instanceof Headers ? (Headers) envelope.getHeaders() : null;
```
jzakaryan, line 115: Do we expect a non Kafka Headers object here? If no, is it worth throwing an exception or logging an error here?

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
```diff
@@ -11,7 +11,6 @@
 
 import org.apache.avro.reflect.Nullable;
 import org.apache.commons.lang.Validate;
-import org.apache.kafka.common.header.Headers;
```
jzakaryan, line 14: Now that this import is removed from here, can you check if the Kafka dependency can be removed from `datastream-common` module? IIRC Headers was the only dependency.

jzakaryan (changes requested):

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaTransportProvider.java
shrinandthakkar: logging would be costly as it would do it for every record, I instead added an exception path which will fail in moving further if something other than kafka type is encountered.

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
shrinandthakkar: removed, but had to add it to datastream-utils since it was transitively getting kafka via datastream common haha

datastream-common/src/main/java/com/linkedin/datastream/common/BrooklinEnvelope.java
```diff
@@ -95,14 +94,14 @@ public void setPreviousValue(Object previousValue) {
   /**
    * Get the Kafka headers
    */
```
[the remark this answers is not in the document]
shrinandthakkar: nice catch! fixed

ehoner (approved):

jzakaryan (approved):

PR #932, opened 2023-04-12
Minor improvements w.r.t #928 & Releasing a newer version

shrinandthakkar:
Added a new constructor of EventProducer to better handle dependencies for non MM use cases. 
***
Also releasing a newer version.
***

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
@@ -10,7 +10,6 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


gradle/maven.gradle
```diff
@@ -1,5 +1,5 @@
 allprojects {
-  version = "5.2.8-SNAPSHOT"
+  version = "5.3.0"
```
jzakaryan, line 2: Keep `SNAPSHOT` in the version number. That prevents developers from accidentally publishing a version.

vmaheshw (dismissed):

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -122,6 +123,20 @@ public class EventProducer implements DatastreamEventProducer {
    * @param customCheckpointing decides whether producer should use custom checkpointing or the datastream server
    *                            provided checkpointing.
    */
+  public EventProducer(DatastreamTask task, TransportProvider transportProvider, CheckpointProvider checkpointProvider,
+      Properties config, boolean customCheckpointing) {
+    this(task, transportProvider, checkpointProvider, config, customCheckpointing, (t) -> new HashSet<>());
```
jzakaryan, line 128: Nit: Consider using `Collections.emptySet()` here when initializing your empty set. This is more intentional because it creates an immutable empty set.

gradle/maven.gradle
shrinandthakkar: Although once a version is published, it could not be overridden afaik but we can keep SNAPSHOT to be on safer side.

jzakaryan (approved):

PR #937, opened 2023-05-23
Add metrics for CoordinatorEvent queue

ehoner:
Adds a `Gauge` and `Counter` to `CoordinatorEventBlockingQueue`. The `Gauge` tracks the current size of the underlying queue; and the `Counter` increments when a duplicate event is `put()`. The class implements `MetricsAware` and registers the underlying metrics to the simple class name + constructor parameter `key`. The existing parameterless constructor is marked deprecated. It registers metrics to the key "CoordinatorEventBlockingQueue.legacy".

The change to `MetricTestUtils` was required to avoid collisions with `TestCoordinator`. The `startsWith` predicate was changed to `match` to ensure that the entire name was used to filter. The same metric verification registration test was added to the `CoordinatorEventBlockingQueue`.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
@@ -39,7 +39,8 @@ public class CoordinatorEventBlockingQueue implements MetricsAware {
 
   private static final Logger LOG = LoggerFactory.getLogger(CoordinatorEventBlockingQueue.class.getName());
   private static final String SIMPLE_NAME = CoordinatorEventBlockingQueue.class.getSimpleName();
-  private static final Set<BrooklinMetricInfo> _metricInfos = ConcurrentHashMap.newKeySet();
+  private static final Set<BrooklinMetricInfo> METRIC_INFOS = ConcurrentHashMap.newKeySet();
+
   static final String COUNTER_KEY = "duplicateEvents";
   static final String GAUGE_KEY = "queuedEvents";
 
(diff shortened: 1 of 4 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
```diff
@@ -5,47 +5,101 @@
...
+  private final DynamicMetricsManager _dynamicMetricsManager;
+  private final Gauge<Integer> _gauge;
+  private final Counter _counter;
 
   /**
    * Construct a blocking event queue for all types of events in {@link CoordinatorEvent.EventType}
+   *
+   * @deprecated Use of this constructor should be replaced with the single parameter variant.
    */
+  @Deprecated
```
jzakaryan, line 57: I wouldn't bother adding this, since `CoordinatorEventBlockingQueue` is only initialized in the datastream-server module and you already integrated this change in `Coordinator`. Doesn't hurt to have it though.

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
```diff
@@ -5,47 +5,101 @@
...
+ *
  * @see CoordinatorEvent.EventType
  */
-public class CoordinatorEventBlockingQueue {
+public class CoordinatorEventBlockingQueue implements MetricsAware {
 
   private static final Logger LOG = LoggerFactory.getLogger(CoordinatorEventBlockingQueue.class.getName());
+  private static final Set<BrooklinMetricInfo> METRIC_INFOS = ConcurrentHashMap.newKeySet();
+
+  static final String COUNTER_KEY = "duplicateEvents";
```
jzakaryan, line 43: I don't see a lot of practical value in measuring duplicateEvents with a counter.

jzakaryan (dismissed):
Looks good to me

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
jzakaryan: Yes, I think that makes more sense (i.e. removing the public access modifier from the queue class)

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorEventBlockingQueue.java
jzakaryan: Let's leave it. Having an extra metric won't hurt.

jzakaryan (approved):

shrinandthakkar (approved):

PR #939, opened 2023-05-25
Fix tests and reduce the overall test execution time

vmaheshw:
1. Fixed the tests that are taking longer time for execution.
2. Added some of the default Kafka configs values, to reduce the overall delay in the tests.
3.  Added checks around pollUtils to ensure that it passes at the end of it.

CircleCI result: 24m 17s (reduced by 25%)

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -136,6 +136,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected final GroupIdConstructor _groupIdConstructor;
 
   protected final KafkaTopicPartitionTracker _kafkaTopicPartitionTracker;
+  private long _commitRetryTimeoutMillis;
 
   protected AbstractKafkaBasedConnectorTask(KafkaBasedConnectorConfig config, DatastreamTask task, Logger logger,
       String metricsPrefix, GroupIdConstructor groupIdConstructor) {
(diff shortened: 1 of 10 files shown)


datastream-kafka/src/main/java/com/linkedin/datastream/kafka/EmbeddedKafkaCluster.java
```diff
@@ -114,6 +114,7 @@ public void startup() {
       properties.setProperty("log.flush.interval.messages", String.valueOf(1));
       properties.setProperty("log.cleaner.enable", Boolean.FALSE.toString()); //to save memory
       properties.setProperty("offsets.topic.num.partitions", "1");
+      properties.setProperty("offsets.topic.replication.factor", "1");
```
surajkn, line 117: Just to understand correctly, this reduced the number of replicas to be maintained to 1 vs the default (which I think is 3)?
Do you know how much time this saves? Curious to know how you identified this to be causing delay

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -686,7 +688,7 @@ protected void commitWithRetries(Consumer<?, ?> consumer, Optional<Map<TopicPart
       }
 
       return true;
-    }, COMMIT_RETRY_INTERVAL_MILLIS, COMMIT_RETRY_TIMEOUT_MILLIS);
+    }, COMMIT_RETRY_INTERVAL_MILLIS, getCommitRetryTimeoutMillis());
```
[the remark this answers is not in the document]
vmaheshw: Some of the tests are written for it to fail. Bypassing the timeout to a shorter duration helps in reducing the time.

datastream-kafka/src/main/java/com/linkedin/datastream/kafka/EmbeddedKafkaCluster.java
vmaheshw: Some configurations are not setting enough brokers for this replication to work. The gain is not enough, but solves unnecessary errors.

datastream-kafka-connector/src/test/java/com/linkedin/datastream/connectors/kafka/mirrormaker/TestKafkaMirrorMakerConnectorTask.java
```diff
@@ -91,6 +91,7 @@
 /**
  * Tests for {@link KafkaMirrorMakerConnectorTask}
  */
+@Test(singleThreaded = false)
```
surajkn, line 94: Should we be doing this in all our test classes? 
I thought by default each test was executed in its own thread and run in parallel. Is that not the case ?

surajkn (approved):

jzakaryan (approved):

PR #943, opened 2023-06-02
[wip] Assignment refactor - performance optimization

ehoner:
The changes here collapse multiple loops into a single traversal. This is an experimental change intended for review and discussion. These changes are incomplete and should be discussed before continuing. 

### Technical Overview
There are 4 requirements to the traversal of `currentAssignment` affected:

1. Group partitions (`Set<String>`) by `taskId`: `newPartitionAssignmentMap`.
2. Determine which tasks had a change in assignment count: `tasksWithChangedPartition`.
3. Find the total throughput for each task: `taskThroughputMap`.
4. Identify which tasks were below the max partition count: `tasks`.

The prototype proposed performs all 4 in a single traversal. The current implementation does 3. The first for 1 and 2. While 3 and 4 each have their own traversal. It isn't clear if the use of `parallel()` will affect performance.


ehoner:
Can be reopened later.

PR #944, opened 2023-06-13
Add stop method to TransportProviderAdmin

vmaheshw:
Some TransportProviderAdmins create AdminClient for Kafka and the lifecycle of the object is until the coordinator stops. Without a proper stop call for TransportProviderAdmin, the AdminClient loses connection to the other entities and result in failure for other unit-tests.


jzakaryan (approved):
LGTM. Just curious, are you going to override the stop method for one of the "fake" transport provider admins used in tests?

shrinandthakkar (approved):

vmaheshw:
> LGTM. Just curious, are you going to override the stop method for one of the "fake" transport provider admins used in tests?

This is a problem for the Kafka Transport provider admins in the test and I will see what is the best way to fix the tests. In production, we don't need it as the JVM itself gets shutdown and there is no dangling resource, but for unit-tests, the dangling resource creates problem for other unit-test.

PR #948, opened 2023-06-30
Log existing datastream for update and delete call

vmaheshw:
Existing datastream object needs to be logged to check the states and rollback based on logs if necessary.


datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -245,6 +245,8 @@ private void doUpdateDatastreams(Map<String, Datastream> datastreamMap) {
           throw new DatastreamValidationException(String.format("Failed to update %s. Can't update numTasks."
               + " old: %s new: %s", key, oldDatastream, datastream));
         }
+
+        LOG.info("[UPDATE] old datastream: {}, new datastream: {}", oldDatastream, datastream);
```
hshukla, line 249: Do we need to log the whole datastream object? Here have mentioned [UPDATE] but for delete the log says `Delete`, can we be consistent?

Also, can we please follow key={value} format when we log, as some logging tool does not work great with `:`

Same for other log like in delete

ehoner (approved):

hshukla (approved):

PR #951, opened 2023-08-03
Added log messages to debug assignPartitions

jzakaryan:
We observed that the leader has a tendency to get stuck on this method (in some rare cases for more than 5 minutes). Added log messages to help debug the issue.

Note that the test suite in `TestLoadBasedPartitionAssigner` do not capture such performance problems. Attempts to reproduce them locally didn't work.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
@@ -80,10 +80,12 @@ public class LoadBasedPartitionAssigner implements MetricsAware {
   public Map<String, Set<DatastreamTask>> assignPartitions(
       ClusterThroughputInfo throughputInfo, Map<String, Set<DatastreamTask>> currentAssignment,
       List<String> unassignedPartitions, DatastreamGroupPartitionsMetadata partitionMetadata, int maxPartitionsPerTask) {
+    LOG.info("START: assignPartitions");
     String datastreamGroupName = partitionMetadata.getDatastreamGroup().getName();
     Map<String, PartitionThroughputInfo> partitionInfoMap = new HashMap<>(throughputInfo.getPartitionInfoMap());
     Set<String> tasksWithChangedPartition = new HashSet<>();
 
+    LOG.info("Filtering out tasks for the current datastream {}", datastreamGroupName);
     // filter out all the tasks for the current datastream group, and retain assignments in a map
     Map<String, Set<String>> newPartitionAssignmentMap = new HashMap<>();
     currentAssignment.values().forEach(tasks ->
@@ -103,6 +105,7 @@ public class LoadBasedPartitionAssigner implements MetricsAware {
     validatePartitionCountAndThrow(datastreamGroupName, numTasks, numPartitions, maxPartitionsPerTask);
 
     // sort the current assignment's tasks on total throughput
+    LOG.info("Extracting throughput info for partitions which have the throughput data (recognized partitions)");
     Map<String, Integer> taskThroughputMap = new HashMap<>();
     PartitionThroughputInfo defaultPartitionInfo = new PartitionThroughputInfo(_defaultPartitionBytesInKBRate,
         _defaultPartitionMsgsInRate, "");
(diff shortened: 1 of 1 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -80,10 +80,12 @@ public LoadBasedPartitionAssigner(int defaultPartitionBytesInKBRate, int default
   public Map<String, Set<DatastreamTask>> assignPartitions(
       ClusterThroughputInfo throughputInfo, Map<String, Set<DatastreamTask>> currentAssignment,
       List<String> unassignedPartitions, DatastreamGroupPartitionsMetadata partitionMetadata, int maxPartitionsPerTask) {
+    LOG.info("START: assignPartitions");
```
hshukla, line 83: do we need a datastream name here?

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -80,10 +80,12 @@ public LoadBasedPartitionAssigner(int defaultPartitionBytesInKBRate, int default
...
     String datastreamGroupName = partitionMetadata.getDatastreamGroup().getName();
     Map<String, PartitionThroughputInfo> partitionInfoMap = new HashMap<>(throughputInfo.getPartitionInfoMap());
     Set<String> tasksWithChangedPartition = new HashSet<>();
 
+    LOG.info("Filtering out tasks for the current datastream {}", datastreamGroupName);
```
hshukla, line 88: I would encourage to use key=value format for the log message, that way it is easier to search on log analysis tool.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
```diff
@@ -137,13 +140,15 @@ public Map<String, Set<DatastreamTask>> assignPartitions(
       }
     }
 
+    LOG.info("Sorting recognized partitions on byte rate");
```
hshukla, line 143: any other metadata needed to know what datastream this log belongs too? by looking into this log line one might not have enough information to interpret what this is for. Or the above log which has datastream name is sufficient for us? https://github.com/linkedin/brooklin/pull/951/files#diff-430d9ed3200fdf27009cd18273178da8a789cdec2e410b90ac5bb00562704d69R88

ehoner (approved):
I am approving these in case they are time-sensitive. But in general I agree with @hshukla's comments. I am not convinced these changes will surface _the_ issue(s), but this is a worthwhile first step before considering more involved changes.

shrinandthakkar:
I remember us talking about putting logs around any IO operations that might be happening, should we add some logs there as well?
- [Throughput Provider Interactions.](https://github.com/linkedin/brooklin/blob/da925e6dcf9d454c52b3f72ed69cf48ca683f1eb/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java#L127)
- ZK Interactions if any.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
jzakaryan: Added the data stream name to the START and END markers. This will be sufficient to trace these in the logs, since coordinator is running in a single thread and invocations of this method won't run concurrently.

datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
jzakaryan: Datastream name in the START and END messages is sufficient. No need to repeat it in every message printed.

jzakaryan:
> I remember us talking about putting logs around any IO operations that might be happening, should we add some logs there as well?
> 
> * [Throughput Provider Interactions.](https://github.com/linkedin/brooklin/blob/da925e6dcf9d454c52b3f72ed69cf48ca683f1eb/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssignmentStrategy.java#L127)
> * ZK Interactions if any.

The consensus at the end of the meeting was that the assignPartitions method is where the CPU time is spent. I avoided adding logs in other places as to keep unnecessary logging to minimum.

hshukla (approved):

shrinandthakkar (approved):

PR #957, opened 2023-08-28
Added a timeout on the producer flush call in KafkaMirrorMakerConnecorTask

jzakaryan:
Even in flushness mode BMM's tasks do a flush call on the producer when shutting down. We have observed that producer flush call tends to get indefinitely stuck and this keeps the tasks from shutting down gracefully. The code change in this PR addresses this by wrapping the producer flush call in a future and blocking on that future with a timeout.

If the producer flush doesn't complete in the given timeout window, the task will proceed to committing safe offsets and shutting down. The timeout window is exposed through a configuration property.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
@@ -16,6 +16,9 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
@@ -124,6 +127,7 @@ public class KafkaMirrorMakerConnectorTask extends AbstractKafkaBasedConnectorTa
   // among Kafka consumer client metrics for different datastreams.
   private final boolean _includeDatastreamNameInConsumerClientId;
   private final String _destinationTopicPrefix;
+  private final long _hardCommitFlushTimeoutMs;
   private FlushlessEventProducerHandler<Long> _flushlessProducer = null;
   private boolean _flowControlEnabled = false;
   private long _maxInFlightMessagesThreshold;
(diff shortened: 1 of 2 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/mirrormaker/KafkaMirrorMakerConnectorTask.java
```diff
@@ -406,7 +411,11 @@ protected void maybeCommitOffsets(Consumer<?, ?> consumer, boolean hardCommit) {
...
         try {
-          _datastreamTask.getEventProducer().flush();
+          Future<?> producerFlushFuture = Executors.newSingleThreadExecutor().
+              submit(() -> _datastreamTask.getEventProducer().flush());
+          producerFlushFuture.get(_hardCommitFlushTimeoutMs, TimeUnit.MILLISECONDS);
+        } catch (Exception ex) {
+          LOG.warn("Producer flush failed with exception: ", ex);
         } finally {
           // Flushless mode tracks the successfully received acks, so it is safe to commit offsets even if flush throws
           // an exception. Commit the safe offsets to reduce send duplication.
```
ehoner, line 421: Should the executor be shutdown? Alternatively, maybe moved to the class level.

```java
ExecutorService executor = Executors.newSingleThreadExecutor();
try {
  executor.submit(() -> _datastreamTask.getEventProducer().flush())
      .get(_hardCommitFlushTimeoutMs, TimeUnit.MILLISECONDS);
} catch (Exception ex) {
  LOG.warn("Producer flush failed with exception: ", ex);
} finally {
  commitSafeOffsets(consumer);
  executor.shutdownNow();
}
```

shrinandthakkar (commented):
why shouldn't we just use the producer config `offset.flush.timeout.ms` instead of creating a newer config to wait on ? 

ref: https://kafka.apache.org/21/documentation.html#producerconfigs

jzakaryan:
> why shouldn't we just use the producer config `offset.flush.timeout.ms` instead of creating a newer config to wait on ?
> 
> ref: https://kafka.apache.org/21/documentation.html#producerconfigs

@shrinandthakkar the problem is that the the tasks were found to be stuck on `flush` after 60 seconds despite the default value of 5 seconds for `offset.flush.timeout.ms`. Connector logs show that it interrupted the producer.flush to shut down the task.
We can reach out to Kafka team and ask if the behavior of LKC wrt producer flush and config values are the same as in Apache Kafka. But that can happen in parallel with us addressing it.

shrinandthakkar:
> > why shouldn't we just use the producer config `offset.flush.timeout.ms` instead of creating a newer config to wait on ?
> > ref: https://kafka.apache.org/21/documentation.html#producerconfigs
> 
> @shrinandthakkar the problem is that the the tasks were found to be stuck on `flush` after 60 seconds despite the default value of 5 seconds for `offset.flush.timeout.ms`. Connector logs show that it interrupted the producer.flush to shut down the task. We can reach out to Kafka team and ask if the behavior of LKC wrt producer flush and config values are the same as in Apache Kafka. But that can happen in parallel with us addressing it.

@jzakaryan 
Within the EventProducer's flush call, I think we already have a configuration defined ([ref](https://github.com/linkedin/brooklin/blob/a3d26a8bbe935d9c6173a96e2a6b786ebf1fb857/datastream-kafka/src/main/java/com/linkedin/datastream/kafka/KafkaProducerWrapper.java#L399)). And the value of that flush timeout config is INT_MAX ? Is that why we are waiting forever ? 

Do you think if we should rather try to reconfigure that value for MM clusters ?

PR #963, opened 2023-10-23
Added new class to help log assignments

amybubu:
### Summary ###
Added class `AssignmentTaskMapLogger` to log assignments one task per log line so that logs stay under 1MB in size. Includes logic for checking if a task is greater than 1MB. If it is, that task will be split and logged in multiple parts.

BUG=[DATAPIPES-27351](https://jira01.corp.linkedin.com:8443/browse/DATAPIPES-27351)

### Testing Done ###
./gradlew build

In unit tests, the time it took to do logAssignment was always <50ms.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
@@ -18,14 +18,15 @@ import static com.linkedin.datastream.common.LogUtils.logStringsUnderSizeLimit;
  */
 public class AssignmentTaskMapLogger {
   private final Logger _log;
-  public static final double SIZE_LIMIT_BYTES = 1024 * 1024;
+  private double _sizeLimit;
 
   /**
    * Constructor
    * @param logger logger to be used for logging
    */
-  public AssignmentTaskMapLogger(Logger logger) {
+  public AssignmentTaskMapLogger(Logger logger, double logSizeLimitInBytes) {
     _log = logger;
+    _sizeLimit = logSizeLimitInBytes;
   }
 
   /**
@@ -35,7 +36,7 @@ public class AssignmentTaskMapLogger {
     for (String host: taskMap.keySet()) {
       for (DatastreamTask task: taskMap.get(host)) {
         String logContext = String.format("Host=%s: Live task", host);
-        logStringsUnderSizeLimit(_log, task.toString(), logContext, 1, SIZE_LIMIT_BYTES);
+        logStringsUnderSizeLimit(_log, task.toString(), logContext, 1, _sizeLimit);
       }
     }
   }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
@@ -5,13 +5,12 @@
  */
 package com.linkedin.datastream.server;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
 import org.slf4j.Logger;
-
-import com.linkedin.datastream.common.zk.ZkClient;
 
 
 /**
@@ -66,7 +65,7 @@ public class AssignmentTaskMapLogger {
    */
   private double getBlobSizeInBytes(String task) {
     if (Objects.nonNull(task) && !task.isEmpty()) {
-      return (task.getBytes(ZkClient.ZK_STRING_SERIALIZER_CHARSET).length);
+      return (task.getBytes(StandardCharsets.UTF_8).length);
     }
     return 0;
   }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
@@ -19,10 +19,9 @@ import com.linkedin.datastream.common.zk.ZkClient;
  */
 public class AssignmentTaskMapLogger {
   private final Logger _log;
-  private static final double ONE_MEBIBYTE = 1024 * 1024;
-  private static final double ONE_MB = 1.0;
-  private static final double BUFFER_1KB =  1.0 / 1024.0;
-  private static final int BUFFER_ADJUSTED_1MB = 1024 * 1020;
+  public static final double SIZE_LIMIT_BYTES = 1024 * 1024;
+  private static final double BUFFER_1KB =  1024;
+  private static final int BUFFER_ADJUSTED_LIMIT = (int) (SIZE_LIMIT_BYTES - BUFFER_1KB);
 
   /**
    * Constructor
@@ -44,41 +43,41 @@ public class AssignmentTaskMapLogger {
   }
 
   /**
-   * prints one log line for each task less than 1MB in size
+   * prints one log line for each task smaller than size limit
    * @param task
    */
   private void logTask(String task, String host, int part) {
-    if (isLessThan1MB(task)) {
+    if (isLessThanSizeLimit(task)) {
       if (part == 1) {
         _log.info("Host={}: Live task={}", host, task);
       } else {
         _log.info("Host={}: Live task (part {})={}", host, part, task);
       }
     } else {
-      _log.info("Host={}: Live task (part {})={}", host, part, task.substring(0, BUFFER_ADJUSTED_1MB));
-      logTask(task.substring(BUFFER_ADJUSTED_1MB), host, part + 1);
+      _log.info("Host={}: Live task (part {})={}", host, part, task.substring(0, BUFFER_ADJUSTED_LIMIT));
+      logTask(task.substring(BUFFER_ADJUSTED_LIMIT), host, part + 1);
     }
   }
 
   /**
-   * helper function to get the size of a task in string form (default charset UTF) in MBs
+   * helper function to get the size of a task in string form (default charset UTF) in bytes
    * @param task string form of task to measure
-   * @return size of task in MBS
+   * @return size of task in bytes
    */
-  private double getBlobSizeInMBs(String task) {
+  private double getBlobSizeInBytes(String task) {
     if (Objects.nonNull(task) && !task.isEmpty()) {
-      return ((double) task.getBytes(ZkClient.ZK_STRING_SERIALIZER_CHARSET).length) / ONE_MEBIBYTE;
+      return (task.getBytes(ZkClient.ZK_STRING_SERIALIZER_CHARSET).length);
     }
     return 0;
   }
 
   /**
-   * helper function to check if an object is less than 1 MB in size
+   * helper function to check if task size is less than size limit
    * @param task task to check size of
-   * @return true if object is less than 1MB in size
+   * @return true if object is less than size limit
    */
-  private boolean isLessThan1MB(String task) {
-    double sizeInMB = getBlobSizeInMBs(task);
-    return sizeInMB + BUFFER_1KB < ONE_MB;
+  private boolean isLessThanSizeLimit(String task) {
+    double sizeInMB = getBlobSizeInBytes(task);
+    return sizeInMB + BUFFER_1KB < SIZE_LIMIT_BYTES;
   }
 }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
@@ -23,6 +23,7 @@ public class AssignmentTaskMapLogger {
   /**
    * Constructor
    * @param logger logger to be used for logging
+   * @param logSizeLimitInBytes size limit for each log line in bytes
    */
   public AssignmentTaskMapLogger(Logger logger, double logSizeLimitInBytes) {
     _log = logger;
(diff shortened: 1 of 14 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * prints one log line for each task less than 1MB in size
+   * @param task
+   */
+  public void logTask(String task, String host, int part) {
+    try {
+      if (isLessThan1MB(task)) {
+        if (part == 1) {
+          _log.info("Host={}: Live task={}", host, part, task);
```
amybubu, line 54: delete "part"

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,92 @@
...
+        logTask(task.toString(), host, 1);
+      }
+    }
+  }
+
+  /**
+   * prints one log line for each task less than 1MB in size
+   * @param task
+   */
+  public void logTask(String task, String host, int part) {
```
amybubu, line 50: will change to private

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  /**
+   * helper function to serialize object for size measurement
+   * @param obj object to serialize
+   * @return serialized byte array of object
+   * @throws Exception
+   */
+  public byte[] serializeObject(Object obj) throws Exception {
```
amybubu, line 73: will change to private

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,92 @@
...
+    return bos.toByteArray();
+  }
+
+  /**
+   * helper function to check if an object is less than 1 MB in size
+   * @param obj object to check size of
+   * @return true if object is less than 1MB in size
+   * @throws Exception
+   */
+  public boolean isLessThan1MB(Object obj) throws Exception {
```
amybubu, line 87: will change to private

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,92 @@
...
+  public void logTask(String task, String host, int part) {
+    try {
+      if (isLessThan1MB(task)) {
+        if (part == 1) {
+          _log.info("Host={}: Live task={}", host, part, task);
+        } else {
+          _log.info("Host={}: Live task (part {})={}", host, part, task);
+        }
+      } else {
+        _log.info("Host={}: Live task (part {})={}", host, part, task.substring(0, NUM_OF_CHAR_IN_1MB));
```
amybubu, line 59: Will potentially change substring logic, since could be unnecessarily expensive. Will try to find library that can divide string into even chunks.

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
amybubu: NOTE: going to try to write unit tests, including checking how long serialization takes, as it could be overly expensive

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,84 @@
...
+        logTask(task.toString(), host, 1);
+      }
+    }
+  }
+
+  /**
+   * prints one log line for each task less than 1MB in size
+   * @param task
+   */
+  private void logTask(String task, String host, int part) {
```
shrinandthakkar, line 50: instead of using this only for task map, lets make this as a logger utility which can be used for parameterized string with parameterized limit.

and lets move it to datastream-utils/src/main/java/com/linkedin/datastream/common/LogUtils.java ?

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Class that logs assignments one task at a time, keeping each log line under 1 MB in size
+ */
+public class AssignmentTaskMapLogger {
+  private final Logger _log;
+  private static final double ONE_MEBIBYTE = 1024 * 1024;
+  private static final double ONE_MB = 1.0;
+  private static final double BUFFER_1KB =  1.0 / 1024.0;
+  private static final int BUFFER_ADJUSTED_1MB = 1024 * 1020;
```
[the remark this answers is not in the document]
amybubu: changing this to work with config

datastream-utils/src/main/java/com/linkedin/datastream/common/LogUtils.java
```diff
@@ -92,4 +95,44 @@ public static String logSummarizedTopicPartitionsMapping(List<String> partitions
             .append(logNumberArrayInRange(topicPartitionsMap.get(topicName))))
         .collect(Collectors.joining(", "));
   }
+
+  /**
+   * prints one log line for each string smaller than size limit
+   * @param  message
+   */
+  public static void logStringsUnderSizeLimit(Logger log, String message, String contextPrefix, int part, double sizeLimit) {
+    int bufferAdjustedSizeLimit = (int) (sizeLimit - BUFFER_1KB);
```
shrinandthakkar, line 104: This calculation can be done only once when you'd have a user facing public method instead of having it in both the this location and the `isLessThanSizeLimit` method during each recursive call.

datastream-utils/src/main/java/com/linkedin/datastream/common/LogUtils.java
```diff
@@ -92,4 +95,44 @@ public static String logSummarizedTopicPartitionsMapping(List<String> partitions
             .append(logNumberArrayInRange(topicPartitionsMap.get(topicName))))
         .collect(Collectors.joining(", "));
   }
+
+  /**
```
shrinandthakkar, line 99: also lets have java docs explaining what each parameter is for eg. `contextPrefix`.

datastream-utils/src/main/java/com/linkedin/datastream/common/LogUtils.java
```diff
@@ -92,4 +95,44 @@ public static String logSummarizedTopicPartitionsMapping(List<String> partitions
...
+   * prints one log line for each string smaller than size limit
+   * @param  message
+   */
+  public static void logStringsUnderSizeLimit(Logger log, String message, String contextPrefix, int part, double sizeLimit) {
+    int bufferAdjustedSizeLimit = (int) (sizeLimit - BUFFER_1KB);
+    if (isLessThanSizeLimit(message, sizeLimit)) {
+      if (part == 1) {
+        log.info("{}={}", contextPrefix,  message);
+      } else {
+        log.info("{} (part {})={}", contextPrefix, part,  message);
```
shrinandthakkar, line 109: what happens if the contextPrefix is more than the 1KB Buffer, the log will be truncated. So in the public facing method, lets also check for the size of that contextPrefix string and fail if it exceeds 1KB Buffer.

datastream-utils/src/main/java/com/linkedin/datastream/common/LogUtils.java
```diff
@@ -92,4 +95,68 @@ public static String logSummarizedTopicPartitionsMapping(List<String> partitions
...
+   * @param log logger to use for logging
+   * @param message string to be logged
+   * @param contextPrefix string that provides context for what is being logged
+   * @param part printed in the log message to keep track of how many parts a larger message has been split into
+   * @param sizeLimit size limit of each log line
+   * @param adjustedSizeLimit buffer adjusted size limit to account for extra bytes in the log line (e.g. timestamp, avro wrapping)
+   */
+  private static void logStringsUnderSizeLimit(Logger log, String message, String contextPrefix, int part,
+      double sizeLimit, int adjustedSizeLimit) {
+    if (isLessThanSizeLimit(message, sizeLimit)) {
```
shrinandthakkar, line 129: just pass the `adjustedSizeLimit` here to simplify the `isLessThanSizeLimit` function.

datastream-server/src/main/java/com/linkedin/datastream/server/AssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,43 @@
...
+ * Class that logs assignments one task at a time, keeping each log line under a size limit
+ */
+public class AssignmentTaskMapLogger {
+  private final Logger _log;
+  private double _sizeLimit;
+
+  /**
+   * Constructor
+   * @param logger logger to be used for logging
+   */
```
shrinandthakkar, line 26: java doc missing a parameter info here.

datastream-server/src/main/java/com/linkedin/datastream/server/zk/ZkAdapter.java
```diff
@@ -201,9 +206,9 @@ public ZkAdapter(String zkServers, String cluster, String defaultTransportProvid
    */
```
shrinandthakkar, line 206: nit: missing java doc update in this constructor.

datastream-server/src/test/java/com/linkedin/datastream/server/TestAssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,147 @@
...
+
+    verify(log, times(3)).info(eq("{}={}"), contains("Live task"), anyString());
+
+    verifyNoMoreInteractions(log);
+  }
+
+  @Test
+  public void testTaskWithSizeExactlySizeLimit() {
+    final Logger log = mock(Logger.class);
+    Mockito.doReturn(createCustomSizeString(SIZE_LIMIT_BYTES)).when(MOCK_LARGE_TASK).toString();
```
shrinandthakkar, line 59: how long does this test takes to pass ? maybe you can reduce the SIZE_LIMIT_BYTES to 1KB or lesser as well for tests and builds to pass quickly ?

shrinandthakkar:
Overall LGTM, just few minor comments! Thanks @amybubu for making this change!

datastream-server/src/test/java/com/linkedin/datastream/server/TestAssignmentTaskMapLogger.java
amybubu: build took 12 seconds but i can make it smaller

amybubu: I decided to make the size limit 4KB because if the limit is too small the buffer adjustment makes it so that message is split up even more than necessary, which is expected in practice because splitting more is more safe, but the unit test expectation would be less intuitive. For example, if the size limit was 2KB and we have a message of size 4KB, instead of the message getting split into 3 parts (4KB should be split into is 2 log lines but we "round up" so we split more to be safe), the buffer adjustment of 1KB would make it so that only 1KB of the message would get logged at a time, and so it actually gets split into at least 4 log lines, which is fine in practice but might be harder to understand when reading the unit test.

shrinandthakkar (approved):

datastream-server/src/test/java/com/linkedin/datastream/server/TestAssignmentTaskMapLogger.java
```diff
@@ -0,0 +1,147 @@
...
+    taskMap.put("host2", ImmutableSet.of(MOCK_LARGE_TASK));
+
+    AssignmentTaskMapLogger logger = new AssignmentTaskMapLogger(log, SIZE_LIMIT_BYTES);
+    long startTime = System.currentTimeMillis();
+    logger.logAssignment(taskMap);
+    long endTime = System.currentTimeMillis();
+    long duration = endTime - startTime;
+    System.out.println("Log assignment duration: " + duration + " ms");
+
+    verify(log, times(1)).info(eq("{}={}"), eq("Host=host1: Live task"), anyString());
```
hshukla, line 99: you do not need to explicitly call times(1), the default is 1.

datastream-utils/src/main/java/com/linkedin/datastream/common/LogUtils.java
```diff
@@ -92,4 +95,67 @@ public static String logSummarizedTopicPartitionsMapping(List<String> partitions
...
+   * @param part printed in the log message to keep track of how many parts a larger message has been split into
+   * @param adjustedSizeLimit buffer adjusted size limit to account for extra bytes in the log line (e.g. timestamp, avro wrapping)
+   */
+  private static void logStringsUnderSizeLimit(Logger log, String message, String contextPrefix, int part,
+      int adjustedSizeLimit) {
+    if (isLessThanSizeLimit(message, adjustedSizeLimit)) {
+      if (part == 1) {
+        log.info("{}={}", contextPrefix,  message);
+      } else {
+        log.info("{} (part {})={}", contextPrefix, part,  message);
```
hshukla, line 132: I see we are printing part here, which means it will show how many part the msg has been split into.  do we know the total part that needs to be printed? how do one know that there will not be more coming and this is last or if we miss something how do we know?

amybubu: The code doesn't predict how many parts it will eventually get split into, it will just print which the part number as it's being split to help the user keep track of which log lines belong to the same task.

hshukla: I think it would be a good feature to have to make sure we are not dropping any logs or the system is not dropping anything. it can be a follow up PR.

hshukla (approved):

shrinandthakkar (approved):

PR #975, opened 2024-01-17
Unblock waiting threads sooner

hshukla:
currently, during the shutdown, we always wait 60seconds before we unblock other waiting threads(zk callback thread or main threads) on coordinator object, this was being added as part of https://github.com/linkedin/brooklin/pull/964, our expectation was  the intrinsic CV will help us with notifying blocking threads sooner, but `notifyAll` was not mentioned at the right place, which was making other threads to be blocked on for 60 seconds always.

This PR uses explicit state variable for intrinsic CV, checks the new state variable `_handleEventCompleted` and wait only if false. From evenThread run method, we update the state variable and call notifyAll() outside the while loop; when shutdown is true. This will help blocking thread to unblock and have a chance to acquire the lock on the coordinator object and perform the clean shutdown.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -385,6 +385,10 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
       // explicitly via this CV.
       _log.info("Thread {} will wait for notification from the event thread for {} ms.",
           Thread.currentThread().getName(), duration.toMillis());
+      // The goal of this wait is to give eventThread a chance to acquire a lock on the coordinator object for the handleEvent
+      // ideally, we would use while loop here to avoid spurious signals, but in our case, we have a while loop
+      // which calls this method, so it is ok. Also, this if condition helps us to not continue go into wait mode,
+      // because it may cause shutdown to not run gracefully if the connecting deployment system has shorter timeouts
       if (!_handleEventCompleted) {
         this.wait(duration.toMillis());
       }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -221,6 +221,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   private final Map<String, SerdeAdmin> _serdeAdmins = new HashMap<>();
   private final Map<String, Authorizer> _authorizers = new HashMap<>();
   private volatile boolean _shutdown = false;
+  private volatile boolean _notified = false;
 
   private CoordinatorEventProcessor _eventThread;
   private ScheduledExecutorService _scheduledExecutor;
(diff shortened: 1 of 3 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -377,12 +378,15 @@ private synchronized boolean waitForEventThreadToJoin() {
...
   private synchronized void waitForNotificationFromEventThread(Duration duration) {
     try {
-      // This intrinsic conditional variable helps to halt threads (zk callback threads, main server thread) before
+      // This intrinsic conditional variable(CV) helps to halt threads (zk callback threads, main server thread) before
       // attempting to acquire the Coordinator object. We never halt the event thread (coordinator thread)
       // explicitly via this CV.
       _log.info("Thread {} will wait for notification from the event thread for {} ms.",
           Thread.currentThread().getName(), duration.toMillis());
-      this.wait(duration.toMillis());
+      while (!_notified) {
```
hshukla, line 386: this is worry some if event thread does not get free for 60 mins, we will keep be waiting which will trigger timeout on the deployment framework.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -219,6 +219,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   private final Map<String, SerdeAdmin> _serdeAdmins = new HashMap<>();
   private final Map<String, Authorizer> _authorizers = new HashMap<>();
   private volatile boolean _shutdown = false;
+  private volatile boolean _notified = false;
```
atoomula, line 222: Should we rename to coordinatorStopped or some such. Notified sounds too generic.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
atoomula: True but what alternate choice do we have ? Hopefully, this doesn’t happen often. Is the heaviest operation partition assignment ? If it is, then it should now take < 10 secs on an average ?

hshukla: in this case, I would just have one time wait, something like

if (!_notified) {
this.wait(duration.toMillis());
}

bcs, all we want is to let the evenThread enter into the handleEvent, we added all this logic for that purpose only, right?

atoomula: Are you suggest changing while to if condition ? While is there to handle other spurious signals and not via notify all

atoomula (dismissed):

atoomula (approved):
lgtm

atoomula (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -219,6 +219,9 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
   private final Map<String, SerdeAdmin> _serdeAdmins = new HashMap<>();
   private final Map<String, Authorizer> _authorizers = new HashMap<>();
   private volatile boolean _shutdown = false;
+  // TODO we have _shutdown, eventThread and now _handleEventCompleted, for some distinct usage,
+  //  we should revisit and refactor to have less variation
+  private volatile boolean _handleEventCompleted = false;
```
shrinandthakkar, line 224: nit: instead of `_handleEventCompleted ` can we name this variable to something like, `_coordinatorEventThreadExiting` ?

shrinandthakkar (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
hshukla: thought about this back and forth, ideally we just need to wait for the handleEvent to get the lock on the coordinator object, once we are out of it, zk expiry or main thread can get another chance to acquire lock. coordintorEventThread exists only after we get out of the run method, which may not be the case.

nothing will be exactly certain here, so tried to came up with the one which is more closer to the logic.

atoomula (approved):

shrinandthakkar (approved):

PR #978, opened 2024-01-19
Move wait after the interrupt

hshukla:
move wait after the interrupt.


amybubu (approved):

atoomula (approved):

atoomula:
Don't see any need for having waitForEventThreadToJoin(). Could we just remove that method definition and the call from stop() ? There is anyways stopEventThread() right after, which anyways calls interrupt() on event thread and then does whatever waitForEventThreadToJoin() does.

amybubu (approved):

atoomula (approved):

PR #979, opened 2024-01-24
Apply lock at Coordinator class level

hshukla:
As part of https://github.com/linkedin/brooklin/pull/975 where we used 

```
synchronized (this) {...}
```
it was giving lock on the `CoordinatorEventProcessor` instance.  However, we need a lock on the `Coordinator` class, hence the notifyAll was not working as expected.

This PR fixes the above regression by introducing 
```
synchronized (Coordinator.class) {
```

### DIFF
--- a/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
+++ b/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
@@ -3230,6 +3230,7 @@ public class TestCoordinator {
       Assert.assertTrue(t != null && t.isAlive());
       Assert.assertTrue(PollUtils.poll(() -> instance1.getIsLeader().getAsBoolean(), 100, 30000));
     }
+    instance1._coordinatorEventThreadExiting = false;
 
     instance1.stop();
     instance1.getDatastreamCache().getZkclient().close();
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -390,7 +390,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
       if (!_coordinatorEventThreadExiting) {
         _log.info("Thread {} will wait for notification from the event thread for {} ms.",
             Thread.currentThread().getName(), duration.toMillis());
-        this.wait(duration.toMillis());
+        wait(duration.toMillis());
       }
 
     } catch (InterruptedException exception) {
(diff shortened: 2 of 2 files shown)


atoomula (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -2370,9 +2370,11 @@ public void run() {
           _log.error("CoordinatorEventProcessor failed", t);
         }
       }
-      synchronized (this) {
+      synchronized (Coordinator.class) {
```
atoomula, line 2373: Should it be on the class or on the object (Coordinator.this) ?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -2370,9 +2370,11 @@ public void run() {
...
         _coordinatorEventThreadExiting = true;
-        this.notifyAll();
+        _log.info("notify waiting threads for Coordinator object lock, "
+            + "_coordinatorEventThreadExiting={}", _coordinatorEventThreadExiting);
+        notifyAll();
```
atoomula, line 2377: notifyAll should also be on the Coordinator.this.notifyAll() ?

atoomula (dismissed):

shrinandthakkar (dismissed):
great find!! 🙌

atoomula (approved):

amybubu (approved):

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -3230,6 +3230,7 @@ void testOnSessionExpired(boolean handleNewSession) throws DatastreamException,
       Assert.assertTrue(t != null && t.isAlive());
       Assert.assertTrue(PollUtils.poll(() -> instance1.getIsLeader().getAsBoolean(), 100, 30000));
     }
+    instance1._coordinatorEventThreadExiting = false;
```
shrinandthakkar, line 3233: interesting that we have to set this explicitly 🤔

shrinandthakkar (approved):

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
hshukla: The test do zkExpiry and then handleNewSession and then call stop. So during zkExpiry, we go from false -> true, and during stop if `_coordinatorEventThreadExiting ` stays false, then wait does not occur and creates deadlock.

PR #981, opened 2024-02-22
Add partitions to auto pause on consumer side topic access error

hshukla:
When Kafka consumer does not have access to topic, it gets `org.apache.kafka.common.errors.TopicAuthorizationException` with `Not authorized to access topics: [topicA, topicB]`. As this is a perm issue, there are high chances that it will not be resolved right away, so the subsequent poll for the same topic may get the same error. This can cause the issue if the consumer was subscribed to more topics' partitions which consumer has access for and no issues.

This PR will add the topics' partitions into the pause list if the consumer does not have access. The auto-pause is for 5 minutes(config driven); same as producer side pause.

before each poll, task checks if there are any partitions to be paused or resumed. This logic is already there, which will take care or resuming pause created due to consumer side access issue.

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -35,7 +35,9 @@ import org.apache.kafka.clients.consumer.NoOffsetForPartitionException;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.clients.consumer.OffsetOutOfRangeException;
 import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.TopicAuthorizationException;
 import org.apache.kafka.common.errors.WakeupException;
 import org.slf4j.Logger;
 
@@ -620,9 +622,23 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
       _logger.warn(
           String.format("Poll threw an exception. Sleeping for %d seconds and repoll from consumer, poll attempt: %d",
               _retrySleepDuration.getSeconds(), _pollAttempts.intValue()), e);
-    }
-    if (!_shutdown) {
-      Thread.sleep(_retrySleepDuration.toMillis());
+      // add partition into pause list on topic not authz for access
+      if (e instanceof TopicAuthorizationException) {
+        TopicAuthorizationException tae = (TopicAuthorizationException) e;
+        Set<String> unauthorizedTopics = tae.unauthorizedTopics();
+        _logger.warn("Not authorized to access, they will be added to auto-pause set, topics={}", unauthorizedTopics);
+        for (String topic: unauthorizedTopics) {
+          List<PartitionInfo> partitionInfos = _consumer.partitionsFor(topic);
+          for (PartitionInfo partitionInfo : partitionInfos) {
+            TopicPartition tp = new TopicPartition(topic, partitionInfo.partition());
+            _logger.warn("Adding source topic partition={} to auto-pause set", tp);
+            _autoPausedSourcePartitions.put(tp, PausedSourcePartitionMetadata.pollError(Instant.now(),
+                _pauseErrorPartitionDuration, PausedSourcePartitionMetadata.Reason.TOPIC_NOT_AUTHORIZED, tae));
+          }
+        }
+        // update datastream task for pause, this will be used in preConsumerPollHook
+        _taskUpdates.add(DatastreamConstants.UpdateType.PAUSE_RESUME_PARTITIONS);
+      }
     }
   }
 
@@ -946,6 +962,7 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
     long numAutoPausedPartitionsOnError = 0;
     long numAutoPausedPartitionsAwaitingDestTopic = 0;
     long numAutoPausedPartitionsOnInFlightMessages = 0;
+    long numAutoPausedPartitionsAwaitingSourceTopicAccess = 0;
     for (PausedSourcePartitionMetadata metadata : _autoPausedSourcePartitions.values()) {
       switch (metadata.getReason()) {
         case SEND_ERROR:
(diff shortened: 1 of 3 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -2364,6 +2368,7 @@ public void run() {
           CoordinatorEvent event = _eventQueue.take();
           if (event != null) {
             handleEvent(event);
+            _isLeaderBusy = false;
```
atoomula, line 2371: nit: should we set and unset _isLeaderBusy in the same scope ? Much easier to maintain the code that way. 

One way is:

_isLeaderBusy = _adapter.isLeader();
handleEvent();
_isLeaderBusy = false;

OR set and unset inside handleEvent() itself. Wdyt ?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -617,12 +619,24 @@ protected void handlePollRecordsException(Exception e) throws Exception {
...
-    if (!_shutdown) {
-      Thread.sleep(_retrySleepDuration.toMillis());
+      _logger.warn("Failed to poll. repoll from consumer, poll attempt={}", _pollAttempts.intValue(), e);
+      // add partition into pause list on topic not authz for access
+      if (e instanceof TopicAuthorizationException) {
+        TopicAuthorizationException tae = (TopicAuthorizationException) e;
+        final Set<String> unauthorizedTopics = tae.unauthorizedTopics();
+        _logger.warn("Not authorized to access, they will be added to auto-pause set, topics={}", unauthorizedTopics);
+        for (String topic: unauthorizedTopics) {
+          List<PartitionInfo> partitionInfos = _consumer.partitionsFor(topic);
```
atoomula, line 629: what is the desired behavior if partitionsFor throws an exception in this scenario ?

hshukla: I don't see it throwing exception. However, `KafkaConsumer::partitionFor()` method description says there could be exception, interesting. In this case, I think we should just catch and ignore it, so the behavior would be same as what we have on prod version right now.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
hshukla: I agree with you that it would be better to set and unset within the same scope. I think I will go with setting/unsetting inside the run() method as that is what the leader coordinator thread scope is. Right now we do `handleEvent()` in the future, if we enhance leader to perform other tasks, then we do not need to worry about changing anything.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -617,12 +619,29 @@ protected void handlePollRecordsException(Exception e) throws Exception {
       _logger.warn("Poll fail with exception", e);
       throw e;
     } else {
-      _logger.warn(
-          String.format("Poll threw an exception. Sleeping for %d seconds and repoll from consumer, poll attempt: %d",
-              _retrySleepDuration.getSeconds(), _pollAttempts.intValue()), e);
-    }
-    if (!_shutdown) {
-      Thread.sleep(_retrySleepDuration.toMillis());
```
shrinandthakkar, line 625: why are we removing this sleep behavior before next retry?

hshukla: that sleep of 5 seconds does not make sense. Ideally we need to have an exponential backoff retry. Here, by simply having 5 seconds sleep on any failure, can cause the lag to increase for other partitions which are assigned on the same task. Also, with this latest change, where we auto pause on source error, the sleep becomes not important.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaBasedConnectorTaskMetrics.java
```diff
@@ -110,6 +118,10 @@ public class KafkaBasedConnectorTaskMetrics extends CommonConnectorMetrics {
         NUM_AUTO_PAUSED_PARTITIONS_WAITING_FOR_DEST_TOPIC_PER_METRIC_KEY.computeIfAbsent(_fullMetricsKey, k -> new AtomicLong(0));
     DYNAMIC_METRICS_MANAGER.registerGauge(_className, AGGREGATE, NUM_AUTO_PAUSED_PARTITIONS_WAITING_FOR_DEST_TOPIC,
         numAutoPausedPartitionsAwaitingDestTopic::get);
+    AtomicLong numAutoPausedPartitionsAwaitingSourceTopicAccess =
+        NUM_AUTO_PAUSED_PARTITIONS_WAITING_SOURCE_TOPIC_ACCESS_PER_METRIC_KEY.computeIfAbsent(_fullMetricsKey, k -> new AtomicLong(0));
+    DYNAMIC_METRICS_MANAGER.registerGauge(_className, AGGREGATE, NUM_AUTO_PAUSED_PARTITIONS_WAITING_SOURCE_TOPIC_ACCESS,
```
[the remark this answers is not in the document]
hshukla: I was referring `numAutoPausedPartitionsAwaitingDestTopic` for example, which uses AGGREGATE, is that not right?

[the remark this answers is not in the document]
shrinandthakkar: But there are two different values updating the same metric key, 

`DYNAMIC_METRICS_MANAGER.registerGauge(_className, AGGREGATE, NUM_AUTO_PAUSED_PARTITIONS_WAITING_SOURCE_TOPIC_ACCESS,`

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
shrinandthakkar: makes sense

shrinandthakkar (approved):

atoomula (approved):

PR #1004, opened 2026-04-28
Add grace period for newly created CDC streams (suppress primary/alternate SLA, redirect latency to newlyOnboardedLatencyMs)

mittalprince:
## Summary                                                                                                                                                                                                                                                                                                                
  - Add a new `newStreamGracePeriodMs` config in `EventProducer` (default 2h, operator-overridable via `brooklin.server.eventProducer.newStreamGracePeriodMs` in `server.properties`). While newly created CDC streams drain their initial CDC backlog from EARLIEST, source-to-destination latencies are inherently large;
  without a grace window every event would be reported as primary-SLA-violating and lag alerts wired to `eventsLatencyMs` would fire.                                                                                                                                                                                       
  - During the grace window:
    - **Primary SLA** (`eventsProducedWithinSla` / `eventsProducedOutsideSla`) is suppressed.                                                                                                                                                                                                                               
    - **Alternate SLA** (`eventsProducedWithinAlternateSla` / `eventsProducedOutsideAlternateSla`) is suppressed.                                                                                                                                                                                                           
    - **`eventsLatencyMs` is redirected to `eventsLatencyMsSlaIneligible`** so lag alerts wired to `eventsLatencyMs` do not fire on initial backlog drain. The catch-up curve stays observable on the dedicated histogram.                                                                                                       
  - Warning logs and other rate / volume / send-latency metrics are NOT gated.                                                                                                                                                                                                                                              
  - Restrict the grace gate to CDC sources via an explicit `isCdcSource()` check (single-slash-URI heuristic, the same one used by the throughput-attribution metrics). MirrorMaker (`kafka://`) and other non-CDC sources are unaffected — neither SLA suppression nor the latency-histogram redirect applies.             
  - When multiple datastreams share a `DatastreamTask` via dedup, use the oldest creation time across the task so a freshly deduped stream cannot suppress metrics on long-running siblings.                                                                                                                                
  - All metric-emission gating goes through a single `shouldEmitMetric()` wrapper. Future suppression conditions (e.g. per-datastream opt-out flags) should be ORed into that one method so call sites stay simple.                                                                                                         
                                                                                                                                                                                                                                                                                                                            
  ## Notes / scope                                                                                                                                                                                                                                                                                                          
  - This PR was originally also going to change the no-offset-found fallback in `AbstractKafkaBasedConnectorTask` from LATEST to EARLIEST, but that change has been reverted (`390132ee`). LinkedIn-internal CDC connectors no longer use this base class (migrated to `AbstractXinfraBasedConnectorTask`); CDC's EARLIEST  
  behaviour is handled in the internal connector subclass.                                                                                                                                                                                                                                                                  
  - MySQL CDC will continue to default to LATEST since it does not depend on existing snapshots.
  - Distinct from #1005: that PR introduces `slaExcludedLatencyMs` for a per-datastream opt-out flag (permanent, manual). This PR introduces `newlyOnboardedLatencyMs` for an automatic time-bounded grace window for new CDC streams. Different histograms, different triggers, no merge conflict.                         
                                                                                                                                                                                                                                                                                                                            
  ## Test plan                                                                                                                                                                                                                                                                                                              
  - [x] 11 tests in `TestEventProducer` covering:                                                                                                                                                                                                                                                                           
    - New CDC stream in grace → primary and alternate SLA both suppressed                                                                                                                                                                                                                                                   
    - Old CDC stream past grace → primary and alternate SLA emit normally                                                                                                                                                                                                                                                   
    - MirrorMaker source not affected even with fresh `CREATION_MS`                                                                                                                                                                                                                                                         
    - Fail-open when `CREATION_MS` is missing or malformed                                                                                                                                                                                                                                                                  
    - `eventsLatencyMs` redirected to `newlyOnboardedLatencyMs` during grace                                                                                                                                                                                                                                                
    - `eventsLatencyMs` fires (and `newlyOnboardedLatencyMs` does not) post-grace                                                                                                                                                                                                                                           
    - Custom `newStreamGracePeriodMs` config override honoured                                                                                                                                                                                                                                                              
    - Deduped task with oldest stream past grace → metrics emit normally                                                                                                                                                                                                                                                    
    - Deduped task with all streams within grace → metrics suppressed/redirected uniformly                                                                                                                                                                                                                                  
  - [x] All 16 tests in `TestEventProducer` pass under JDK 8 (project's required toolchain):
        `JAVA_HOME=<jdk8> ./gradlew :datastream-server:test --tests com.linkedin.datastream.server.TestEventProducer`

### DIFF
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -604,9 +604,8 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected void handleNoOffsetForPartitionException(NoOffsetForPartitionException e) {
     _logger.info("Poll threw NoOffsetForPartitionException for partitions {}.", e.partitions());
     if (!_shutdown) {
-      // Default to EARLIEST for CDC connectors to avoid data loss for CDC+Bootstrap use-case.
-      // Starting from EARLIEST ensures the consumer reads all records including bootstrap
-      // snapshot data and change events from the bootstrap window. (DATAPIPES-33203)
+      // Default to EARLIEST to reuse any existing snapshot/data already in the topic when no
+      // committed offset is found. The auto.offset.reset consumer property wins if explicitly set.
       seekToStartPosition(_consumer, e.partitions(),
           _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_EARLIEST));
     }
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -604,9 +604,11 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected void handleNoOffsetForPartitionException(NoOffsetForPartitionException e) {
     _logger.info("Poll threw NoOffsetForPartitionException for partitions {}.", e.partitions());
     if (!_shutdown) {
-      // Seek to start position, by default we are starting from latest one as we just start consumption
+      // Default to EARLIEST for CDC connectors to avoid data loss for CDC+Bootstrap use-case.
+      // Starting from EARLIEST ensures the consumer reads all records including bootstrap
+      // snapshot data and change events from the bootstrap window. (DATAPIPES-33203)
       seekToStartPosition(_consumer, e.partitions(),
-          _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_LATEST));
+          _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_EARLIEST));
     }
   }
 
--- a/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
+++ b/datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
@@ -604,10 +604,9 @@ abstract public class AbstractKafkaBasedConnectorTask implements Runnable, Consu
   protected void handleNoOffsetForPartitionException(NoOffsetForPartitionException e) {
     _logger.info("Poll threw NoOffsetForPartitionException for partitions {}.", e.partitions());
     if (!_shutdown) {
-      // Default to EARLIEST to reuse any existing snapshot/data already in the topic when no
-      // committed offset is found. The auto.offset.reset consumer property wins if explicitly set.
+      // Seek to start position, by default we are starting from latest one as we just start consumption
       seekToStartPosition(_consumer, e.partitions(),
-          _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_EARLIEST));
+          _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_LATEST));
     }
   }
 
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
@@ -73,6 +73,7 @@ public class EventProducer implements DatastreamEventProducer {
   public static final String DEFAULT_FLUSH_INTERVAL_MS = String.valueOf(Duration.ofMinutes(5).toMillis());
 
   static final String EVENTS_LATENCY_MS_STRING = "eventsLatencyMs";
+  static final String SLA_EXCLUDED_LATENCY_MS_STRING = "slaExcludedLatencyMs";
   static final String EVENTS_SEND_LATENCY_MS_STRING = "eventsSendLatencyMs";
   static final String THROUGHPUT_VIOLATING_EVENTS_LATENCY_MS_STRING = "throughputViolatingEventsLatencyMs";
   static final String THROUGHPUT_VIOLATING_EVENTS_SEND_LATENCY_MS_STRING = "throughputViolatingEventsSendLatencyMs";
@@ -456,11 +457,15 @@ public class EventProducer implements DatastreamEventProducer {
     if (eventsSourceTimestamp > 0) {
       // Report availability metrics
       long sourceToDestinationLatencyMs = System.currentTimeMillis() - eventsSourceTimestamp;
-      reportEventLatencyMetrics(topicOrDatastreamName, metadata, sourceToDestinationLatencyMs, EVENTS_LATENCY_MS_STRING);
+      // Redirect the latency histogram to slaExcludedLatencyMs during the CDC catch-up grace window
+      // so lag alerts wired to eventsLatencyMs do not fire on the initial backlog drain. The catch-up
+      // curve is still observable on slaExcludedLatencyMs.
+      String latencyMetricName = isWithinSlaGracePeriod() ? SLA_EXCLUDED_LATENCY_MS_STRING : EVENTS_LATENCY_MS_STRING;
+      reportEventLatencyMetrics(topicOrDatastreamName, metadata, sourceToDestinationLatencyMs, latencyMetricName);
 
-      // During the CDC catch-up grace window suppress only the primary SLA counter so the strict
-      // 1-minute dashboard isn't polluted by initial backlog drain. The alternate SLA pair keeps
-      // emitting throughout — operators retain a coarser SLA signal end-to-end.
+      // Primary SLA pair is suppressed during grace so the strict 1-minute dashboard isn't polluted
+      // by initial backlog drain. The alternate SLA pair keeps emitting throughout — operators retain
+      // a coarser SLA signal end-to-end.
       if (!isWithinSlaGracePeriod()) {
         reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdSlaMs,
             EVENTS_PRODUCED_WITHIN_SLA, EVENTS_PRODUCED_OUTSIDE_SLA);
@@ -735,6 +740,9 @@ public class EventProducer implements DatastreamEventProducer {
     metrics.add(new BrooklinCounterInfo(METRICS_PREFIX + EVENTS_PRODUCED_OUTSIDE_ALTERNATE_SLA));
     metrics.add(new BrooklinCounterInfo(METRICS_PREFIX + DROPPED_SENT_FROM_SERIALIZATION_ERROR));
     metrics.add(new BrooklinHistogramInfo(METRICS_PREFIX + EVENTS_LATENCY_MS_STRING, Optional.of(
+        Arrays.asList(BrooklinHistogramInfo.PERCENTILE_50, BrooklinHistogramInfo.PERCENTILE_99,
+            BrooklinHistogramInfo.PERCENTILE_999))));
+    metrics.add(new BrooklinHistogramInfo(METRICS_PREFIX + SLA_EXCLUDED_LATENCY_MS_STRING, Optional.of(
         Arrays.asList(BrooklinHistogramInfo.PERCENTILE_50, BrooklinHistogramInfo.PERCENTILE_99,
             BrooklinHistogramInfo.PERCENTILE_999))));
     metrics.add(new BrooklinHistogramInfo(METRICS_PREFIX + EVENTS_SEND_LATENCY_MS_STRING));
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
@@ -194,13 +194,7 @@ public class EventProducer implements DatastreamEventProducer {
 
     _newStreamSlaGracePeriodMs = Long.parseLong(
         config.getProperty(NEW_STREAM_SLA_GRACE_PERIOD_MS, DEFAULT_NEW_STREAM_SLA_GRACE_PERIOD_MS));
-
-    // Source-database parsing pulled forward so we can gate stream-creation-time parsing on
-    // whether this is a CDC source. Non-CDC sources (BMM kafka://, Inlogs, etc.) don't have a
-    // bootstrap window worth masking, so skip the work entirely for them.
-    String[] sourceParts = getSourcePathParts();
-    _sourceDatabase = (sourceParts != null && sourceParts.length > 1) ? sourceParts[1] : null;
-    _streamCreationTimeMs = (_sourceDatabase != null) ? parseStreamCreationTimeMs(task) : 0L;
+    _streamCreationTimeMs = parseStreamCreationTimeMs(task);
 
     _warnLogLatencyEnabled =
         Boolean.parseBoolean(config.getProperty(WARN_LOG_LATENCY_ENABLED, DEFAULT_WARN_LOG_LATENCY_ENABLED));
@@ -226,6 +220,9 @@ public class EventProducer implements DatastreamEventProducer {
 
     _enableThroughputMetrics =
         Boolean.parseBoolean(config.getProperty(CONFIG_ENABLE_THROUGHPUT_METRICS, Boolean.FALSE.toString()));
+
+    String[] sourceParts = getSourcePathParts();
+    _sourceDatabase = (sourceParts != null && sourceParts.length > 1) ? sourceParts[1] : null;
 
     _logger.info("Created event producer with customCheckpointing={}", customCheckpointing);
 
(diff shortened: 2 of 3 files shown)


datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -604,9 +604,11 @@ protected void handleOffsetOutOfRangeException(OffsetOutOfRangeException e) {
   protected void handleNoOffsetForPartitionException(NoOffsetForPartitionException e) {
     _logger.info("Poll threw NoOffsetForPartitionException for partitions {}.", e.partitions());
     if (!_shutdown) {
-      // Seek to start position, by default we are starting from latest one as we just start consumption
+      // Default to EARLIEST for CDC connectors to avoid data loss for CDC+Bootstrap use-case.
```
akshayrai: > to avoid data loss for CDC+Bootstrap use-case.

We are seeking to earliest to reuse existing snapshots when available. Not because of data loss issues.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -604,9 +604,11 @@ protected void handleOffsetOutOfRangeException(OffsetOutOfRangeException e) {
...
+      // Starting from EARLIEST ensures the consumer reads all records including bootstrap
```
akshayrai: >  reads all records including bootstrap

Nit: Not true. We are still reading CDC data here. This isn't using log compacted topic.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -409,11 +439,24 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
...
-          EVENTS_PRODUCED_WITHIN_ALTERNATE_SLA, EVENTS_PRODUCED_OUTSIDE_ALTERNATE_SLA);
+      // Skip SLA reporting for newly created CDC streams during grace period (DATAPIPES-33203).
+      // CDC streams starting from EARLIEST offset have large initial lag during bootstrap catch-up
+      // which would trigger false SLA violations. CDC sources are identified by single-slash URI
+      // scheme (e.g. espresso:/, mysql:/, tidb:/) which yields a non-null _sourceDatabase; BMM
+      // double-slash URIs (kafka://) yield null and are excluded so MirrorMaker reports SLA from
+      // the first event.
+      boolean isCdcSource = _sourceDatabase != null;
+      boolean isWithinGracePeriod = isCdcSource
+          && _streamCreationTimeMs > 0
```
akshayrai: drop this condition (redundant)

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -409,11 +439,24 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
       long sourceToDestinationLatencyMs = System.currentTimeMillis() - eventsSourceTimestamp;
       reportEventLatencyMetrics(topicOrDatastreamName, metadata, sourceToDestinationLatencyMs, EVENTS_LATENCY_MS_STRING);
 
-      reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdSlaMs,
-          EVENTS_PRODUCED_WITHIN_SLA, EVENTS_PRODUCED_OUTSIDE_SLA);
-
-      reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdAlternateSlaMs,
-          EVENTS_PRODUCED_WITHIN_ALTERNATE_SLA, EVENTS_PRODUCED_OUTSIDE_ALTERNATE_SLA);
+      // Skip SLA reporting for newly created CDC streams during grace period (DATAPIPES-33203).
```
akshayrai: drop jira ticket ref. since it doesn't make sense in open source code.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -409,11 +439,24 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
...
+      // CDC streams starting from EARLIEST offset have large initial lag during bootstrap catch-up
+      // which would trigger false SLA violations. CDC sources are identified by single-slash URI
+      // scheme (e.g. espresso:/, mysql:/, tidb:/) which yields a non-null _sourceDatabase; BMM
+      // double-slash URIs (kafka://) yield null and are excluded so MirrorMaker reports SLA from
+      // the first event.
+      boolean isCdcSource = _sourceDatabase != null;
+      boolean isWithinGracePeriod = isCdcSource
+          && _streamCreationTimeMs > 0
+          && (System.currentTimeMillis() - _streamCreationTimeMs) < _newStreamSlaGracePeriodMs;
```
akshayrai: This is specific to CDC. It would be cleaner to wrap this up in a separate method.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -185,6 +192,29 @@ public EventProducer(DatastreamTask task, TransportProvider transportProvider, C
...
+            .map(ds -> ds.getMetadata().getOrDefault(DatastreamMetadataConstants.CREATION_MS, "0"))
+            .mapToLong(Long::parseLong)
+            .filter(ms -> ms > 0)
+            .min()
+            .orElse(0L);
+      }
+    } catch (Exception e) {
+      _logger.warn("Failed to parse stream creation time, SLA grace period will not be applied", e);
+    }
+    _streamCreationTimeMs = creationMs;
```
akshayrai: The grace period concept is specific to CDC. Can we modify the code for this to be explicit? We need not compute this for BMM/Inlogs right?

mittalprince: Done — _streamCreationTimeMs = (_sourceDatabase != null) ? parseStreamCreationTimeMs(task) : 0L;. BMM / Inlogs skip the parsing entirely; only CDC sources walk the task.getDatastreams() metadata.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
mittalprince: Extracted into isWithinSlaGracePeriod() (and the parsing into parseStreamCreationTimeMs(task)). Call site in reportMetrics is now a single if (!isWithinSlaGracePeriod()) { ... }.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
mittalprince: removed

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
mittalprince: updated

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
mittalprince: updated

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -409,11 +454,16 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
       long sourceToDestinationLatencyMs = System.currentTimeMillis() - eventsSourceTimestamp;
       reportEventLatencyMetrics(topicOrDatastreamName, metadata, sourceToDestinationLatencyMs, EVENTS_LATENCY_MS_STRING);
 
-      reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdSlaMs,
-          EVENTS_PRODUCED_WITHIN_SLA, EVENTS_PRODUCED_OUTSIDE_SLA);
+      // SLA reporting is suppressed while a CDC stream is still draining its initial bootstrap
```
akshayrai: nit: initial cdc backlog

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -342,6 +356,37 @@ private DatastreamRecordMetadata helperSendOrBroadcast(DatastreamProducerRecord
     return broadcastMetadata;
   }
 
+  /**
+   * Returns true while a CDC stream is still within its bootstrap-catch-up grace window.
```
akshayrai: nit: cdc-catch-up

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -185,6 +192,16 @@ public EventProducer(DatastreamTask task, TransportProvider transportProvider, C
...
 
+    _newStreamSlaGracePeriodMs = Long.parseLong(
+        config.getProperty(NEW_STREAM_SLA_GRACE_PERIOD_MS, DEFAULT_NEW_STREAM_SLA_GRACE_PERIOD_MS));
+
+    // Source-database parsing pulled forward so we can gate stream-creation-time parsing on
+    // whether this is a CDC source. Non-CDC sources (BMM kafka://, Inlogs, etc.) don't have a
+    // bootstrap window worth masking, so skip the work entirely for them.
+    String[] sourceParts = getSourcePathParts();
+    _sourceDatabase = (sourceParts != null && sourceParts.length > 1) ? sourceParts[1] : null;
+    _streamCreationTimeMs = (_sourceDatabase != null) ? parseStreamCreationTimeMs(task) : 0L;
```
akshayrai: It's a bit hacky to hide BMM by assuming creation time as 0. I think we should be explicit about it.

My suggestion would be to wrap this inside isWithinSlaGracePeriod method. If isCDCSource(), then compute grace period as below, otherwise, return 0 grace period. wdyt?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
```diff
@@ -604,9 +604,10 @@ protected void handleOffsetOutOfRangeException(OffsetOutOfRangeException e) {
   protected void handleNoOffsetForPartitionException(NoOffsetForPartitionException e) {
     _logger.info("Poll threw NoOffsetForPartitionException for partitions {}.", e.partitions());
     if (!_shutdown) {
-      // Seek to start position, by default we are starting from latest one as we just start consumption
+      // Default to EARLIEST to reuse any existing snapshot/data already in the topic when no
+      // committed offset is found. The auto.offset.reset consumer property wins if explicitly set.
       seekToStartPosition(_consumer, e.partitions(),
-          _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_LATEST));
+          _consumerProps.getProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, CONSUMER_AUTO_OFFSET_RESET_CONFIG_EARLIEST));
```
akshayrai: Does this cover all CDC connectors? Is it possible few are not using this class or overriding this?

mittalprince: In this repo, only AbstractKafkaBasedConnectorTask defines handleNoOffsetForPartitionException and no subclass (KafkaConnectorTask, KafkaMirrorMakerConnectorTask) overrides it. Any CDC connector that extends this base class inherits the new EARLIEST default automatically.

akshayrai (dismissed):

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/AbstractKafkaBasedConnectorTask.java
tanmaychandekar98: FYI, we have moved all our connectors to Xinfra, so this class would not be used anyways

mittalprince: Thanks for pointing out this Tanmay. I have removed this logic

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -72,6 +73,7 @@ public class EventProducer implements DatastreamEventProducer {
   public static final String DEFAULT_FLUSH_INTERVAL_MS = String.valueOf(Duration.ofMinutes(5).toMillis());
 
   static final String EVENTS_LATENCY_MS_STRING = "eventsLatencyMs";
+  static final String SLA_EXCLUDED_LATENCY_MS_STRING = "slaExcludedLatencyMs";
```
khandelwal-ayush: Since we are now emitting latency for newly onboarded streams separately - this is only to avoid oncall toil and not SLO , we should rename this metric to "newlyOnboardedLatencyMs"

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -114,6 +123,10 @@ public class EventProducer implements DatastreamEventProducer {
   private final int _availabilityThresholdSlaMs;
   // Alternate SLA for comparison with the main SLA
   private final int _availabilityThresholdAlternateSlaMs;
+  // Grace period for new streams: skip SLA reporting for streams newer than this duration
+  private final long _newStreamSlaGracePeriodMs;
```
khandelwal-ayush: Rename this and remove "Sla" from here. the same grace period is used in latency emission which has nothing to do today with SLA.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -407,13 +475,25 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
...
+      // While SLA emission is suppressed, the alternate SLA is still measured against its
+      // threshold but routed to a dedicated slaExcluded* counter pair so operators can track
+      // suppression-period stats without polluting the regular alternate-SLA dashboard.
+      if (shouldEmitSlaMetric()) {
+        reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdSlaMs,
+            EVENTS_PRODUCED_WITHIN_SLA, EVENTS_PRODUCED_OUTSIDE_SLA);
+
+        reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdAlternateSlaMs,
+            EVENTS_PRODUCED_WITHIN_ALTERNATE_SLA, EVENTS_PRODUCED_OUTSIDE_ALTERNATE_SLA);
+      } else {
```
khandelwal-ayush: we can remove this else block.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -90,8 +92,15 @@ public class EventProducer implements DatastreamEventProducer {
...
   private static final String NUM_EVENTS_OUTSIDE_ALT_SLA_LOG_FREQUENCY_MS = "numEventsOutsideAltSlaFrequencyMs";
+  private static final String NEW_STREAM_SLA_GRACE_PERIOD_MS = "newStreamSlaGracePeriodMs";
+  private static final String DEFAULT_NEW_STREAM_SLA_GRACE_PERIOD_MS = "7200000"; // 2 hours
   private static final String EVENTS_PRODUCED_OUTSIDE_SLA = "eventsProducedOutsideSla";
   private static final String EVENTS_PRODUCED_OUTSIDE_ALTERNATE_SLA = "eventsProducedOutsideAlternateSla";
+  // Counterparts to EVENTS_PRODUCED_*_ALTERNATE_SLA, emitted only while a CDC stream is inside its
+  // catch-up grace window so the regular alternate-SLA dashboards stay clean while operators can
+  // still track grace-period stats on a dedicated counter.
+  private static final String SLA_EXCLUDED_WITHIN_ALTERNATE_SLA = "slaExcludedWithinAlternateSla";
+  private static final String SLA_EXCLUDED_OUTSIDE_ALTERNATE_SLA = "slaExcludedOutsideAlternateSla";
```
khandelwal-ayush: remove this too.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -407,13 +472,21 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
...
-
-      reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdSlaMs,
-          EVENTS_PRODUCED_WITHIN_SLA, EVENTS_PRODUCED_OUTSIDE_SLA);
-
-      reportSLAMetrics(topicOrDatastreamName, sourceToDestinationLatencyMs <= _availabilityThresholdAlternateSlaMs,
-          EVENTS_PRODUCED_WITHIN_ALTERNATE_SLA, EVENTS_PRODUCED_OUTSIDE_ALTERNATE_SLA);
+      // Redirect the latency histogram to newlyOnboardedLatencyMs while SLA emission is suppressed
+      // so lag alerts wired to eventsLatencyMs do not fire on the initial CDC catch-up. The
+      // catch-up curve is still observable on newlyOnboardedLatencyMs.
+      String latencyMetricName = shouldEmitMetric() ? EVENTS_LATENCY_MS_STRING : NEWLY_ONBOARDED_LATENCY_MS_STRING;
```
manishGoyalCode: why are we creating a new metric NEWLY_ONBOARDED_LATENCY_MS_STRING . we can resue SLAExcluded latency metrics itself to keep it generic.

mittalprince: we are not emitting any SLA counters for SLA-excluded topics (throughput-violating topics) — we only emit latency metrics for those. So a generic slaExcludedLatencyMs name would imply SLA semantics that the metric doesn't actually have.

khandelwal-ayush (dismissed):
Looks good now.

harshOSS (dismissed):

harshOSS (dismissed):

akshayrai (approved):

khandelwal-ayush (approved):

PR #1005, opened 2026-04-29
Add per-datastream SLA metric opt-out via system.disableSlaMetric 

manishGoyalCode:
## Summary                                                                                 
  - Introduces a new datastream metadata flag `system.disableSlaMetric` that lets individual streams opt out of SLA metric emission.
  - When enabled, `EventProducer` suppresses both primary and alternate SLA counters (`eventsProducedWithinSla` / `eventsProducedOutsideSla`) and redirects the latency histogram from `eventsLatencyMs` to `eventsLatencyMsSlaIneligible`, so lag   
  alerts wired to `eventsLatencyMs` do not fire for opted-out streams while their latency curve remains observable.                                                                                                                                  
  - The flag is read once at construction (`getDisableSlaMetric`) and gated through the existing `shouldEmitMetric()` chokepoint, keeping all suppression conditions (grace period + opt-out) in one place.                                          
  - Honored only for non-deduped tasks (single datastream owners). Deduped tasks short-circuit to `false` so one stream's opt-out cannot suppress `eventsLatencyMs` for every stream sharing the `EventProducer`.                                    
                                                                                                                                                                                                                                                     
  ## Test plan                                                                                                                                                                                                                                       
  - [ ] Unit tests cover: opt-out suppresses SLA counters and redirects latency histogram; deduped tasks ignore the flag; default behavior unchanged when flag is absent or `false`.                                                                 
  - [ ] Verify on a canary stream with `system.disableSlaMetric=true` that `eventsLatencyMs` / `eventsProducedWithinSla` / `eventsProducedOutsideSla` are not emitted and `eventsLatencyMsSlaIneligible` is.                                         
  - [ ] Verify a regular stream alongside it still emits the standard SLA metrics.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
@@ -383,7 +383,7 @@ public class EventProducer implements DatastreamEventProducer {
    * sites do not need to know about every gating condition individually.
    */
   private boolean shouldEmitMetric() {
-    if(_disableSlaMetric){
+    if (_disableSlaMetric) {
       return false;
     }
     if (isWithinGracePeriod()) {
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
@@ -592,27 +592,20 @@ public class EventProducer implements DatastreamEventProducer {
    * Looks for config {@value CFG_DISABLE_SLA_METRIC} in the datastream metadata and returns its value.
    * Default value is false.
    *
-   * <p>The flag is only honored for tasks owning a single datastream. When a task hosts multiple
-   * datastreams (i.e. they are deduped onto the same task and share one EventProducer), one
-   * stream's opt-out would otherwise suppress {@code eventsLatencyMs} for every stream in the
-   * group. In that case the flag is ignored and a warning is logged.
+   * <p>Only honored for tasks owning a single datastream. Deduped tasks (multiple datastreams
+   * sharing one EventProducer) always return false, since one stream's opt-out would otherwise
+   * suppress {@code eventsLatencyMs} for every stream in the group.
    */
   private boolean getDisableSlaMetric(DatastreamTask task) {
-    boolean requested = Boolean.parseBoolean(task.getDatastreams()
+    if (task.getDatastreams().size() > 1) {
+      return false;
+    }
+    return Boolean.parseBoolean(task.getDatastreams()
         .stream()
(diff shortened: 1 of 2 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -207,6 +209,8 @@ public EventProducer(DatastreamTask task, TransportProvider transportProvider, C
         Boolean.parseBoolean(config.getProperty(CONFIG_ENABLE_PER_TOPIC_EVENT_LATENCY_METRICS,
             Boolean.FALSE.toString()));
 
+    _disableSlaMetric = getDisableSlaMetric(task);
```
akshayrai, line 225: We need to to be mindful of not creating deduped test streams and then disabling sla since it will disable it for all streams in that group.

Can we include some additional check to ensure this flag works only for non-deduped streams?

manishGoyalCode: Updated getDisableSlaMetric to short-circuit when task.getDatastreams().size() > 1 (the deduped case), with a warning log. Added testDisableSlaMetricIgnoredOnDedupedTask to lock in the behavior. The flag now only takes effect for tasks  owning a single datastream.

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -582,6 +588,33 @@ private Boolean getSkipMessageOnSerializationErrors(DatastreamTask task, Propert
...
+  /**
+   * Looks for config {@value CFG_DISABLE_SLA_METRIC} in the datastream metadata and returns its value.
+   * Default value is false.
+   *
+   * <p>The flag is only honored for tasks owning a single datastream. When a task hosts multiple
+   * datastreams (i.e. they are deduped onto the same task and share one EventProducer), one
+   * stream's opt-out would otherwise suppress {@code eventsLatencyMs} for every stream in the
+   * group. In that case the flag is ignored and a warning is logged.
+   */
+  private boolean getDisableSlaMetric(DatastreamTask task) {
```
akshayrai, line 674: I think we can clean up and make this a bit more readable?

basically return false whenever task.getDatastreams().size() is more than 1, irrespective of requested flag?

manishGoyalCode: cleaned up

akshayrai (dismissed):

akshayrai (dismissed):

kanishkjaiswal2015 (approved):
LGTM

akshayrai (approved):

PR #1012, opened 2026-05-16
Add CDC-bootstrap-aware dedup logic to SourceBasedDeduper

mittalprince:
## Summary

  - **SourceBasedDeduper**: Added two time-window guards to prevent unsafe cross-type dedup between CDC-only and CDC+BST streams:
    - `newStreamGracePeriodMs`: blocks a new CDC-only stream from deduping against a CDC+BST stream still in its catch-up grace period. The CDC+BST stream's offsets are behind the live tail — deduping would cause the CDC-only stream to miss events.
    - `snapshotWorstRefreshDays`: blocks a new CDC+BST stream from deduping against a CDC-only stream created within the worst-case GPFS snapshot staleness window (4 days). The CDC-only stream's offset 
  history would not reach back far enough for the CDC+BST stream's required earlier start position.
  - **SourceBasedDeduperFactory**: Pass `deduperProperties` through to the `SourceBasedDeduper` constructor so domain-sliced connector config reaches the deduper correctly.
  
  Both windows default to `0` (disabled), preserving existing behavior for all non-CDC connectors. When `system.cdcBootstrapRequired` is absent on the new stream the legacy source-equality path is taken 
  unchanged.
  
  ## Testing Done
  
  - [x] `testCdcBstNewVsExistingCdcBst` — same type always dedups
  - [x] `testCdcOnlyNewVsExistingCdcBstGraceActive` — CDC-only blocked during CDC+BST grace window
  - [x] `testCdcOnlyNewVsExistingCdcBstGraceExpired` — CDC-only allowed after grace expires
  - [x] `testCdcBstNewVsExistingCdcOnlySnapshotWindowActive` — CDC+BST blocked during snapshot window
  - [x] `testCdcBstNewVsExistingCdcOnlySnapshotWindowExpired` — CDC+BST allowed after window expires
  - [x] `testCdcOnlyNewVsExistingCdcOnly` — same type always dedups
  - [x] `testLegacyPathNoCdcBootstrapKey` — legacy source-equality path unaffected
  - [x] `testCdcBstNewPrefersExistingCdcBstOverExpiredCdcOnly` — priority ordering verified
  - [x] `testCdcOnlyNewPrefersExpiredCdcBstOverCdcOnly` — priority ordering verified
  - [x] `testCdcOnlyNewVsCdcBstMissingCreationMs` — conservative block when `system.creation.ms` absent
  - [x] `testCdcBstNewVsCdcOnlyMissingCreationMs` — conservative block when `system.creation.ms` absent
  - [x] `testDefaultDedupeConfigWindowsDisabled` — `windowMs=0` always allows dedup (backward-compatible default)
  - [x] `testCdcAwarePathDifferentSourceNoMatch` — different source → no match

### DIFF
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
@@ -143,13 +143,4 @@ public class DatastreamMetadataConstants {
    */
   public static final String CDC_BOOTSTRAP_REQUIRED_KEY = "system.cdcBootstrapRequired";
 
-  /**
-   * Start position value indicating the consumer should begin from the earliest available offset.
-   */
-  public static final String CDC_START_POSITION_EARLIEST = "earliest";
-
-  /**
-   * Start position value indicating the consumer should begin from the latest offset.
-   */
-  public static final String CDC_START_POSITION_LATEST = "latest";
 }
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
@@ -142,4 +142,14 @@ public class DatastreamMetadataConstants {
    * initialized to an earlier position to replay historical data before catching up to the live tail.
    */
   public static final String CDC_BOOTSTRAP_REQUIRED_KEY = "system.cdcBootstrapRequired";
+
+  /**
+   * Start position value indicating the consumer should begin from the earliest available offset.
+   */
+  public static final String CDC_START_POSITION_EARLIEST = "earliest";
+
+  /**
+   * Start position value indicating the consumer should begin from the latest offset.
+   */
+  public static final String CDC_START_POSITION_LATEST = "latest";
 }
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
@@ -136,4 +136,10 @@ public class DatastreamMetadataConstants {
    * which at least one partition violates the brooklin's permissible throughput bounds.
    */
   public static final String THROUGHPUT_VIOLATING_TOPICS = "throughputViolatingTopics";
+
+  /**
+   * Indicates whether the datastream requires CDC bootstrap, i.e. the consumer offset is
+   * initialized to an earlier position to replay historical data before catching up to the live tail.
+   */
+  public static final String CDC_BOOTSTRAP_REQUIRED_KEY = "system.cdcBootstrapRequired";
 }
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
@@ -70,7 +70,8 @@ public class SourceBasedDeduper extends AbstractDatastreamDeduper {
   public static final String CONFIG_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY = "cdcBootstrapRequiredMetadataKey";
 
   /** Default value of the metadata key used to identify CDC+BST streams. */
-  public static final String DEFAULT_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY = "system.cdcBootstrapRequired";
+  public static final String DEFAULT_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY =
+      DatastreamMetadataConstants.CDC_BOOTSTRAP_REQUIRED_KEY;
 
   /**
    * Grace-period window in <em>milliseconds</em> applied to existing CDC+BST streams when a new
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
@@ -7,36 +7,326 @@ package com.linkedin.datastream.server;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.linkedin.datastream.common.Datastream;
+import com.linkedin.datastream.common.DatastreamMetadataConstants;
 import com.linkedin.datastream.server.api.connector.DatastreamValidationException;
 
 
 /**
- * Deduper that uses the source to figure out whether two datastreams can be deduped.
+ * Deduper that uses the source connection string to determine whether two datastreams can share
+ * the same destination and tasks.
+ *
+ * <h3>Legacy behavior (default)</h3>
+ * <p>When the new stream does not carry the CDC-bootstrap metadata key (configurable via
+ * {@link #CONFIG_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY},
+ * default {@value #DEFAULT_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY}), the original logic applies:
+ * any existing stream with an identical source connection string is a dedup candidate and the
+ * first match is returned.
+ *
+ * <h3>CDC-bootstrap-aware behavior</h3>
+ * <p>When {@value #DEFAULT_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY} is present on the new stream,
+ * CDC-bootstrap-aware dedup is activated. Two stream types are recognized:
+ * <ul>
+ *   <li><em>CDC-only</em>  — metadata key absent or set to {@code false}. The consumer offset
+ *       is initialized to the <em>latest</em> Kafka offset, so the stream only reads events
+ *       produced after it was created.</li>
+ *   <li><em>CDC+BST</em>   — metadata key set to {@code true}. The consumer offset of source topic is
+ *       intentionally initialized to an <em>earlier</em> offset (before the current tail) so
+ *       that the stream can catch up on historical lag before switching to live replication.
+ *       During this catch-up window the stream is considered "in grace period".</li>
+ * </ul>
+ *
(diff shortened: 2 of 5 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
```diff
@@ -7,36 +7,326 @@
...
+  private Optional<Datastream> findDedupCandidateCdcBootstrapAware(Datastream newStream,
+      List<Datastream> candidates) {
+    boolean newStreamRequiresCdcBootstrap = Boolean.parseBoolean(
+        newStream.getMetadata().getOrDefault(_cdcBootstrapRequiredMetadataKey, "false"));
+
+    List<Datastream> sameSourceCandidates = candidates.stream()
+        .filter(d -> d.getSource().equals(newStream.getSource()))
+        .collect(Collectors.toList());
+
+    List<Datastream> cdcOnlyCandidates = sameSourceCandidates.stream()
```
akshayrai, line 175: To be precise - Since legacy datastreams do not have this metadata and we are defaulting to false now, this list could potentially have bootstrap enabled streams as well. But I think that is ok since those streams would have already caught up. Right?

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
```diff
@@ -7,36 +7,326 @@
...
+  private boolean isCdcBootstrapRequiredStream(Datastream stream) {
+    if (!stream.hasMetadata() || stream.getMetadata() == null) {
+      return false;
+    }
+    return Boolean.parseBoolean(stream.getMetadata().getOrDefault(_cdcBootstrapRequiredMetadataKey, "false"));
+  }
+
+  /**
+   * Returns {@code true} when the given stream's time window has expired, i.e.
+   * {@code creationTime + windowMs < currentTime}.
```
akshayrai: nit: Javadoc says creationTime + windowMs < currentTime, code uses >= (System.currentTimeMillis() >= creationMs + windowMs)

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
```diff
@@ -7,36 +7,326 @@
...
+   *
+   * <p>Returns {@code false} when {@code system.creation.ms} metadata is absent (cannot
+   * determine age; conservatively treat as not expired).
+   *
+   * @param stream   the existing candidate stream
+   * @param windowMs the window duration in milliseconds
+   * @return {@code true} if the window has expired, {@code false} otherwise
+   */
+  private static boolean hasWindowExpired(Datastream stream, long windowMs) {
+    if (!stream.hasMetadata() || stream.getMetadata() == null
```
akshayrai: nit: I think it will be safer to return true when windowMs == 0 regardless of creation_ms metadata present or not. Suppose some legacy datastream doesn't have this config.

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
```diff
@@ -7,36 +7,326 @@
...
+   *
+   * @param props deduper-domain properties (keys without the {@code deduper.} prefix,
+   *              as produced by {@code VerifiableProperties.getDomainProperties("deduper")})
+   */
+  public SourceBasedDeduper(Properties props) {
+    _cdcBootstrapRequiredMetadataKey = props.getProperty(CONFIG_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY,
+        DEFAULT_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY);
+    _newStreamGracePeriodMs = Long.parseLong(props.getProperty(CONFIG_NEW_STREAM_GRACE_PERIOD_MS,
+        String.valueOf(DEFAULT_NEW_STREAM_GRACE_PERIOD_MS)));
+    _snapshotWorstRefreshMs = Long.parseLong(props.getProperty(CONFIG_SNAPSHOT_WORST_REFRESH_DAYS,
```
akshayrai: use Math.multiplyExact to avoid any potential overflows.

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
akshayrai: here and elsewhere in the code : getMetadata() == null check is redundant. I think hasMetadata() already coveres this.

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
```diff
@@ -7,36 +7,326 @@
...
+  /**
+   * Creates a deduper configured from {@code props}.
+   *
+   * @param props deduper-domain properties (keys without the {@code deduper.} prefix,
+   *              as produced by {@code VerifiableProperties.getDomainProperties("deduper")})
+   */
+  public SourceBasedDeduper(Properties props) {
+    _cdcBootstrapRequiredMetadataKey = props.getProperty(CONFIG_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY,
+        DEFAULT_CDC_BOOTSTRAP_REQUIRED_METADATA_KEY);
+    _newStreamGracePeriodMs = Long.parseLong(props.getProperty(CONFIG_NEW_STREAM_GRACE_PERIOD_MS,
```
akshayrai: Use verifiableProperties.getLongInRange instead of Long.parseLong

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
mittalprince: Akshay I don't think that should be the case bootstrap stream can't be candidate for cdc dedup. Here allStreams coming as input to deduper  are only the streams with same connectorName only. So a BST stream (e.g. EspressoBootstrap) will never appear as a candidate when deduping a CDC stream (e.g. Espresso). Coordinator already filters allDatastreams by connectorName before passing them to the deduper.

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
mittalprince: Akshay, I didn’t fully understand your thought process here. This flow executes only when the `system.cdcBootstrapRequired` flag is set. If that condition is true, then windowMs should definitely exist.
Otherwise, returning true would mean the stream can always be deduped, which shouldn’t be the case in every scenario.

datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
```diff
@@ -136,4 +136,20 @@ public class DatastreamMetadataConstants {
...
+
+  /**
+   * Start position value indicating the consumer should begin from the earliest available offset.
+   */
+  public static final String CDC_START_POSITION_EARLIEST = "earliest";
+
+  /**
+   * Start position value indicating the consumer should begin from the latest offset.
+   */
+  public static final String CDC_START_POSITION_LATEST = "latest";
```
akshayrai: these 2 don't really fit in this class since they are not brooklin system properties. Is there any other place we can put them?

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
akshayrai: sorry, I meant to say bootstrap enabled cdc streams and not bootstrap stream.

datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
mittalprince: Added in `KafkaDatastreamMetadataConstants`

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
mittalprince: yes Akshay, those old created stream already caught up the lag and good to dedup.

akshayrai (approved):
Thanks for the quick turn-around.

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
```diff
@@ -7,36 +7,330 @@
...
+   * @param windowMs the window duration in milliseconds
+   * @return {@code true} if the window has expired, {@code false} otherwise
+   */
+  private static boolean hasWindowExpired(Datastream stream, long windowMs) {
+    if (windowMs == 0) {
+      return true;
+    }
+    if (!stream.hasMetadata()
+        || !stream.getMetadata().containsKey(DatastreamMetadataConstants.CREATION_MS)) {
+      return false;
```
kanishkjaiswal2015, line 331: so for streams that dont have this metadata, we would not dedupe against it? Should this be other way around as they would mostly be older streams that probably would have enough overlap with the source topic?

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
```diff
@@ -7,36 +7,330 @@
...
+   *
+   * <p>Returns {@code false} when {@code windowMs > 0} and {@code system.creation.ms} metadata
+   * is absent (cannot determine age; conservatively treat as not expired).
+   *
+   * @param stream   the existing candidate stream
+   * @param windowMs the window duration in milliseconds
+   * @return {@code true} if the window has expired, {@code false} otherwise
+   */
+  private static boolean hasWindowExpired(Datastream stream, long windowMs) {
+    if (windowMs == 0) {
```
kanishkjaiswal2015, line 326: is this condition ever true?

kanishkjaiswal2015 (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/SourceBasedDeduper.java
mittalprince: I have checked there are no such streams where creationtime.ms field doesn't exist. I have checked in all prod cluster for tidb & espresso.

PR #1014, opened 2026-05-19
Stream provisioning metric changes

dhananjay-sawner:
## Summary

- **`timeToReadyMs` histogram (Coordinator):** records the duration between datastream creation (`system.creation.ms`) and the `INITIALIZING -> READY` transition. Uses a **5-minute sliding window reservoir**; exposed via the standard `BrooklinHistogramInfo` attributes. In the live RRD pipeline only `.50thPercentile` and `.99thPercentile` are emitted — for sparse provisioning events (typically minutes-to-hours apart) the `.99thPercentile` time series effectively becomes a per-event scatter view, with each event landing as an isolated ~5-min plateau.
- **`slowProvisioningRate` meter (Coordinator):** incremented whenever a datastream's time-to-ready exceeds a configurable threshold. Emitted as `.PerSecondRate` in RRD, matching the established Brooklin rate-metric convention (e.g., `eventProduceRate`, `errorRate`). Enables a simple `threshold > 0` AutoAlerts rule with no derivative arithmetic
- **New config `brooklin.server.coordinator.slowProvisioningThresholdMs`:** controls the threshold above which a provisioning is counted as slow. Default 10 minutes. Threshold lives in config, so brooklin-server can tune per fabric without a library bump.
- **Per-datastream log line on READY:** `"Datastream {name} transitioned to READY in {N} ms"` — useful for debugging individual provisionings.
- Resume-from-STOPPED path is explicitly excluded from the metrics — only the initial `INITIALIZING -> READY` transition is counted. The new resume regression test guards this.

## Querying & alerting

- **Per-event scatter view (30/90 days):** plot `Coordinator.timeToReadyMs.99thPercentile` over the desired range in MDM Explorer / inGraphs — each plateau ≈ one provisioning event's duration.
- **Alert when stream provisioning takes longer time:** Alert on slowProvisioningRate > 0       
                                                                                                            
## Testing Done                                                                                                                                                          
- [x] `testTimeToReadyMetricsOnCreate` — with threshold = 0 ms, asserts both the histogram count and meter count increment after the stream goes READY, and the histogram sample is non-negative
- [x] `testSlowProvisioningRateDoesNotIncrementWhenBelowThreshold` — with threshold = 1 h, asserts the meter count stays unchanged after the stream goes READY
- [x] `testTimeToReadyMetricsNotEmittedOnResume` — STOPPED → READY resher the histogram or the meter (regression net)
- [x] `mint build`

### DIFF
--- a/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
+++ b/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
@@ -2492,13 +2492,6 @@ public class TestCoordinator {
     setup._datastreamKafkaCluster.shutdown();
   }
 
-  /**
-   * Verifies that the INITIALIZING -> READY transition for a newly created datastream:
-   *   1. Increments the Coordinator.timeToReadyMs histogram with a non-negative sample.
-   *   2. Increments the Coordinator.slowProvisioningCount counter when the configured
-   *      threshold is exceeded. A threshold of 0 ms guarantees any positive duration
-   *      is classified as slow.
-   */
   @Test
   public void testTimeToReadyMetricsOnCreate() throws Exception {
     String testCluster = "testTimeToReadyMetricsOnCreate";
@@ -2525,13 +2518,13 @@ public class TestCoordinator {
     ZkClient zkClient = new ZkClient(_zkConnectionString);
     DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, connectorType, datastreamName);
     Assert.assertTrue(PollUtils.poll(() -> DatastreamStatus.READY.equals(
-        DatastreamTestUtils.getDatastream(zkClient, testCluster, datastreamName).getStatus()), 200, WAIT_TIMEOUT_MS));
+        DatastreamTestUtils.getDatastream(zkClient, testCluster, datastreamName).getStatus()), 200, 30000));
 
     Histogram histogram =
         DynamicMetricsManager.getInstance().getMetric("Coordinator.timeToReadyMs");
     Assert.assertNotNull(histogram,
         "Coordinator.timeToReadyMs histogram should be registered after a stream goes READY");
-    Assert.assertTrue(PollUtils.poll(() -> histogram.getCount() > histogramCountBefore, 100, WAIT_TIMEOUT_MS),
+    Assert.assertTrue(PollUtils.poll(() -> histogram.getCount() > histogramCountBefore, 100, 30000),
         "timeToReadyMs histogram count did not increase after stream transitioned to READY");
     long maxValue = histogram.getSnapshot().getMax();
     Assert.assertTrue(maxValue >= 0, "timeToReadyMs sample should be non-negative, got: " + maxValue);
@@ -2540,7 +2533,7 @@ public class TestCoordinator {
         DynamicMetricsManager.getInstance().getMetric("Coordinator.slowProvisioningCount");
     Assert.assertNotNull(counter,
         "Coordinator.slowProvisioningCount counter should be registered after a stream goes READY");
-    Assert.assertTrue(PollUtils.poll(() -> counter.getCount() > counterCountBefore, 100, WAIT_TIMEOUT_MS),
+    Assert.assertTrue(PollUtils.poll(() -> counter.getCount() > counterCountBefore, 100, 30000),
         "slowProvisioningCount should increment when duration exceeds threshold");
 
     coordinator.stop();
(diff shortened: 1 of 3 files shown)


datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -2490,6 +2492,104 @@ public void testCreateDatastreamHappyPathDefaultRetention() throws Exception {
     setup._datastreamKafkaCluster.shutdown();
   }
 
+  @Test
```
akshayrai, line 2494: can you update PR description to reflect the latest changes in the diff.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -2686,7 +2725,8 @@ public boolean isEagerlyRegistered() {
      * Coordinator metrics of type {@link com.codahale.metrics.Counter}
      */
     public enum Counter {
-      NUM_HEARTBEATS("numHeartbeats");
+      NUM_HEARTBEATS("numHeartbeats"),
+      SLOW_PROVISIONING_COUNT("slowProvisioningCount");
```
akshayrai: Counter is monotonic and cumulative. Maybe we need a Meter?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -2583,6 +2615,13 @@ private void registerCounterMetrics() {
       Arrays.stream(Counter.values()).forEach(this::registerCounter);
     }
 
+    private void registerHistogramMetrics() {
+      // The metric is created lazily on first update via createOrUpdateSlidingWindowHistogram.
+      // Registering the BrooklinHistogramInfo here is required for the external metrics bridge
+      // to pick up the metric name
+      _metricInfos.add(new BrooklinHistogramInfo(_coordinator.buildMetricName(MODULE, TIME_TO_READY_MS)));
```
akshayrai, line 2638: Good to check if 15 min histogram will help and how. Datastream creation qps can be very very low right?

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorConfig.java
```diff
@@ -56,6 +58,7 @@ public final class CoordinatorConfig {
   public static final int DEFAULT_MARK_DATASTREMS_STOPPED_TIMEOUT_MS = 60 * 1000;
   public static final int DEFAULT_MARK_DATASTREMS_STOPPED_RETRY_PERIOD_MS = 10 * 1000;
   public static final int DEFAULT_LOG_SIZE_LIMIT_IN_BYTES = 1024 * 1024;
+  public static final long DEFAULT_SLOW_PROVISIONING_THRESHOLD_MS = Duration.ofMinutes(10).toMillis();
```
akshayrai: This is excluding validations right? should we split and account differently?

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1378,6 +1385,25 @@ private void handleDatastreamAddOrDelete() {
     _log.info("END: Coordinator::handleDatastreamAddOrDelete.");
   }
 
+  private void recordTimeToReadyMs(Datastream ds) {
+    String creationMsStr = Objects.requireNonNull(ds.getMetadata()).get(CREATION_MS);
```
akshayrai, line 1405: add in-line comment/javadoc in the code on when the creation_ms is populated for clarity

datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
```diff
@@ -2490,6 +2492,104 @@ public void testCreateDatastreamHappyPathDefaultRetention() throws Exception {
...
+    if (readyCounter != null) {
+      Assert.assertEquals(readyCounter.getCount(), countBefore,
+          "slowProvisioningCount must not increment when duration is below threshold");
+    }
+
+    coordinator.stop();
+    coordinator.getDatastreamCache().getZkclient().close();
+    zkClient.close();
+  }
+
```
akshayrai, line 2592: maybe add a regression test - metric should't be updated in the STOPPED → READY path. this should be emitted only on creation.

datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
dhananjay-sawner: Plan is to plot the max or p99 of the histogram window to get the individual stream provisioning times.
Using 15 min window in histogram to ensure that we have enough time to scrape the value.

Given that our current scrape window is per minute, so 5 minutes histogram window is also good enough to capture the value.

datastream-server/src/main/java/com/linkedin/datastream/server/CoordinatorConfig.java
dhananjay-sawner: Correct, updated it to 8 minutes, we can make it 10 again once we plan to accurately track provisioning time by setting validation time in datastream metadata in future.

akshayrai (approved):

mittalprince (approved):

PR #1018, opened 2026-05-21
Add eventsCommitToAckLatencyMs SLA metric

akshayrai:
## Summary

Adds a new SLA metric — `eventsCommitToAckLatencyMs` — that measures source DB commit → destination ack latency.

The existing `eventsLatencyMs` is computed from `DatastreamProducerRecord.eventsSourceTimestamp`, which on some CDC connectors reflects when Brooklin *read* the event (e.g. from an intermediate Kafka hop) rather than the original DB commit. This change introduces a separate, opt-in timestamp on the producer record and a parallel metric so connectors that can supply a true commit time expose a clean end-to-end SLA without changing existing metric semantics.

## Tests

New unit tests covering both the data path and the metric path:

`TestDatastreamProducerRecordBuilder`
- `eventsCommitTimestamp` absent by default
- round-trips through the builder
- propagated by `copyProducerRecord`
- absence preserved by `copyProducerRecord`

`TestEventProducer`
- metric is a no-op when no commit timestamp is supplied; existing `eventsLatencyMs` SLA still fires (regression guard)
- `eventsCommitWithinSla` increments when latency is inside the threshold
- `eventsCommitOutsideSla` increments when latency exceeds the threshold (tight 1ms threshold + 100ms-old commit)
- during grace, histogram redirects to `eventsCommitToAckLatencyMsSlaIneligible` and both counters are suppressed

### DIFF
--- a/datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamProducerRecord.java
+++ b/datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamProducerRecord.java
@@ -35,6 +35,11 @@ public class DatastreamProducerRecord {
   // timestamp for the events obtained from kafka header
   private Optional<Long> _eventsKafkaHeaderTimestamp = Optional.empty();
 
+  // Epoch-millis of the source DB commit for this event, when the connector can supply one (CDC connectors only).
+  // Distinct from _eventsSourceTimestamp, which on some connectors reflects the time Brooklin read the event from an
+  // intermediate hop rather than the original DB commit.
+  private Optional<Long> _eventsCommitTimestamp = Optional.empty();
+
   DatastreamProducerRecord(List<BrooklinEnvelope> events, Optional<Integer> partition, Optional<String> partitionKey,
       String checkpoint, long eventsSourceTimestamp) {
     this(events, partition, partitionKey, Optional.empty(), checkpoint, eventsSourceTimestamp, false);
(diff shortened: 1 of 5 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
```diff
@@ -530,6 +556,35 @@ private void reportMetrics(DatastreamRecordMetadata metadata, long eventsSourceT
...
+   * did not supply a commit timestamp (non-CDC sources, or CDC bootstrap/heartbeat paths). Gated by the same
+   * shouldEmitMetric() suppression as the existing source-to-ack SLA so the new metric inherits grace-period and
+   * per-datastream opt-out behavior.
+   */
+  private void reportCommitToAckMetrics(DatastreamRecordMetadata metadata, Optional<Long> eventsCommitTimestamp) {
+    if (!eventsCommitTimestamp.isPresent()) {
+      return;
+    }
+    long commitTs = eventsCommitTimestamp.get();
+    if (commitTs <= 0) {
```
mittalprince, line 570: Nit: In `reportCommitToAckMetrics`, we guard `commitTs <= 0` but not the case where the connector supplies a timestamp in the future. In that case System.currentTimeMillis() - commitTs becomes negative and we’ll emit a negative histogram value / classify SLA incorrectly. Just checking Do we need to add that check ?

mittalprince (approved):
LGTM

datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamProducerRecord.java
```diff
@@ -117,6 +122,21 @@ public void setEventsKafkaHeaderTimestamp(long timestamp) {
     _eventsKafkaHeaderTimestamp = Optional.of(timestamp);
   }
 
+  /**
+   * Get the source DB commit timestamp (Epoch-millis) if the connector supplied one. Present for CDC connectors
+   * that surface a true commit time; absent otherwise.
+   */
+  public Optional<Long> getEventsCommitTimestamp() {
```
kanishkjaiswal2015, line 129: So this field will only get set for CDC events?

kanishkjaiswal2015 (approved):

datastream-server/src/main/java/com/linkedin/datastream/server/EventProducer.java
akshayrai: commitTS cannot be in the future since the event has already happened right.

datastream-server-api/src/main/java/com/linkedin/datastream/server/DatastreamProducerRecord.java
akshayrai: yes, this applied only for CDC

PR #1028, opened 2026-06-03
Address security-scan findings from CodeQL java-security-extended suite and the Semgrep/CodeQL findings

akshayrai:
## Summary
Reviewed OSS Brooklin against the CodeQL java-security-extended suite and the Semgrep/CodeQL findings.
Fixes the genuine, low-risk issues with code changes and documents/suppresses the false-positive and compatibility-bound findings with justifications in code. Fixed:
- FileUtils.constructRandomDirectoryInTempDir created a predictably named directory under the shared system temp dir via new File(java.io.tmpdir, prefix + random) + mkdirs(), which is world-readable and racy (CWE-379, local information disclosure). Replaced with Files.createTempDirectory(), which atomically creates the directory with owner-only permissions. Callers (EmbeddedZookeeper, EmbeddedKafkaCluster) only use the returned path, so behavior is preserved; the now-unused Random field/import is removed.
- TestAssignmentTaskMapLogger.createCustomSizeString compared an int loop counter directly against a double bound (java/comparison-with-wider-type); the target size is now converted to int. Documented / suppressed with justification (false positive or compatibility-bound):
- NetworkUtils.getAvailablePort (Semgrep unencrypted-socket): the ServerSocket is bound to an ephemeral port only to discover a free port number and is closed immediately; no data is ever transmitted, so transport encryption does not apply. False positive.
- GroupIdConstructor.hashGroupId and AvroMessageEncoderUtil.md5 (Semgrep use-of-md5, java/potentially-weak-cryptographic-algorithm): MD5 is used as a non-cryptographic identity hash, not a security signature. The group-id hash determines the live Kafka consumer group.id and the schema-id hash is embedded in the on-wire message envelope; changing the algorithm would reset consumer offsets and break decoding of existing data. Must remain MD5 for compatibility. Each suppressed site carries an explanatory comment and a // nosemgrep directive so the rationale travels with the code (and the brooklin-server in-tree port stays clean). Verified on JDK 8: TestAssignmentTaskMapLogger and TestZkClient (which exercises the changed FileUtils via EmbeddedZookeeper) pass; checkstyle passes for the touched modules.

## Testing Done
Build passes

### DIFF
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/AvroMessageEncoderUtil.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/AvroMessageEncoderUtil.java
@@ -97,6 +97,11 @@ public class AvroMessageEncoderUtil {
 
   private static byte[] md5(byte[] bytes) {
     try {
+      // MD5 here is a non-cryptographic fingerprint embedded in the encoded message envelope (the
+      // schema-id hash prepended to the payload), not a security signature. Changing the algorithm
+      // would alter the on-wire format and break decoding of previously encoded data, so it must
+      // remain MD5 for wire compatibility. The Semgrep use-of-md5 finding is not applicable here.
+      // nosemgrep: java.lang.security.audit.crypto.use-of-md5.use-of-md5
       MessageDigest digest = MessageDigest.getInstance("md5");
       return digest.digest(bytes);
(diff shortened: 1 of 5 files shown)


harshcum (approved):
LGTM

dhananjay-sawner (approved):

harshOSS (approved):

PR #1030, opened 2026-06-04
Remove log4j 1.2.17 and migrate logging backend to log4j2

akshayrai:
## Summary
log4j 1.x is end-of-life and carries known CVEs. Brooklin logs entirely through SLF4J, so log4j was only ever the backend binding. Replace it with log4j2 — the same backend the LinkedIn brooklin-server already runs (log4j-api + log4j-core + log4j-slf4j-impl). Dependencies:
- Drop log4j:log4j:1.2.17 and org.slf4j:slf4j-log4j12.
- Add log4j-api/log4j-core/log4j-slf4j-impl as runtimeOnly. Keeping log4j-core off the compile classpath mirrors brooklin-server and prevents its bundled annotation processor (PluginProcessor) from activating annotation processing and tripping -Werror.
- Exclude any transitively-pulled log4j 1.x / slf4j-log4j12 so there is a single backend. Configuration:
- config/log4j.properties -> config/log4j2.properties (DailyRollingFileAppender becomes a RollingFile appender with an hourly TimeBasedTriggeringPolicy).
- Add config/tools-log4j2.properties for the CLI tools path in run-class.sh.
- Convert the six test log4j.xml files to log4j2.xml and add one for datastream-client, which previously had no config file and depended on the programmatic level call. Code:
- Remove the three org.apache.log4j.Logger.getRootLogger().setLevel(INFO) test calls; the effective level now comes from the test log4j2.xml, so no source compiles against log4j. Scripts:
- Update launch scripts (sh and bat) from -Dlog4j.configuration=...log4j.properties to the log4j2 system property -Dlog4j.configurationFile=...log4j2.properties.

## Testing Done
Build succeeds


harshOSS (approved):

khandelwal-ayush (approved):
Looks good. Thanks for taking this up.

PR #1031, opened 2026-06-12
Add generic DatastreamLifecycleListener extension point for DMS lifecycle events

akshayrai:
## Summary
Introduce an optional, best-effort DatastreamLifecycleListener notified after datastream lifecycle transitions are persisted through the DMS REST layer: create, update, delete, and pause/resume/stop status changes.
The listener is held on the Coordinator (default no-op) and invoked from DatastreamResources at a single, well-defined point per REST verb. STOP is emitted once, on the confirmed STOPPED transition, rather than for the intermediate STOPPING transition. Listener invocations are guarded: a listener failure is caught and logged and never affects the outcome of the REST request. This is a generic observability seam with no external dependencies; a downstream service can register an implementation to emit an audit/event stream of datastream lifecycle changes.

## Testing Done
unit test

### DIFF
--- a/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
+++ b/datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
@@ -50,6 +50,7 @@ import com.linkedin.datastream.server.DatastreamServer;
 import com.linkedin.datastream.server.ErrorLogger;
 import com.linkedin.datastream.server.HostTargetAssignment;
 import com.linkedin.datastream.server.api.connector.DatastreamValidationException;
+import com.linkedin.datastream.server.api.lifecycle.DatastreamLifecycleEventType;
 import com.linkedin.datastream.server.api.security.AuthorizationException;
 import com.linkedin.restli.common.HttpStatus;
 import com.linkedin.restli.server.ActionResult;
@@ -323,6 +324,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
 
         // invoke post datastream state change action for recently updated datastream
         invokePostDSStateChangeAction(datastreamMap.get(key));
+        notifyDatastreamLifecycle(DatastreamLifecycleEventType.UPDATE, datastreamMap.get(key));
       }
       _coordinator.broadcastDatastreamUpdate();
     } catch (DatastreamException e) {
@@ -381,6 +383,7 @@ public class DatastreamResources extends CollectionResourceTemplate<String, Data
           _store.updateDatastream(d.getName(), d, true);
           // invoke post datastream state change action for recently paused datastream
           invokePostDSStateChangeAction(d);
+          notifyDatastreamLifecycle(DatastreamLifecycleEventType.PAUSE, d);
         } else {
           LOG.warn("Cannot pause datastream {}, as it is not in READY state. State: {}", d, datastream.getStatus());
         }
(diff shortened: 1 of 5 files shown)


harshcum (dismissed):
LGTM

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
```diff
@@ -1001,6 +1008,20 @@ private void invokePostDSStateChangeAction(Datastream datastream) {
     }
   }
 
+  /**
+   * Best-effort notification of the configured {@link com.linkedin.datastream.server.api.lifecycle.DatastreamLifecycleListener}
+   * after a datastream lifecycle transition has been persisted. Listener failures are swallowed and never
+   * affect the outcome of the REST request.
+   */
+  private void notifyDatastreamLifecycle(DatastreamLifecycleEventType eventType, Datastream datastream) {
```
dhananjay-sawner, line 1016: 1. nit: Should we consider catching Throwable here instead of Exception to so that nothing escapes the catch and listener is truly ensured to not cause any issue in usual  datastream actions.
2. Can we make this dispatch async via single threaded executor for protecting against slow or stalled listener.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/dms/TestDatastreamResources.java
```diff
@@ -285,6 +288,80 @@ public void testStopDatastream() {
...
+    // Exactly one lifecycle event per verb, in order, all for this datastream. STOP fires once (on the
+    // confirmed STOPPED transition), not for the intermediate STOPPING transition.
+    Assert.assertEquals(recorded, Arrays.asList(
+        DatastreamLifecycleEventType.CREATE + ":" + datastreamName,
+        DatastreamLifecycleEventType.PAUSE + ":" + datastreamName,
+        DatastreamLifecycleEventType.RESUME + ":" + datastreamName,
+        DatastreamLifecycleEventType.STOP + ":" + datastreamName,
+        DatastreamLifecycleEventType.DELETE + ":" + datastreamName));
+  }
+
```
dhananjay-sawner, line 330: Can we add a unit test for updates as well.

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
akshayrai: 1. I think it is better to keep this at Exception. Catching Throwable would also swallow Error's like  OutOfMemoryError, StackOverflowError which are JVM-level signals that the server is in a bad state, and silently absorbing them inside an observability hook could mask a genuinely unhealthy instance.

2. The current design intentionally pushes async to the implementation. The
DatastreamLifecycleListener Javadoc requires implementations to be non-blocking and offload any
I/O to their own executor. Wrapping dispatch in an executor here would be a bit non-trivial.

datastream-server-restli/src/test/java/com/linkedin/datastream/server/dms/TestDatastreamResources.java
akshayrai: Added. thanks

datastream-server-restli/src/main/java/com/linkedin/datastream/server/dms/DatastreamResources.java
dhananjay-sawner: Sure, if we would be handling implementation in async that is also fine. Thanks

dhananjay-sawner (approved):
LGTM

harshcum (approved):
LGTM

PR #1032, opened 2026-06-13
Add stream provisioning SLO metric

dhananjay-sawner:
## Summary

- **Three provisioning SLO counters (Coordinator).** `recordStreamProvisioningTime` now emits, atomically at the `INITIALIZING -> READY` transition, `numStreamsProvisioned` (total) plus exactly one of `numStreamsProvisionedWithinSla` / `numStreamsProvisionedOutsideSla`, classified by the `brooklin.server.coordinator.provisioningSlaThresholdMs` threshold.
- **Per-connector breakdown.** Each counter is emitted both as an aggregate and keyed by connector name, so the SLO and its misses can be split per connector type. The keyed series render as `Coordinator.<connectorName>.numStreamsProvisioned` (and the within/outside SLA variants). 
- **Renamed the threshold config** from `slowProvisioningThresholdMs` to `provisioningSlaThresholdMs` to match the within/outside SLA framing. 
- **Removed the `slowProvisioningRate` meter.** `numStreamsProvisionedOutsideSla` replaces it. A Counter exposes `Count` in RRD, whereas the meter only exposed `PerSecondRate`. The outside-SLA counter is also the source for the slow-provisioning alert.
- **Enriched the `streamProvisioningTimeMs` histogram** to emit critical attributes (`Count`, `Max`, `Mean`, PERCENTILE_50, PERCENTILE_95, PERCENTILE_99.

## SLO definition

Provisioning SLO, computed over a reporting window:

```
SLO = numStreamsProvisionedWithinSla / numStreamsProvisioned
```

This measures, out of the datastreams that finished provisioning (reached READY), the fraction that did so within the threshold. All three counters are updated at the same moment, when a stream reaches READY, so the numerator and denominator always move together.

The same ratio can be computed per connector from the keyed series, for example `Coordinator.Espresso.numStreamsProvisionedWithinSla / Coordinator.Espresso.numStreamsProvisioned`. The per-connector miss count (`numStreamsProvisionedOutsideSla`) is meaningful at any volume and is the simplest way to attribute SLA misses to a connector.

## Why the denominator is provisioned (READY) streams, not create requests

We could instead use every create request as the denominator for SLO calculation but it has potential for lot of irregularities:

A create request is counted the moment a stream is created. But that same stream only counts as within SLA once it reaches READY, which is minutes later. In the gap between the two, the stream is in the denominator with nothing in the numerator, so the ratio drops even though nothing is wrong.

Example: the last 46 streams were all within SLA, so the SLO reads 46/46 = 100%. A new, healthy stream is created at 12:00 and reaches READY at 12:05. Between 12:00 and 12:05 the SLO shows 46/47 = 97.9%. At 12:05 it jumps back to 47/47 = 100%. That dip was not a breach. It only meant a stream was still being provisioned.

This causes two problems.

1. **The SLO graph dips when nothing is wrong.** Every time a stream is being provisioned, the SLO line drops for a few minutes and then recovers on its own. Someone looking at the dashboard sees the SLO fall below target and assumes something broke, when a stream was simply still in progress. These dips carry no real meaning, and an SLO chart that drops for reasons unrelated to the SLO is worse than no chart. This is the main reason we avoid the create-request denominator.

2. **Alerts fire falsely.** If we page when the ratio drops, a normal in-progress stream sets off the alert. One workaround is to alert only after the dip has lasted several minutes, long enough for a healthy stream to finish. That helps but does not fully solve it: it delays every real alert, and it still misfires when two streams are provisioned at overlapping times, because their dips add up and last longer than the wait. It also does nothing for problem 1, since the graph still dips.

There is also a smaller issue: a create request is counted before it is validated, so the count includes requests rejected for bad input or because the stream already exists. Those are never provisioned, so they do not belong in a provisioning SLO.

**What we do instead.** Count both the numerator and the denominator at the same moment, when the stream reaches READY. `numStreamsProvisioned` and `numStreamsProvisionedWithinSla` move together, so the ratio only changes when a stream actually finishes. Every point on the graph is a real result, and the SLO never dips while a stream is still provisioning.

**Caveat: Stuck streams** A stream that gets stuck and is deleted before it ever reaches READY is not counted in the SLO. This is fine: stuck streams are already caught by the existing alert on non-ready datastreams, which pages when a stream stays not-ready past 15 minutes. If a stuck stream is later fixed and reaches READY, it then counts as outside SLA.

## Alerting

- **Slow provisioning:** alert when `numStreamsProvisionedOutsideSla` goes up. This only happens after a stream has reached READY and was measured over the threshold, so a single increment is a real breach, not a false signal.
- **Stuck provisioning:** already covered by the existing alert on non-ready datastreams (pages when a stream stays not-ready past 15 minutes), so the SLO does not need to detect it.

## Testing Done

- [x] `testStreamProvisioningTimeMetricsOnCreate` (threshold 0): asserts the histogram count, `numStreamsProvisioned`, and `numStreamsProvisionedOutsideSla` all increment after the stream goes READY.
- [x] `testProvisioningSloCountsWithinSlaWhenBelowThreshold` (threshold 1 hour): asserts `numStreamsProvisionedWithinSla` increments and `numStreamsProvisionedOutsideSla` does not.
- [x] `testStreamProvisioningTimeMetricsNotEmittedOnResume` (threshold 0): asserts the histogram and counters do not change on a STOPPED to READY resume.
- [x] `testProvisioningSloCountersAreKeyedByConnector` (threshold 0): asserts the per-connector keyed series (`Coordinator.<connectorName>.numStreamsProvisioned` and the outside-SLA variant) increment alongside the aggregate.
- [x] `:datastream-server:checkstyleMain` and `:datastream-server-restli:checkstyleTest` pass.
- [x] `compileJava` and `compileTestJava` pass.

## Notes for Reviewer
- Intentionally kept the validation threshold to 8 minutes for now. Adding the validation time in the provisioning time will be taken up in a separate PR, will update the threshold to 10 mins with that change.

### DIFF
--- a/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
+++ b/datastream-server-restli/src/test/java/com/linkedin/datastream/server/TestCoordinator.java
@@ -52,6 +52,7 @@ import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
+import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
@@ -2509,9 +2510,13 @@ public class TestCoordinator {
         DynamicMetricsManager.getInstance().getMetric("Coordinator.timeToReadyMs");
     long histogramCountBefore = histogramBefore == null ? 0L : histogramBefore.getCount();
 
-    Meter meterBefore =
-        DynamicMetricsManager.getInstance().getMetric("Coordinator.slowProvisioningRate");
-    long meterCountBefore = meterBefore == null ? 0L : meterBefore.getCount();
+    Counter totalBefore =
+        DynamicMetricsManager.getInstance().getMetric("Coordinator.numStreamsProvisioned");
+    long totalCountBefore = totalBefore == null ? 0L : totalBefore.getCount();
+
+    Counter outsideSlaBefore =
+        DynamicMetricsManager.getInstance().getMetric("Coordinator.numStreamsProvisionedOutsideSla");
+    long outsideSlaCountBefore = outsideSlaBefore == null ? 0L : outsideSlaBefore.getCount();
 
     ZkClient zkClient = new ZkClient(_zkConnectionString);
     DatastreamTestUtils.createAndStoreDatastreams(zkClient, testCluster, connectorType, datastreamName);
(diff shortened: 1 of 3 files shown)


akshayrai (approved):

khandelwal-ayush (approved):
Looks good.

PR #1033, opened 2026-06-22
Add create-validation time to stream provisioning metric

dhananjay-sawner:
## Summary

- `recordStreamProvisioningTime` now adds the create-validation time to the provisioning duration. The value comes from the datastream metadata property `system.createValidationTime.ms`. Provisioning time becomes `(now - system.creation.ms) + system.createValidationTime.ms`, so the `streamProvisioningTimeMs` histogram and the within/outside SLA counters reflect validation time plus creation time.
- Absent or malformed `system.createValidationTime.ms` is treated as 0, so streams created before the property existed keep computing `now - system.creation.ms` exactly as before.
- Raise the default provisioning SLA threshold (`brooklin.server.coordinator.provisioningSlaThresholdMs`) from 8 to 10 minutes.
- Add `CREATE_VALIDATION_TIME_MS` to `DatastreamMetadataConstants`.

## Testing Done

- Added `TestCoordinatorStandalone` covering `getCreateValidationTimeMs`: present-and-positive (included), absent, non-positive, and malformed (all excluded as 0).
- Added `testProvisioningSlaThresholdConfig` to `TestCoordinatorConfig`: default is 10 minutes and the config override is honored.
- `./gradlew :datastream-server:test` passes.

## Notes for Reviewers

The new unit tests live in a new `TestCoordinatorStandalone` class rather than the existing `TestCoordinator`, for following reasons:

- Module. The code under test (`Coordinator.getCreateValidationTimeMs`) is in `datastream-server`, but the existing `TestCoordinator` is in `datastream-server-restli`. A unit test belongs in the same module as the code so `./gradlew :datastream-server:test` covers it.
- Type. The existing `TestCoordinator` is an integration test that stands up an embedded cluster. This is a pure, cluster-free unit test and does not belong in that test suite.
- No existing home. `datastream-server` had no Coordinator unit-test class, and the name cannot be `TestCoordinator` because that fully-qualified name is already used by the restli test. `TestCoordinatorStandalone` is the general home for cluster-free Coordinator unit tests.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -1427,11 +1427,7 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
     try {
       long streamProvisioningTimeMs = System.currentTimeMillis() - Long.parseLong(creationMsStr);
 
-      // Include the create-validation time (set by the resource provider via
-      // system.createValidationTime.ms) so the provisioning metric reflects validation time plus
-      // creation time. Read it from the metadata: add the value when present and greater than 0,
-      // otherwise add 0. This keeps the creation-time measurement intact for streams created before
-      // this property existed.
+      // Include the create-validation time in the stream provisioning time
       long createValidationTimeMs = 0L;
       String validationMsStr = ds.getMetadata().get(CREATE_VALIDATION_TIME_MS);
       if (validationMsStr != null) {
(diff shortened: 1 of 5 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -1425,7 +1426,11 @@ private void recordStreamProvisioningTime(Datastream ds) {
     }
     try {
       long streamProvisioningTimeMs = System.currentTimeMillis() - Long.parseLong(creationMsStr);
-      _log.info("Datastream {} transitioned to READY in {} ms", ds.getName(), streamProvisioningTimeMs);
+
+      // Include the create-validation time in the stream provisioning time
+      streamProvisioningTimeMs += getCreateValidationTimeMs(ds);
+
+      _log.info("Total stream provisioning time for Datastream {} - {} ms", ds.getName(), streamProvisioningTimeMs);
```
akshayrai: might be good to log all the details - validation time, creation time and total?

akshayrai (approved):

mittalprince (approved):

PR #1038, opened 2026-06-29
Checkstyle class-design cleanup: add private utility ctors, mark final, drop redundant modifiers, add missing @Override

akshayrai:
## Summary
No behavior changes in any file. All fixes are declaration-level no-ops that make existing intent explicit without changing runtime behavior.

* HideUtilityClassConstructorCheck (26): Added explicit private constructors to utility classes to prevent accidental instantiation, matching their existing all-static design.
* FinalClassCheck (6): Marked classes (and one nested class) as final where they were already effectively non-subclassable due to private constructors.
* RedundantModifierCheck (16): Removed redundant public modifiers from constructors of private or package-private nested types, since their visibility is already constrained by the enclosing type.
* MissingOverrideCheck (1): Added a missing @Override annotation to AbstractKafkaConnector#getActiveTasks(), aligning with its existing inherited Javadoc.

## Testing Done
No logic change. PR checks only

### DIFF
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
@@ -8,7 +8,7 @@ package com.linkedin.datastream.common;
 /**
  * Various well known config keys used in datastream metadata.
  */
-public class DatastreamMetadataConstants {
+public final class DatastreamMetadataConstants {
 
   private DatastreamMetadataConstants() {
   }
--- a/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
@@ -10,6 +10,9 @@ package com.linkedin.datastream.common;
  */
 public class DatastreamMetadataConstants {
 
+  private DatastreamMetadataConstants() {
+  }
+
   /**
    * Represents whether the datastream has an User managed destination (a.k.a BYOT - Bring your own topic)
    */
--- a/datastream-common/src/main/java/com/linkedin/datastream/kafka/KafkaDatastreamMetadataConstants.java
+++ b/datastream-common/src/main/java/com/linkedin/datastream/kafka/KafkaDatastreamMetadataConstants.java
@@ -10,6 +10,9 @@ package com.linkedin.datastream.kafka;
  */
 public class KafkaDatastreamMetadataConstants {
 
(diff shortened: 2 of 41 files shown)


datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
```diff
@@ -10,6 +10,9 @@
  */
 public class DatastreamMetadataConstants {
```
dhananjay-sawner: Should we define this class also as final

datastream-common/src/main/java/com/linkedin/datastream/kafka/KafkaDatastreamMetadataConstants.java
```diff
@@ -10,6 +10,9 @@
  */
 public class KafkaDatastreamMetadataConstants {
```
dhananjay-sawner: same, should this be defined a final class.

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectorDiagUtils.java
```diff
@@ -23,6 +23,10 @@
  * Utility class for Kafka based connectors
  */
 public class KafkaConnectorDiagUtils {
```
dhananjay-sawner: I think utils class should also be defined as final.

dhananjay-sawner:
Thanks Akshay for sanitizing this. 

One nit: since we are picking this up, should we also consider defining all the Utils and Constant classes as final?

datastream-kafka-connector/src/main/java/com/linkedin/datastream/connectors/kafka/KafkaConnectorDiagUtils.java
akshayrai: fixed

datastream-common/src/main/java/com/linkedin/datastream/kafka/KafkaDatastreamMetadataConstants.java
akshayrai: fixed

datastream-common/src/main/java/com/linkedin/datastream/common/DatastreamMetadataConstants.java
akshayrai: fixed

dhananjay-sawner (approved):
LGTM

harshOSS (approved):

PR #1040, opened 2026-06-29
Centralize ObjectMapper construction via JsonUtils.newObjectMapper()

akshayrai:
## Summary

Centralizes Jackson ObjectMapper creation through a single factory, JsonUtils.newObjectMapper(), establishing a single source of truth for mapper configuration. Every ObjectMapper returned by the factory is preconfigured with FAIL_ON_UNKNOWN_PROPERTIES = false, preventing failures during rolling deployments, schema evolution, or cross-version communication when additional JSON fields are introduced.

Previously, each call site instantiated its own new ObjectMapper() and had to configure (or remember to configure) the appropriate defaults. By routing construction through a shared factory, future configuration changes—such as registering new modules or enabling additional feature flags—are applied consistently across the codebase without requiring updates to every caller.

## Testing Done
PR checks

PR #1041, opened 2026-07-01
Fix duplicate JsonUtils() constructor causing build failure

akshayrai:
## Summary
Two parallel PRs (#1038 'Checkstyle class-design cleanup' and #1040 'Centralize ObjectMapper construction via JsonUtils.newObjectMapper()') each independently added a private no-arg constructor to JsonUtils, merging into master with two definitions of the same constructor. This broke compilation of datastream-common with:
error: constructor JsonUtils() is already defined in class JsonUtils Removed the duplicate, keeping the field declarations before the single remaining private constructor.

## Testing Done
PR checks


dhananjay-sawner (approved):
LGTM

harshOSS (approved):

PR #1042, opened 2026-07-07
Add configurable periodic refresh of throughput-violating topics cache

harshOSS:
## Summary
The coordinator's host-level throughput-violating topics cache was only rebuilt on assignment and datastream-update events, so between rebalances it could remain stale for a long time. This change adds a periodic rebuild from the current assignment so cache freshness is bounded by a configurable interval regardless of rebalance timing, and makes the behavior fully config-driven:

- `brooklin.server.coordinator.throughputViolatingTopicsRefreshPeriodMs` (default 5 minutes) — the refresh interval.
- `brooklin.server.coordinator.enableThroughputViolatingTopicsPeriodicRefresh` (default `true`) — a dedicated on/off toggle, additionally gated by the existing `enableThroughputViolatingTopicsHandling` feature flag.

Additional hardening in the same code path:
- `onAssignmentChange` now rebuilds the cache **before** queuing the async task-start event (matching `onDatastreamUpdate`), so a newly started producer cannot emit before its topic is present in the map.
- The shared rebuild path logs a `WARN` when an assigned task cannot be resolved and is dropped, or when a non-empty assignment yields an empty/partial rebuild, so silent partial rebuilds become detectable.

## Testing Done
- [x] Local code review completed
- [x] Unit tests added/updated
- [x] Integration tests pass
- [ ] Manual testing performed

Added and ran (JDK 17, Gradle 7.6.4):
- `TestCoordinatorConfig#testThroughputViolatingTopicsRefreshPeriodConfig`, `#testEnableThroughputViolatingTopicsPeriodicRefreshConfig`
- `TestCoordinator#testThroughputViolatingTopicsPeriodicRefreshRebuildsStaleMap`, `#testThroughputViolatingTopicsPeriodicRefreshEnablementGating`
- Existing throughput-violating `TestCoordinator` tests still pass; `checkstyleMain`/`checkstyleTest` are clean for `datastream-server` and `datastream-server-restli`.

🤖 Generated with [GitHub Copilot CLI](https://docs.github.com/copilot/github-copilot-cli)

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
@@ -192,6 +192,27 @@ public class Coordinator implements ZkAdapter.ZkAdapterListener, MetricsAware {
 
   private static final AtomicLong MAX_PARTITION_COUNT = new AtomicLong(0L);
 
+  // Trigger identifiers and log messages for (re)building the host-level throughput-violating topics cache.
+  private static final String TRIGGER_CREATE = "create";
+  private static final String TRIGGER_UPDATE = "update";
+  private static final String TRIGGER_PERIODIC_REFRESH = "periodic refresh";
+  private static final String POPULATE_VIOLATING_TOPICS_MSG =
+      "Populating the datastream violating topics to host level cache from the datastream objects on the {} trigger";
+  private static final String POPULATE_VIOLATING_TOPICS_EXCEPTION_MSG =
+      "Received an exception while populating the datastream violating topics to host level cache from the datastream "
+          + "objects on the {} trigger";
+  private static final String DROP_UNRESOLVED_TASK_MSG =
+      "Dropping task {} while rebuilding throughput-violating topics map on the {} trigger; its datastream task could "
+          + "not be resolved and will be omitted from this rebuild";
+  private static final String EMPTY_REBUILD_MSG =
+      "Throughput-violating topics map rebuild on the {} trigger produced an EMPTY result from a non-empty assignment "
+          + "of {} task(s); the host-level cache will be cleared. Possible partial assignment during rebalance.";
+  private static final String PARTIAL_REBUILD_MSG =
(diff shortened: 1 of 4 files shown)


datastream-server/src/main/java/com/linkedin/datastream/server/Coordinator.java
```diff
@@ -774,29 +823,66 @@ private void getAssignmentsFuture(List<Future<Boolean>> assignmentChangeFutures,
...
+  // Rebuilds the throughput-violating topics host-level cache from the current instance assignment.
+  // Shared by onAssignmentChange (create trigger) and the periodic refresh, so map freshness is
+  // decoupled from rebalance timing. Rebuild uses replace-all semantics via populateThroughputViolatingTopicsMap.
+  //
+  // Observability: a task that cannot be resolved (e.g. its ZNode disappeared mid-rebalance so
+  // getDatastreamTask returns null) is silently omitted from the replace-all rebuild; we WARN and count
+  // it so partial rebuilds are detectable. A non-empty assignment that yields no datastream groups
+  // would clear the cache to empty, so we WARN on that too.
+  private void refreshThroughputViolatingTopicsMap(String trigger) {
+    List<String> assignment = _adapter.getInstanceAssignment(_adapter.getInstanceName());
```
kanishkjaiswal2015, line 853: will this be updated with the latest assignment or older assignment? 

Can we just decouple the violating topics map from the assignments itself? Basically all throughout violating topics updated in the cache irrespective of whether or not they are assigned to the node?

harshOSS: > will this be updated with the latest assignment or older assignment?

- this will be always the latest assignment present in Zookeeper .



> Can we just decouple the violating topics map from the assignments itself? Basically all throughout violating topics updated in the cache irrespective of whether or not they are assigned to the node?

- IMO , we should not do this as on small cluster it will not be an issue but on large cluster it will be unncessary parsing of all assignments across the cluster (assignment can go upto 1k or even more) , so it will result in additional network call without any benefits.

kanishkjaiswal2015 (approved):

khandelwal-ayush (approved):

PR #1043, opened 2026-07-09
Add observability logs for load-based partition assignment

dhananjay-sawner:
## Summary

Adds INFO logs at the critical decision points of throughput-based partition assignment to verify that load balancing is actually working — i.e. distinguish **true load-based (throughput-balanced) assignment** from **round-robin fallback**, and observe whether the resulting per-task load is balanced.

Today the assigner logs `min/max partitions across tasks`, but load-based assignment intentionally makes partition **counts** uneven while balancing **byte rate**, so those counts do not prove balance. There is also no signal for whether throughput data was actually used, or whether an assignment fell back to round-robin.

## Changes (logging only — no behavior change)

**`LoadBasedPartitionAssignmentStrategy`**
- Log whether an assignment is **initial** (throughput fetched → true load-based balancing) or **incremental** (throughput not refetched → newly unassigned partitions placed round-robin). Throughput is only fetched when no partitions are currently assigned (STOPPED → READY, first assignment, or restart).
- After a successful fetch, log throughput-provider coverage: provider class, cluster name, and number of partition entries returned.

**`LoadBasedPartitionAssigner`**
- Log throughput coverage: `recognized` partitions split into partition-level vs topic-level fallback, vs `unrecognized` (round-robin) partitions, plus provider entry count.
- Log the resulting per-task throughput distribution (`min/max/avg/total/maxMinusMin` in KBps) — the direct signal that byte-rate load is balanced across tasks.

Together these answer: which assignment mode ran → did the provider return data → how many partitions got real rates vs round-robin → is the resulting per-task load even.

## Testing Done
- `./gradlew :datastream-server:compileJava :datastream-server:checkstyleMain` pass.
- `./gradlew :datastream-server:test --tests "*TestLoadBasedPartitionAssigner" --tests "*TestLoadBasedPartitionAssignmentStrategy"` — 15/15 pass.

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/LoadBasedPartitionAssigner.java
@@ -48,6 +48,7 @@ public class LoadBasedPartitionAssigner implements MetricsAware {
   private static final DynamicMetricsManager DYNAMIC_METRICS_MANAGER = DynamicMetricsManager.getInstance();
   private static final String MIN_PARTITIONS_ACROSS_TASKS = "minPartitionsAcrossTasks";
   private static final String MAX_PARTITIONS_ACROSS_TASKS = "maxPartitionsAcrossTasks";
+  private static final int KEY_SAMPLE_SIZE = 10;
 
   private final int _defaultPartitionBytesInKBRate;
   private final int _defaultPartitionMsgsInRate;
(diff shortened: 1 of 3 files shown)


harshOSS (dismissed):

harshOSS (approved):

manishGoyalCode (approved):
Looks good.
Please ensure there is no Null pointer exception due to this.

dhananjay-sawner:
> Looks good. Please ensure there is no Null pointer exception due to this.

Yes, verified again, there shouldn't be any NPE that can be newly introduced because of this.

PR #1048, opened 2026-07-23
Emit estimated tasks per datastream as an INFO-level metric

harshOSS:
## Summary

Add a per-datastream `estimatedNumTasks` gauge metric (with an accompanying INFO log) that reports the estimated number of tasks calculated for each stream during elastic task assignment.

- Previously we only emitted the `numTasksCappedByMaxTasks` meter, which fires **only** when the estimate exceeds the datastream's `maxTasks`. There was no visibility into the estimated task count for streams that were not capped.
- This change emits the estimated (uncapped) task count per datastream, so operators can observe the calculated value for every stream, compare it against the actual `numTasks` (which is capped at `maxTasks`), and understand scaling behavior.
- The metric is emitted from `StickyPartitionAssignmentStrategy.validateNumTasksAgainstMaxTasks()` — the common code path shared by both `StickyPartitionAssignmentStrategy` (partition-count-based estimate) and `LoadBasedPartitionAssignmentStrategy` (throughput/load-based estimate) — so both strategies are covered. The gauge is registered/declared/unregistered alongside the existing per-datastream gauges (`numTasks`, `actualPartitionsPerTask`).

Metric name: `StickyPartitionAssignmentStrategy.<datastreamTaskPrefix>.estimatedNumTasks`

## Testing Done

- [x] Local code review completed
- [x] Unit tests added/updated — extended `testElasticTaskPartitionAssignmentCreatesAtMostMaxTasks` to assert the new gauge reports the uncapped estimate (13) while `numTasks` reports the capped actual (5)
- [x] `TestStickyPartitionAssignmentStrategy` (15 tests) and `TestLoadBasedPartitionAssignmentStrategy` pass under JDK 17
- [x] `checkstyleMain` and `checkstyleTest` pass

🤖 Generated with [GitHub Copilot CLI](https://docs.github.com/copilot/github-copilot-cli)

### DIFF
--- a/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
+++ b/datastream-server/src/main/java/com/linkedin/datastream/server/assignment/StickyPartitionAssignmentStrategy.java
@@ -80,6 +80,8 @@ public class StickyPartitionAssignmentStrategy extends StickyMulticastStrategy i
   static final String ELASTIC_TASK_PARAMETERS_NEED_ADJUSTMENT = "elasticTaskParametersNeedAdjustment";
   @VisibleForTesting
   static final String NUM_TASKS_CAPPED_BY_MAX_TASKS = "numTasksCappedByMaxTasks";
+  @VisibleForTesting
+  static final String ESTIMATED_NUM_TASKS = "estimatedNumTasks";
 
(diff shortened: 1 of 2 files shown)


mittalprince (approved):
LGTM

dhananjay-sawner (commented):
Just curious why do we need this metric?

dhananjay-sawner (approved):
LGTM

